PR #197, opened 2014-01-03
Integrated announcements

acolebourne:
This pull requests goes along with https://github.com/Jasig/AnnouncementsPortlet/pull/25


drewwills:
I'm going to to a bit of housecleaning and close this PR.

I'm delighted with the notion of better-integrating content (especially content from Apereo portlet projects) into the portal experience, but that's an area where we have had significant changes and innovations in the 15 months since this PR was opened.  I think we might update this contribution more than a little if we were to propose it today -- and that would be completely welcome, should the opportunity present itself.

PR #210, opened 2014-01-24
UP-3307: Leveraging unload function to fix back when page is returned to via the back button

shawnconnolly:
Leveraging unload function to reset the page back to a fresh state if the back button is pressed, by running same javascript as ready. Refactored the script into its own function to avoid dup code.


drewwills:
I'm going to do some housecleaning and close this one.

PR #317, opened 2014-05-07
UP-4612 (Ref) - SpEL contextPath for Background Preferences

jgribonvald:
Adding automatic contextpath translation using SpEL to avoid to define a
static contextpath in background portlet preferences. Using same scheme as it is done for image portlet

### DIFF
--- a/uportal-war/src/main/data/quickstart_entities/portlet-definition/background-preference.portlet-definition.xml
+++ b/uportal-war/src/main/data/quickstart_entities/portlet-definition/background-preference.portlet-definition.xml
@@ -73,26 +73,26 @@
     </parameter>
     <portlet-preference>
         <name>defaultBackgroundImages</name>
-        <value>/uPortal/media/skins/universality/backgroundImages/galaxy.jpg</value>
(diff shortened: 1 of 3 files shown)


jgribonvald:
Just a reminder !
any news about this PR ?

Thx

drewwills:
@jgribonvald -- I have just merged a fix based on your solution...

  - https://github.com/Jasig/uPortal/commit/b8cd3e97fa74413c106752ea2d203e0f5d5fd1bc

A lot has changed since this Pull Request.  In testing, I discovered some other issues with the BackgroundPreference portlet.  I updated the solution in the PR to address those.

I also changed it slightly.  I believe...

* The use case you brought forth is very valid and reasonably common;  needs supporting
* This use case is likely to be the only one we will need to support with SpEL

So I simplified it slightly.

Image paths may include the token '${portalContext}', which will evaluate to '/uPortal' in the default setup.  (The portlet-definition.xml has been updated to reflect this feature.)

I moved all the SpEL setup to a `@PostConstruct` method so it can occur one time.

PR #379, opened 2014-07-09
[UP-3315] Fix some XSS issue around inputting usernames.  The issue was ...

jhelmer-unicon:
...occurring because "</script>" always terminates the

JS block -- even if that's inside a string.  The JS interpreter would terminate the block and then misinterpret the unterminated
contents of the string as executable code.  The *right* solution for this is probably to start whitelisting the
input characters for every input field. Any field that will be part of a RESTful URL should not be allowed to contain the '/'
character.  The "solution" in this commit is intended as a stop-gap.  It solves this biggest issue -- the XSS vulnerability --
but it doesn't fully solve the issues caused by invalid chars in the input.  IMO, this is better than what we had, but
still not great.
(cherry picked from commit b034dec)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
         <artifactId>jasig-parent</artifactId>
         <version>35</version>
     </parent>
-    
+
     <prerequisites>
         <maven>3.0.0</maven>
     </prerequisites>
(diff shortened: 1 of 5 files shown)


drewwills:
Looks good to me too.  Merging.

drewwills:
Should probably have been opened on master, however.

PR #399, opened 2014-08-06
UP-4200 Decrypt password for portlets when replaying credentials

jameswennmacher:
https://issues.jasig.org/browse/UP-4200

Only applies when using local login form to authenticate against DB or LDAP.


drewwills:
I like both @apetro's suggestion ("make CachedPasswordUserInfoService detect whether the password needs decrypting and decrypt only if needed") as well as @jameswennmacher's proposal ("change CacheSecurityContext to return the password in plaintext rather than encrypted and remove decryptPassword from CachedPasswordUserInfoService") *much better than the solution in this PR*.

I'm going to update the JIRA with these, and close this PR.

PR #401, opened 2014-08-11
External Database Group store implementation

cjgwhite:
Implementation of the GroupStore API for using an external Database to hold username<-->group_name information.

Back in November 2013 (I believe) Anthony Colebourne started a discussion concerning building a custom Group source for uPortal that would work against an external Database source…

http://jasig.275507.n4.nabble.com/Custom-group-source-tt4661293.html

http://jasig.275507.n4.nabble.com/Custom-Group-Source-td4663482.html

We still have a requirement to be able to drive groups and group membership from an external database. To that end we have recently got round to doing the work to create an implementation of the GroupService to meet this requirement. 

this is the pull request of our implementation. It is heavily based on the smartLDAP implementation. Any comments / suggestions would be very much appreciated.

This implementation assumes no hierarchy in the groups held within the external database -- all groups are direct children of the Root Group for the GroupStore ('DB Root'). The implication being that DB backing this is the simplest possible structure (username <--> group_name pairs held in a single table). There is no reason why this could not be backed by more complex structures if required with changes to the supplied queries (implementors should change these for their own implementation anyway).

Best Regards
Chris

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/groups/db/ExternalDBGroupStore.java
+++ b/uportal-war/src/main/java/org/jasig/portal/groups/db/ExternalDBGroupStore.java
@@ -1,0 +1,345 @@
+/*
+ * Copyright 2014 Jasig.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jasig.portal.groups.db;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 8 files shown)


timlevett:
:+1:

uportal-war/src/main/java/org/jasig/portal/groups/db/ExternalDBGroupStore.java
```diff
@@ -0,0 +1,345 @@
...
+            
+            synchronized (groups) {
+                groups.clear();
+                groups.putAll(newGroups);
+            }
+            log.warn(groups);
+    }
+    
+    @Override
+    public boolean contains(IEntityGroup group, IGroupMember member) throws GroupsException {
```
drewwills: Two questions about this method:

- Can this flavor of groups contain child groups?  (Can those groups be other flavors?)  If not, you may want to check for those qualities and `return false` quickly
- Doesn't this method get called a lot?  Is there a way to implement this method that doesn't involve a database query for every invocation?

uportal-war/src/main/java/org/jasig/portal/groups/db/ExternalDBGroupStore.java
```diff
@@ -0,0 +1,345 @@
...
+
+            List<IEntity> entities = new ArrayList<IEntity>();
+            for (Map<String, Object> row : rows) {
+                IEntity entity = newInstance((String) row.get("username"), IPerson.class);
+                entities.add(entity);
+            }
+            
+            return entities.iterator();
+        } else {
+            return findMemberGroups(group);
```
drewwills: I'm not sure this is what's supposed to happen (though I haven't read the method yet).  I think findEntitiesForGroup is supposed to return only non-group members.

drewwills:
I propose closing this PR due to age and inactivity.

drewwills:
It seems we're prepared to close this item at this point.

PR #463, opened 2014-11-11
UP-4291 MarketplacePortletDefinition renderUrl is alternativeMaximizedLink when available.

apetro:
When launching a portlet from Marketplace, that launch URL should be the alternative maximized URL of the portlet if it has one, so that the user can cut to the chase and launch the best available experience of that portlet.

Implementation:

When the underlying `IPortletDefinition` of a `MarketplacePortletDefinition` has an `alternativeMaximizedLink`, then `MarketplacePortletDefinition` `getRenderUrl()` returns that alternative maximized link instead of a URL that renders the portlet by `fname`. ([UP-4291](https://issues.jasig.org/browse/UP-4291)).

Which entailed adding `getAlternativeMaximizedLink()` syntactic sugar method on `IPortletDefinition` so that computation of whether and what alternative maximized link a portlet definition has can be encapsulated in the portlet definition implementations. ([UP-4290](https://issues.jasig.org/browse/UP-4290)).

I'm not loving the name "alternativeMaximizedLink" rather than "alternativeMaximizedUrl", but the existing portlet definition parameter is named "alternativeMaximizedLink", so this solution retains that.

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/rest/LayoutRESTController.java
+++ b/uportal-war/src/main/java/org/jasig/portal/rest/LayoutRESTController.java
@@ -155,9 +155,9 @@ public class LayoutRESTController {
                     }
                     
                     //get alt max URL
-                    IPortletDefinitionParameter alternativeMaximizedLink = def.getParameter("alternativeMaximizedLink");
+                    String alternativeMaximizedLink = def.getAlternativeMaximizedLink();
                     if( alternativeMaximizedLink != null) {
-                    	portlet.setUrl(alternativeMaximizedLink.getValue());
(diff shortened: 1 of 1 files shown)


jgribonvald:
Just a question, there is already a alternativeMaximi**x**edLink but with two x, this should be the same parameter or not ? this should be renamed ?
cf https://github.com/Jasig/uPortal/search?utf8=%E2%9C%93&q=alternativeMaximixedLink&type=Code

I didn't see any renaming ...

timlevett:
@jgribonvald - that is interesting, guessing just a typo in the original one with the x? We should fix the x to a z.

timlevett:
Created https://issues.jasig.org/browse/UP-4506

PR #489, opened 2014-12-12
UP-4352 Updating PostgreSQL dependency in uportal-db

asgrant:
Updating dependency for PostgreSQL to have group "org.postgresql" instead of just "postgresql" this is the group where new artifacts for PostgreSQL driver (v9.2+) are being stored.


gcarroll:
:+1:  I had encountered this awhile back.  The latest version of PostgreSQL is v9.4 I haven't had a chance to test on the version yet.

PR #490, opened 2014-12-12
UP-4320 Demonstrate BROWSE permission filtering

apetro:
Adds an example portlet definition and example `BROWSE` permissions configuration to demonstrate that the student user does not see a related portlet that the staff user does see, due to the student user lacking a BROWSE permission ( [UP-4320](https://issues.jasig.org/browse/UP-4320) ).

This demonstrates that `BROWSE` permission filtering on related portlets works and sets the stage to validate that Marketplace (and other?) search filters on `BROWSE` when that feature is implemented in the future.


jameswennmacher:
@apetro I think we can close this PR without merging at this point.  BROWSE is heavily used now with https://github.com/Jasig/uPortal/pull/532 which I plan to merge tomorrow.

PR #494, opened 2014-12-17
[UP-4342, UP-4343] make name and title match

jhelmer-unicon:
This PR addresses 2 tickets (too related to easily do as separate PRs).

https://issues.jasig.org/browse/UP-4342
https://issues.jasig.org/browse/UP-4343

1.  On the admin screen, as the user types in the title field, it will automatically fill in the name/fname.  name will match the title and fname will be convert title to all lowercase and then convert any sequence of non-alpha characters to a dash  (eg.  Sample & Demo Portlet -> sample-demo-portlet)
2. If the name or fname field is manually edited to be different than the calculated value, will disable the automatic updates for that field.
3.  If the name and title are changed so that they no longer match an info alert will be displayed to warn users that that is not common.

Screens
up-4342


uportal-war/src/main/webapp/WEB-INF/flows/edit-portlet/basicInfo.jsp
```diff
@@ -199,4 +228,179 @@ PORTLET DEVELOPMENT STANDARDS AND GUIDELINES
...
+                NameTitleModel,
+                nameTitleModel,
+                nameTitleMismatchTimer;
+
+        /**
+         * Define a backbone model to capture the name/title/fname values and
+         * orchestrate their updates.  Changes on the page should update the model
+         * and allow listeners attached to the model to handle propagating the changes
+         * to the appropriate places in the UI
+         */
```
apetro: :heart: :heart: :heart: the inline documentation.

apetro:
Invoking Lazy Consensus.  Unless otherwise actioned, mergeable without further discussion on `2015-01-05`.

timlevett:
:+1:

PR #496, opened 2014-12-30
UP-4371 wrap Cobertura <excludes> in <instrumentation>.

apetro:
[UP-4371](https://issues.jasig.org/browse/UP-4371): As [documented](http://mojo.codehaus.org/cobertura-maven-plugin/usage.html), the `<excludes>` Cobertura configuration ought to be wrapped in an `<instrumentation>`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1559,13 +1559,13 @@
                         <maxmem>256m</maxmem>
                         <!-- aggregated reports for multi-module projects -->
                         <aggregate>true</aggregate>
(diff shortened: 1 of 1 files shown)


apetro:
Trivial technical fix.  Invoking Lazy Consensus.  Unless otherwise actioned, mergeable without further discussion on `2015-01-05`.

apetro:
Merged per Lazy Consensus.

PR #498, opened 2014-12-30
UP-4209 require Servlet3 / Tomcat 7+.

apetro:
[UP-4209](https://issues.jasig.org/browse/UP-4209) : Updates `README.md` and release notes documentation to state requirement of Servlet v3 / Tomcat 7 (rather than stating support for Servlet 2.5 which implies Tomcat 6.)

This requirement is only documentation-deep at this point.  This updates the documentation so that the project can confidently not worry about Servlet 2.5 / Tomcat 6 support for uPortal 4.2 and beyond.  Presumably future opportunities will arise to do servlet-spec-y things that benefit from that Servlet 3 version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,8 @@ uPortal uses Travis-CI for lightweight continuous integration.  You can see buil
 
 ## Requirements
 * JDK 1.7 or later - Just a JRE is not sufficient, a full JDK is required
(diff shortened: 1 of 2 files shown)


apetro:
Invoking Lazy Consensus.  If not otherwise actioned mergeable without further discussion on `2015-01-05`.

PR #500, opened 2014-12-31
UP-4369 Respondr: Focus on portlet also present in region displays same portlet twice

codepuppet:
Altered the Respondr content.xsl to determine if the the focused portlet exists in a region.  If it exists, then there is no need to display it again in the main content area.  Particularly problematic for the Notification portlet when it is added to the sidebar along with having the Notification Icon.

before
after

### DIFF
--- a/uportal-war/src/main/resources/layout/theme/respondr/content.xsl
+++ b/uportal-war/src/main/resources/layout/theme/respondr/content.xsl
@@ -322,12 +322,34 @@
    | These two templates render the focused portlet content.
   -->
   <xsl:template match="focused">
+    <!--
+        Check for the existance of the "focused" portlet in a region.  If one is found, prevent the portlet from rendering
+        in the content portion of the layout.  Test for a match against the fname, ID, and channel ID.
+    -->
+    <xsl:variable name="focusedFname">
+        <xsl:value-of select="//focused/channel/@fname"/>
+    </xsl:variable>  
+
+    <xsl:variable name="focusedChannelID">
+        <xsl:value-of select="//focused/channel/@chanID"/>
+    </xsl:variable>
+
+    <xsl:variable name="focusedID">
+        <xsl:value-of select="//focused/channel/@ID"/>
+    </xsl:variable>
+
+    <xsl:variable name="focusedMatchTest">
+        <xsl:value-of select="not(((//region/channel[@fname=$focusedFname]) and (//region/channel/@chanID=$focusedChannelID)) and (//region/channel[@ID=$focusedID]))"/>
+    </xsl:variable> 
+    
     <div class="fluid-row">
         <div class="col-md-12">
             <div id="portalPageBodyColumns" class="columns-1">
                 <div class="portal-page-column column-1">
                     <div class="portal-page-column-inner"> <!-- Column inner div for additional presentation/formatting options.  -->
-                    <xsl:apply-templates select="channel|blocked-channel"/>
+                        <xsl:if test="not($focusedMatchTest = 'false')">
+                            <xsl:apply-templates select="channel|blocked-channel"/>
+                        </xsl:if>
                   </div>
                 </div>
             </div>
--- a/uportal-war/src/main/resources/layout/theme/respondr/content.xsl
+++ b/uportal-war/src/main/resources/layout/theme/respondr/content.xsl
@@ -323,8 +323,9 @@
   -->
   <xsl:template match="focused">
     <!--
-        Check for the existance of the "focused" portlet in a region.  If one is found, prevent the portlet from rendering
+        Check for the existence of the "focused" portlet in a region.  If one is found, prevent the portlet from rendering
         in the content portion of the layout.  Test for a match against the fname, ID, and channel ID.
+        See UP-4369 for additional information.
     -->
     <xsl:variable name="focusedFname">
         <xsl:value-of select="//focused/channel/@fname"/>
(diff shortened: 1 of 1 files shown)


uportal-war/src/main/resources/layout/theme/respondr/content.xsl
```diff
@@ -322,12 +322,34 @@
    | These two templates render the focused portlet content.
   -->
   <xsl:template match="focused">
+    <!--
+        Check for the existance of the "focused" portlet in a region.  If one is found, prevent the portlet from rendering
```
apetro: "existance" --> "existence".

uportal-war/src/main/resources/layout/theme/respondr/content.xsl
```diff
@@ -322,12 +322,34 @@
...
+
+    <xsl:variable name="focusedChannelID">
+        <xsl:value-of select="//focused/channel/@chanID"/>
+    </xsl:variable>
+
+    <xsl:variable name="focusedID">
+        <xsl:value-of select="//focused/channel/@ID"/>
+    </xsl:variable>
+
+    <xsl:variable name="focusedMatchTest">
```
apetro: `focusedMatchTest` needs a comment documenting what concept it is trying to represent, under what circumstances it will be `true

uportal-war/src/main/resources/layout/theme/respondr/content.xsl
```diff
@@ -322,12 +322,34 @@
...
+        <xsl:value-of select="not(((//region/channel[@fname=$focusedFname]) and (//region/channel/@chanID=$focusedChannelID)) and (//region/channel[@ID=$focusedID]))"/>
```
apetro: Does this have to be this complicated?  I admit I'm not fully understanding what this code is trying to do, but stepping back a bit -- either matching on portlet publication matters, so that would just be `@fname`, or matching on window ID matters, so that's, what, `@ID`?

uportal-war/src/main/resources/layout/theme/respondr/content.xsl
```diff
@@ -322,12 +322,34 @@
...
+    
     <div class="fluid-row">
         <div class="col-md-12">
             <div id="portalPageBodyColumns" class="columns-1">
                 <div class="portal-page-column column-1">
                     <div class="portal-page-column-inner"> <!-- Column inner div for additional presentation/formatting options.  -->
-                    <xsl:apply-templates select="channel|blocked-channel"/>
+                        <xsl:if test="not($focusedMatchTest = 'false')">
+                            <xsl:apply-templates select="channel|blocked-channel"/>
+                        </xsl:if>
```
apetro: So, what should Respondr display in the content pane when it doesn't display the addressed portlet (because that portlet exists in a region)?  **Not displaying anything yields the large blank space shown in the screenshot, and that's confusing.**

The motivating issue for this pull request seems to be clicking the Notification Icon in the header and the resulting focusing of the Notifications portlet, which might also be in a region if you've, say, included Notifications in a sidebar.  So.  **What should clicking the Notification Icon really do in this case?**  I imagine users expect it to focus the Notifications portlet?  So maybe rather than suppressing the portlet rendering in the content pane, the solution should be suppressing the portlet rendering in the sidebar?

drewwills:
When we say...

> [...] determine if the the focused portlet exists in a region.

how are we measuring sameness?  Are the portlets in region+content the same...

1. By implementation (<portlet-definition> in a portlet .war)
2. By publication record (same fname);  or
3. By windowId?

This list is hierarchical -- to adhere to #2, you also have to be #1l  and to match #3, you have to match both #1 & #2.

#1 & #2 (versions of "sameness") should work with multiple copies on the page -- whether in regions or not.  There are ways (to be certain) to break it, but this class of issues is what the normal citizenship (js/css) rules are for -- follow them, and it will work.

#3 does not work and should not occur in deployments... afaik.

I'm not sure we need anything for #3.

apetro:
I think @drewwills is saying this changeset is not needed in general and in specific Notifications would better handle multiple inclusions on the page if it namespaced element `ID`s better.  I think.

That seems the most fruitful line of inquiry: avoiding introducing this wrinkle to Respondr sounds good?

However:

If something like this changeset *is* needed, then my previous comment stands.  It disappeared since it was attached to a line of code touched again in  053be7b , so reproduced below now scoped to the Pull Request generally. :smile:

This changeset causes Respondr to not display the addressed portlet in the content pane when that portlet exists in a region.  Supposing that not-displaying is desired, **what should it display in that content pane instead?**  Not displaying anything yields the large blank space shown in the screenshot, and that's confusing.

The motivating issue for this pull request seems to be clicking the Notification Icon in the header and the resulting focusing of the Notifications portlet, which might also be in a region if you've, say, included Notifications in a sidebar. So. What should clicking the Notification Icon really do in this case? I imagine users expect it to focus the Notifications portlet? So maybe rather than suppressing the portlet rendering in the content pane, the solution should be suppressing the portlet rendering in the sidebar?

drewwills:
> Supposing that not-displaying is desired, what should it display in that content pane instead? Not displaying anything yields the large blank space shown in the screenshot, and that's confusing.

:+1: 

I'm not sure a uPortal change is needed here, but the most likely/attractive one would be a change to the rendering pipeline such that it only matches (by fname) portlets in the tab/column hierarchy.  (See UPDATE above.)

codepuppet:
After discussion, this pull request will be closed.  An alternate solution will be implemented in the private project.

PR #503, opened 2015-01-14
UP-4380 Updating Less code so that portlet-options-link-color works

asgrant:
By changing the portal-titlebar color in Less it would also update the portal-options-links color. This patch makes it so the two variables can work independently.


apetro:
:+1:

drewwills:
Looks good -- :+1:

PR #505, opened 2015-01-28
UP-4307 Show most recent login timestamp in user details

apetro:
DO NOT MERGE THIS PULL REQUEST.  

# Meta

[UP-4307](https://issues.jasig.org/browse/UP-4307).

*This "Pull Request" is currently a way of communicating progress that is reviewable, without being an actual request that these changes be pulled.*

*The branch behind this Pull Request may and probably will `rebase` before this becomes an for-real request that the changeset be pulled (merged) to `master`.*

# Summary

In the (administrative) user admin details page on a user, show the timestamp of the user's most recent login to the portal, if available.

Screenshot when available:

most_recent_login_timestamp

Screenshot when not:

dunno_when_last_logged_in

## Implementation notes

Stores this data via JPA behind a DAO and a registry.  

last_login_table

Adds an `AuditService` in which this noting and checking last login timestamp is the first functionality, the only functionality so far.

 * Impersonated logins do not bump the impersonated user's last login timestamp.
 * Last login timestamp of guest user is not recorded.
 * While the offered implementation of the introduced APIs only stores the most recent login, the API is multiply realizable such that an alternative implementation could store all the logins for a user in more of an audit trail style.

## Tradeoffs made, false starts

I started down the road of modeling last login timestamp as a user attribute.  That would have some nice properties in that the portal already knows how to permiss access to user attributes on users, the user admin portlet already knows how to display user attributes, it would make the last login data available as a user attribute that could potentially be useful in other contexts, fed to portlets, etc.

However, that had the significant barrier to entry of having to write an `IPersonAttributeDao` implementation.  Also, a naive implementation would add that `DAO` to the attribute gathering at user login, which would contribute to the login process taking so much clock time without adding present value.  Ended up giving up on modeling as a user attribute for now.  Of course a subsequent changeset could add such a DAO, and that DAO could even rely upon an abstraction introduced by this API (the `AuditService`, say.)

## Known shortcomings:

 * While it seems to work, that `JPA` `DAO` implementation is atrocious.  Patches to accomplish that less awkwardly are especially welcome.
 * While the new user-facing text in the user details `JSP` is extracted into message bundle key-value pairs, the way the text prefixes the timestamp is hard-coded in the JSP rather than using placeholder interpolation.  This may prove limiting for translating to some locales.
 * Aside from being triggered by the pre-existing `LoginEvent`, does not make use of the existing event aggregation API stuff.
 * Probably ought to batch process events rather than having each login yield a database transaction.


uportal-war/src/main/java/org/jasig/portal/events/LoginEvent.java
```diff
@@ -64,6 +67,40 @@ private LoginEvent() {
...
+     * @param userAttributes non-null user attributes of the user who logged in
+     * @param source non-null source of the LoginEvent
+     * @param serverName non-null name of the server logged in to
+     * @param eventSessionId on-null id of the session the user logged in to
+     * @param person non-null person who has logged in
+     * @param portalRequest non-null request in which this login has happened
+     *
+     * @since uPortal 4.2
+     */
+    public LoginEvent (
```
timlevett: Noticed this is `public` but the other constructor is not.

uportal-war/src/main/java/org/jasig/portal/security/audit/UserLoginRegistry.java
```diff
@@ -0,0 +1,56 @@
...
+@Component
+public class UserLoginRegistry
+    implements IUserLoginRegistry {
+
+    private IUserLoginDao userLoginDao;
+
+    @Override
+    public void storeUserLogin(final String username, final ReadableInstant momentOfLogin) {
+
+        if (!"guest".equals(username)) {
```
timlevett: There is a variable in portal.properties that says what the guest user name is.

uportal-war/src/main/java/org/jasig/portal/security/audit/dao/jpa/MostRecentLoginRecordingDao.java
```diff
@@ -0,0 +1,82 @@
...
+        if (null != storedMostRecentUserLogin) {
+
+            // remove the old most recent login so it can be replaced with the new one
+            // which has the same natural ID key (username).
+            getEntityManager().remove(storedMostRecentUserLogin);
+            getEntityManager().flush();
+        }
+
+        // this delete-then-insert strategy is naive and ridiculous.
+        // TODO: refactor to a create-or-update strategy.
```
timlevett: Agreed.

timlevett:
overall :+1: with the things in shortcomings being cleaned up.

uportal-war/src/main/webapp/WEB-INF/flows/user-manager/viewUserDetails.jsp
```diff
@@ -86,6 +86,32 @@
...
+                        </div>
+                </c:when>
+                <c:otherwise>
+                    <div>
+                    <%-- TODO: Improve this message to use a placeholder to interpolate the
+                    (formatted) date rather than assuming prefixing the message is good enough --%>
+                    <spring:message code="most.recent.login.at"
+                                         text="This user most recently logged in at"/>
+                        <fmt:formatDate pattern="yyyy-MM-dd hh:mm:ss a z (E)"
+                             value="${viewedUserLastLoginTimestamp.toInstant().toDate()}"/>.</div>
```
timlevett: This may not be the time or place but JodaTime does have a JSP tag lib.

http://mvnrepository.com/artifact/joda-time/joda-time-jsptags

uportal-war/src/main/java/org/jasig/portal/events/LoginEvent.java
apetro: Yeah, `LoginEvent` was weirdly obnoxious in that it's a public class and in concept new things should be able to instantiate `LoginEvents` and fire them, and yet it was hard to instantiate.  The existing package-private constructor is effectively `protected`, because it takes as an argument a which is a `protected static class` in the super-class (`PortalEvent`).

I might be missing some kind of intended elegance here, but adding a public constructor that simply accepts the state that goes into a `LoginEvent` as constructor arguments, well, that seems a lot more reasonable, and makes it handy to instantiate these for unit testing.

drewwills:
Is Wisconsin using this code yet?  Are we closer to merging it?

apetro:
Nope, Wisconsin isn't using this code.  This is abandonware.  On review it turned out implementing this feature didn't seem to be very important.  It might resurface again someday.

PR #506, opened 2015-01-29
[UP-4364] Replace the WindowStateSettingsStAXComponent event filter that...

jhelmer-unicon:
... used to set the portlet to minimized with a new attribute source that returns a theme-specific window state instead of the actual window state for the portlet.  This eliminates the issue of the WindowStateSettingsStAXComponent actually changing the window state on the portlet when it intends to only temporarily set the window state for that one render.

https://issues.jasig.org/browse/UP-4364

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/rendering/FixedWindowStateAttributeSource.java
+++ b/uportal-war/src/main/java/org/jasig/portal/rendering/FixedWindowStateAttributeSource.java
@@ -1,0 +1,53 @@
+package org.jasig.portal.rendering;
+
+import org.jasig.portal.layout.om.IStylesheetDescriptor;
+import org.jasig.portal.layout.om.IStylesheetParameterDescriptor;
+import org.jasig.portal.portlet.PortletUtils;
+import org.jasig.portal.portlet.om.IPortletWindow;
+import org.springframework.beans.factory.annotation.Required;
+
+import javax.portlet.WindowState;
+import javax.servlet.http.HttpServletRequest;
+
+
+/**
+ * Custom attribute source that ignore the actual window state of a portlet
+ * and uses the window state specified in the theme, if present.
+ *
+ * This is intended for the JSON layout theme where it puts all the portlets
+ * in minimized mode to reduce render cost.   It previously tried to set the
+ * windowState on the PortletWindow object and then used a flag to disable
+ * writing the window state to the DB.  Because the PortletWindow is
+ * cached in the session, it would still break the layout for
+ * each subsequent render in that session though.  New approach is to not
+ * modify the PortletWindow and just customize the channel XML tag attributes
+ * without ever updating the PortletWindow or persisting the windowState
+ * change.
+ *
+ * See UP-4364
+ *
+ * @author Josh Helmer, jhelmer.unicon.net
+ */
+public class FixedWindowStateAttributeSource extends PortletWindowAttributeSource {
+    private StylesheetAttributeSource stylesheetAttributeSource;
+
+
+
+    @Required
+    public void setStylesheetAttributeSource(StylesheetAttributeSource stylesheetAttributeSource) {
+        this.stylesheetAttributeSource = stylesheetAttributeSource;
+    }
+
+    @Override
+    protected WindowState getWindowState(HttpServletRequest request, IPortletWindow window) {
+        final IStylesheetDescriptor stylesheetDescriptor = stylesheetAttributeSource.getStylesheetDescriptor(request);
+
+        final IStylesheetParameterDescriptor defaultWindowStateParam = stylesheetDescriptor.getStylesheetParameterDescriptor("dashboardForcedWindowState");
+        if (defaultWindowStateParam != null) {
+            WindowState windowState = PortletUtils.getWindowState(defaultWindowStateParam.getDefaultValue());
+            return windowState;
+        }
+
+        return super.getWindowState(request, window);
+    }
+}
--- a/uportal-war/src/main/java/org/jasig/portal/rendering/PortletWindowAttributeSource.java
+++ b/uportal-war/src/main/java/org/jasig/portal/rendering/PortletWindowAttributeSource.java
@@ -91,7 +91,7 @@ public class PortletWindowAttributeSource implements AttributeSource, BeanNameAw
         final Collection<Attribute> attributes = new LinkedList<Attribute>();
         
         //Add window state data
-        final WindowState windowState = portletWindow.getWindowState();
+        final WindowState windowState = getWindowState(request, portletWindow);
         final Attribute windowStateAttribute = xmlEventFactory.createAttribute("windowState", windowState.toString());
         attributes.add(windowStateAttribute);
         
@@ -121,5 +121,20 @@ public class PortletWindowAttributeSource implements AttributeSource, BeanNameAw
         }
         
         return cacheKeyBuilder.build();
+    }
+
+
+    /**
+     * Get the window state for the given IPortletWindow object.
+     *
+     * Extension point.  Used by the FixedWindowStateAttributeSource to render a theme specific
+     * window state instead of the actual window state specified by the portletWindow object.
+     *
+     * @param request the HTTP request
+     * @param window the portlet window to get the state for
+     * @return the window state to use for 'window'
+     */
+    protected WindowState getWindowState(HttpServletRequest request, IPortletWindow window) {
+        return window.getWindowState();
     }
 }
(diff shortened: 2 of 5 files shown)


uportal-war/src/main/java/org/jasig/portal/rendering/FixedWindowStateAttributeSource.java
```diff
@@ -0,0 +1,53 @@
...
+ * each subsequent render in that session though.  New approach is to not
+ * modify the PortletWindow and just customize the channel XML tag attributes
+ * without ever updating the PortletWindow or persisting the windowState
+ * change.
+ *
+ * See UP-4364
+ *
+ * @author Josh Helmer, jhelmer.unicon.net
+ */
+public class FixedWindowStateAttributeSource extends PortletWindowAttributeSource {
```
apetro: Could this be named something more descriptive than `Fixed` ?  I'm struggling to give this class a pithy name, but maybe something more like `ThemeDefaultWindowStateHonoringAttributeSource` to emphasize that what it is doing is honoring the default window state specified by the theme (and using the goofy word "Honoring" to avoid saying something stronger like "Setting" that would imply a state change on the underlying `IPortletWindow`.

uportal-war/src/main/java/org/jasig/portal/rendering/PortletWindowAttributeSource.java
```diff
@@ -122,4 +122,19 @@ public final CacheKey getCacheKey(HttpServletRequest request, HttpServletRespons
...
+    /**
+     * Get the window state for the given IPortletWindow object.
+     *
+     * Extension point.  Used by the FixedWindowStateAttributeSource to render a theme specific
+     * window state instead of the actual window state specified by the portletWindow object.
+     *
+     * @param request the HTTP request
+     * @param window the portlet window to get the state for
+     * @return the window state to use for 'window'
+     */
```
apetro: An `@since uPortal 4.2` or so would be nice here.

uportal-war/src/main/resources/logback.xml
```diff
@@ -146,11 +146,11 @@
   </logger>
   -->
   
-  <!-- Uncomment to see the XML at various stages in the rendering pipeline
+  <!-- Uncomment to see the XML at various stages in the rendering pipeline -->
```
apetro: This is going to add a lot of noise to the default logs, isn't it?  Maybe better to leave this readily available in the comment but turned off by default?

uportal-war/src/main/java/org/jasig/portal/rendering/FixedWindowStateAttributeSource.java
```diff
@@ -0,0 +1,53 @@
...
+    @Required
+    public void setStylesheetAttributeSource(StylesheetAttributeSource stylesheetAttributeSource) {
+        this.stylesheetAttributeSource = stylesheetAttributeSource;
+    }
+
+    @Override
+    protected WindowState getWindowState(HttpServletRequest request, IPortletWindow window) {
+        final IStylesheetDescriptor stylesheetDescriptor = stylesheetAttributeSource.getStylesheetDescriptor(request);
+
+        final IStylesheetParameterDescriptor defaultWindowStateParam = stylesheetDescriptor.getStylesheetParameterDescriptor("dashboardForcedWindowState");
```
apetro: Given that the JSON theme has its own rendering pipeline, is it worth the complexity to go get this forced dashboard window state from the theme stylesheet descriptor, or would it be healthier and more to the point to get this from a `defaultWindowState` JavaBean property and have the XML be something more like

```
<bean class="org.jasig.portal.rendering.WindowStateDefaultingAttributeSource">
  <property name="defaultWindowState" value="MINIMIZED"/>
</bean>
```

That is, configure that default right in the `jsonRenderingPipelineContext.xml` which is specific to the JSON rendering pipeline anyway.

apetro:
Seems like an opportunity for a JUnit test. :smile:

:+1: for merge.  Fix the bug.  Better is better.

uportal-war/src/main/java/org/jasig/portal/rendering/FixedWindowStateAttributeSource.java
jhelmer-unicon: Ack.   Agreed.   Will rename.

uportal-war/src/main/java/org/jasig/portal/rendering/PortletWindowAttributeSource.java
jhelmer-unicon: Agreed.   Fixing.

uportal-war/src/main/resources/logback.xml
jhelmer-unicon: Grr.   I meant to omit this from the commit.   Will revert.

uportal-war/src/main/java/org/jasig/portal/rendering/FixedWindowStateAttributeSource.java
jhelmer-unicon: In retrospect, I think I agree.   Reading the state from the theme seems unnecessarily complicated since the only use-case right now -- I believe -- is the json rendering.   Easier to just do as you suggest.   Will update.

uportal-war/src/main/resources/logback.xml
jameswennmacher: @apetro As always, **great** attention to detail!  I appreciate that about your reviews.  Glad you caught this.

uportal-war/src/main/java/org/jasig/portal/rendering/FixedWindowStateAttributeSource.java
jameswennmacher: @apetro great idea.  Simpler is better.

uportal-war/src/main/java/org/jasig/portal/rendering/WindowStateAttributeSource.java
```diff
@@ -0,0 +1,49 @@
...
+ * each subsequent render in that session though.  New approach is to not
+ * modify the PortletWindow and just customize the channel XML tag attributes
+ * without ever updating the PortletWindow or persisting the windowState
+ * change.
+ *
+ * See UP-4364
+ *
+ * @since 4.2
+ *
+ * @author Josh Helmer, jhelmer.unicon.net
```
jameswennmacher: Just curious,  wouldn't this normally be yourId@unicon.net? Trying to avoid spam bots?

uportal-war/src/main/java/org/jasig/portal/rendering/WindowStateAttributeSource.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+public class WindowStateAttributeSource extends PortletWindowAttributeSource {
+    private WindowState windowState;
+
+
+    public void setWindowState(final WindowState windowState) {
+        this.windowState = windowState;
+    }
```
jameswennmacher: Seems to me that this class is worthless without having a window state set.  I suggest adding @Required and removing the null check in getWindowState.

jameswennmacher:
:+1:  Looks pretty good.  

Since it is no longer used (via configuration), I suggest also removing the default stylesheet setting at https://github.com/Jasig/uPortal/blob/uportal-4.2.0-M1/uportal-war/src/main/data/default_entities/stylesheet-descriptor/JsonLayout.stylesheet-descriptor.xml#L31.

I have a question.  It looks like the regular pipline processing is still using WindowStateSettingsStAXComponent (affects mUniversality, Universality, and Respondr).  However only mUniversality and JSON use this override/set window state feature (the latter no longer true with your change).  
* Is there any way mUniversality behavior could cause persistent errors similar to what you are describing in the issue?  I don't think so since you can't switch themes AFAIK during a session and I've never noticed an issue when selecting a theme at login (via an URL parameter) or selecting a theme when impersonating a user, but I thought worth asking.
* Could the issue you described also be responsible for https://issues.jasig.org/browse/UP-4216?

References:
* https://github.com/Jasig/uPortal/blob/uportal-4.2.0-M1/uportal-war/src/main/java/org/jasig/portal/rendering/WindowStateSettingsStAXComponent.java#L103 for WindowStateSettingsStAXComponent determining if theme has an override value and effectively setting the window state at https://github.com/Jasig/uPortal/blob/uportal-4.2.0-M1/uportal-war/src/main/java/org/jasig/portal/rendering/WindowStateSettingsStAXComponent.java#L151
* https://github.com/Jasig/uPortal/blob/uportal-4.2.0-M1/uportal-war/src/main/data/default_entities/stylesheet-descriptor/UniversalityMobile.stylesheet-descriptor.xml#L40

jhelmer-unicon:
Hey James.   Answers to your questions.

- Could mUniversality be affected?   I don't think so since you wouldn't be switching between views like you do in the desktop view.   In theory, I believe we could change muniversality to use this instead of the WindowStateSettingsStAXComponent, but that's untested.
- It is probably not responsible for UP-4216 although I suspect that they are similar.
- We could, but it's untested.   I'm pretty sure it would be fine though.

PR #507, opened 2015-02-02
[UP-4398] Hide the "Remove portlet" menu item if the portlet is marked a...

jhelmer-unicon:
...s unremovable or has a "dlm:deleteAllowed" value of false.

https://issues.jasig.org/browse/UP-4398

### DIFF
--- a/uportal-war/src/main/resources/layout/theme/respondr/content.xsl
+++ b/uportal-war/src/main/resources/layout/theme/respondr/content.xsl
@@ -411,7 +411,8 @@
       </xsl:if>
 
       <!-- Remove Icon -->
-      <xsl:if test="@unremovable='false' and not(//focused) and /layout/navigation/tab[@activeTab='true']/@immutable='false'">
+      <!-- note: deleteAllowed will either be false or not present if set from the admin ui -->
+      <xsl:if test="(@unremovable='false' and not(@dlm:deleteAllowed='false')) and not(//focused) and /layout/navigation/tab[@activeTab='true']/@immutable='false'">
         <!-- calls a layout api on click that removes the current node from the layout -->
         <li>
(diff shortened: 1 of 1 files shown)


uportal-war/src/main/resources/layout/theme/respondr/content.xsl
```diff
@@ -411,7 +411,8 @@
       </xsl:if>
 
       <!-- Remove Icon -->
-      <xsl:if test="@unremovable='false' and not(//focused) and /layout/navigation/tab[@activeTab='true']/@immutable='false'">
+      <!-- note: deleteAllowed will either be false or not present if set from the admin ui -->
+      <xsl:if test="(@unremovable='false' and not(@dlm:deleteAllowed='false')) and not(//focused) and /layout/navigation/tab[@activeTab='true']/@immutable='false'">
```
drewwills: I can't recall any meaning that the 'unremovable' attribute has had in recent uPortal history.  I don't believe it was read anywhere in Universality, and I don't think there's any way to manage it in the UI.

If that's the case, I'd be happier simply to ignore it here, to prevent confusion.

jameswennmacher: I verified there is no other reference in theme/*/*.xsl to that attribute so I also suggest removing it to avoid confusion.  I updated https://wiki.jasig.org/display/UPM41/Setup+a+fragment+layout+by+modifying+fragment-layout+XML+files indicating it is obsolete as well.  There are some code references to it, but they should be deleted.  I see you created https://issues.jasig.org/browse/UP-4399 so I updated it to also remove "immutable"

jhelmer-unicon: I've got some additional work that's not quite ready to commit that basically deprecates the "unremovable" flag.  Unfortunately, it's trickier than I expected it to be so I couldn't quite get it finished in the time I had allotted.   I filed a new bug at:  UP-4399 and if I can find the time I hope to finish that up. I had planned on removing this then.   For now, just concerned about this breaking existing configurations (eg. reimporting existing portlet-definition files).

drewwills:
Merged out-of-band.

PR #511, opened 2015-02-04
UP-4403 : Adds json export feature

timlevett:
With this you can now do `/api/{entity-type}/{entity-id}?format=JSON` (example `/api/portlet-definition/portal-activity?format=JSON`) which will return a JSON representation of the entity file.  Import may come in the future, but this is a step in that direction. By not passing in a format it will give you XML (just like it is today).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -204,6 +204,7 @@
         <mockito.version>1.9.5</mockito.version>
         <net.oauth.version>20100527</net.oauth.version>
         <objenesis.version>2.1</objenesis.version>
+        <org.json.version>20090211</org.json.version>
         <org.springframework.webflow.version>2.3.2.RELEASE</org.springframework.webflow.version>
         <oro.version>2.0.8</oro.version>
         <persistence-api.version>1.0</persistence-api.version>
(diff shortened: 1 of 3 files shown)


drewwills:
How do the permissions work for this feature?

timlevett:
@drewwills - that is unchanged. See this line for what it currently does: https://github.com/timlevett/uPortal/blob/add_json_export/uportal-war/src/main/java/org/jasig/portal/rest/ImportExportController.java#L177

I'm creating another PR that has a more user facing call to get portlet info :) Stay tuned.

timlevett:
@drewwills - as promised : https://github.com/Jasig/uPortal/pull/512

timlevett:
Any reason not to merge this?

vertein:
:+1:

jameswennmacher:
+1

@timlevett RE your 2/4 comment about another PR more user-facing for portlet info, was that done?  I'm not sure what you mean by user-facing and I was wondering how it might overlap with https://github.com/Jasig/uPortal/pull/541

timlevett:
@jameswennmacher - user facing meaning a normal user might use it, vs this where its more for admins.

#512 doesn't have much to do with #514 .

PR #512, opened 2015-02-04
UP-4404 : Add in fetch single portlet

timlevett:
+ Adds 3 new fields to a `LayoutPortlet` object. staticContent, pithyContent, and isAltMaxUrl. We are using static content to display simple content portlets in an Angular display. We are using pithy content to display on a dashboard view. The Alt max URL flag is used to determine some behavior. See [our AngularJS usage](https://github.com/UW-Madison-DoIT/angularjs-portal/blob/master/angularjs-portal-home/src/main/webapp/js/layout/layout-controller.js#L17) for more details.
+ Created new call `/api/portlet/{fname}.json` that returns a `LayoutPortlet` JSON object.
+ Permission on this is based on if the user can render that portlet.

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/rest/LayoutRESTController.java
+++ b/uportal-war/src/main/java/org/jasig/portal/rest/LayoutRESTController.java
@@ -16,6 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+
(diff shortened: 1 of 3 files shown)


drewwills:
Tested locally, works gorgeously!  :+1:

PR #519, opened 2015-02-14
UP-4412: Fixed issue with links in portal-global not being clickable in ...

doodelicious:
When uPortal is in mobile Bootstrap view (viewport < 768px), any links in the portal-global div is no longer clickable. This is because the div is floating and "hides" behind any other div that is absolute (such as Bootstrap rows and columns).

I added appropriate Bootstrap classes to make portal-global a row and removed absolute styling so when it collapses, links don't fold under other rows.

uportal-header-link-issue


drewwills:
Sounds great!

drewwills:
Looks great -- tested locally.  :+1:

PR #520, opened 2015-02-19
UP-4414 Allow Google to index the guest page

jameswennmacher:
https://issues.jasig.org/browse/UP-4414

Added ignoring configured User-Agent strings to the cookie check process so Universities desiring to have their guest pages indexed so their portal site shows up in Google can do so.

Defaults to allow Google to bypass the cookie check.


vertein:
Cool beans. :+1:

timlevett:
:+1:

PR #521, opened 2015-02-24
UP-4410 Don't query ldap for user attribute names in permissions check f...

jameswennmacher:
...or personDir search

https://issues.jasig.org/browse/UP-4410

Previously all possible person directory attribute values resulted in searches of groups and group membership for the permissions target. The latter resulted in a query of person directory per possible attribute to figure out what PAGS groups a target was in.  If the permissions target is not an entity, this is a waste of time.  Uses the PermissionsActivity entries (src/main/data/default_entities/permission_owner/* files) to filter out those targets that aren't entities and therefore can't be in groups.

### DIFF
--- a/uportal-war/src/main/data/default_entities/permission_owner/UP_USERS.permission-owner.xml
+++ b/uportal-war/src/main/data/default_entities/permission_owner/UP_USERS.permission-owner.xml
@@ -45,7 +45,7 @@
         <name>Edit user attribute</name>
         <fname>EDIT_USER_ATTRIBUTE</fname>
         <desc>Edit a specific user attribute.  This permission applies to that attribute for all editable users</desc>
-        <targetProvider>peopleAndGroupsTargetProvider</targetProvider>
+        <targetProvider>userAttributesTargetProvider</targetProvider>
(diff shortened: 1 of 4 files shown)


uportal-war/src/main/data/default_entities/permission_owner/UP_USERS.permission-owner.xml
```diff
@@ -45,7 +45,7 @@
         <name>Edit user attribute</name>
         <fname>EDIT_USER_ATTRIBUTE</fname>
         <desc>Edit a specific user attribute.  This permission applies to that attribute for all editable users</desc>
-        <targetProvider>peopleAndGroupsTargetProvider</targetProvider>
+        <targetProvider>userAttributesTargetProvider</targetProvider>
```
jameswennmacher: FYI I believe from the comment that the target is an attribute name, not a person who can edit all attributes, so that's why I changed this.  I.e. it is similar to VIEW_USER_ATTRIBUTE.

jameswennmacher:
Will invoke lazy consensus and merge by 3/5

PR #522, opened 2015-02-26
UP-4411 optimize person dir search

jameswennmacher:
https://issues.jasig.org/browse/UP-4411

When directory search gets results from a person directory search (primarily LDAP in most instances), code used to loop through all returned results and fetch each result from personDir.  If you searched for Thomas you may get several hundred results from LDAP and if you are not an admin user (so permission checks apply), each LDAP result is re-queried from personDir which would go out to LDAP or other external sources to get all the attributes associated with an entity prior to displaying.  Now:

1. limits the amount of re-querying personDir (to get all the attributes that are associated with an entity) to the search result max if displayName is present in all the search results (because the code wanted to display n results of a sorted list).  As I'm writing this I just realized that if users are initially found but dropped out because permission checks indicate you cannot view the user, the final end result may be fewer than the configured maxResults size.  However its a rare case and IMHO not worth worrying about.  Added comments accordingly.

2. For the remaining results, determines the list of visible users and attributes in a parallel fashion.  

It would be nice to do something different, such as don't check for permission to view a user if the principal doesn't have any DENY on VIEW_USER, but the default data set does have DENY on Fragment Owners and the system account so I don't see a good way around running each user through the permission check on whether to view a user (which causes PersonDirectory to do an LDAP or other query).


jameswennmacher:
Will invoke lazy consensus and merge by 3/5

PR #525, opened 2015-03-02
[WiP] UP-4075 Spring Framework 4.0.5 Upgrade

groybal:
The intent of upgrading to Spring Framework 4.0.5 is to gain Java 8 support (binary and feature).  Note that the latest 4.0.x version is currently 4.0.9, however, there was a change committed to 4.0.6 which has caused issues and must be addressed before the move to 4.0.9 (and in the future, 4.1.x) can be made.  Also, note that with the changes included here, there are two unit tests failing.  These have been commented out for now, but more investigation on these is required.   Details on these issues can be found here:  https://wiki.jasig.org/display/UPC/uPortal+-+Java+8+Upgrade

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -146,7 +146,7 @@
         <aspectjweaver.version>1.7.4</aspectjweaver.version>
         <casclient.version>3.2.2</casclient.version>
         <ccpp.version>1.0</ccpp.version>
-        <cernunnos.version>1.2.2</cernunnos.version>
+        <cernunnos.version>1.3.0-SNAPSHOT</cernunnos.version>
         <cobertura-maven-plugin.version>2.6</cobertura-maven-plugin.version>
         <commons-cli.version>1.2</commons-cli.version>
         <commons-codec.version>1.9</commons-codec.version>
@@ -203,7 +203,7 @@
         <mockito.version>1.9.5</mockito.version>
         <net.oauth.version>20100527</net.oauth.version>
         <objenesis.version>2.1</objenesis.version>
-        <org.springframework.webflow.version>2.3.2.RELEASE</org.springframework.webflow.version>
+        <org.springframework.webflow.version>2.4.0.RELEASE</org.springframework.webflow.version>
         <oro.version>2.0.8</oro.version>
         <persistence-api.version>1.0</persistence-api.version>
         <person-directory.version>1.6.0</person-directory.version>
@@ -213,7 +213,7 @@
         <resource-server.version>1.0.41</resource-server.version>
         <servlet-api.version>3.0.1</servlet-api.version>
         <slf4j.version>1.7.7</slf4j.version>
-        <spring-framework.version>3.2.9.RELEASE</spring-framework.version>
+        <spring-framework.version>4.0.5.RELEASE</spring-framework.version>
         <spring-ldap.version>1.3.1.RELEASE</spring-ldap.version>
         <spring-security.version>3.1.3.RELEASE</spring-security.version>
         <standard.version>1.1.2</standard.version>
@@ -381,6 +381,17 @@
                 <groupId>commons-httpclient</groupId>
                 <artifactId>commons-httpclient</artifactId>
                 <version>${commons-httpclient.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>commons-logging</groupId>
+                        <artifactId>commons-logging</artifactId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.httpcomponents</groupId>
+                <artifactId>httpclient</artifactId>
+                <version>${commons-httpcomponents.version}</version>
                 <exclusions>
                     <exclusion>
                         <groupId>commons-logging</groupId>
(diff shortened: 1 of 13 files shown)


pom.xml
```diff
@@ -146,8 +146,7 @@
         <aspectjweaver.version>1.7.4</aspectjweaver.version>
         <casclient.version>3.2.2</casclient.version>
         <ccpp.version>1.0</ccpp.version>
-        <cernunnos.version>1.2.2</cernunnos.version>
-        <cglib.version>3.1</cglib.version>
+        <cernunnos.version>1.3.0-SNAPSHOT</cernunnos.version>
```
jhelmer-unicon: Unfortunately, unreleased SNAPSHOT version breaks the build.  See travis-ci output:  https://travis-ci.org/Jasig/uPortal/builds/52943779

apetro: I think the build is trying to tell us something important here.  Far preferable to depend only upon real releases and not snapshots, since presumably any snapshot dependencies would need to be made concrete before uPortal release, and technical debt is to be avoided. :bank:

pom.xml
```diff
@@ -1402,12 +1411,14 @@
                                 <version>2.8</version>
                                 <configuration>
                                     <links>
+                                        <link>http://docs.oracle.com/javase/8/docs/api/</link>
                                         <link>http://docs.oracle.com/javase/7/docs/api/</link>
                                         <link>http://java.sun.com/javase/6/docs/api/</link>
                                         <link>http://java.sun.com/javaee/5/docs/api/</link>
```
apetro: Is there are reason to keep around links to the old JavaDoc versions, or should these old versions be retired out of these `<links>` with the introduction of newer versions?

pom.xml
jameswennmacher: Gary R is working with Drew to get an updated version of cernunnos published that will allow us to resolve this.

uportal-war/src/main/java/org/springframework/orm/jpa/JpaAccessor.java
```diff
@@ -0,0 +1,158 @@
...
+ * @since 2.0
+ * @see #setEntityManagerFactory
+ * @see #setEntityManager
+ * @see #setJpaDialect
+ * @see #setFlushEager
+ * @see JpaTemplate
+ * @see JpaInterceptor
+ * @see JpaDialect
+ * @deprecated as of Spring 3.1, in favor of native EntityManager usage
+ * (typically obtained through {@code @PersistenceContext})
```
apetro: What would it take to *not* draw `JpaAccessor` into uPortal source control and instead follow this advice to use the native `EntityManager`?

Alternatively, if forking this deprecated code into the uPortal source tree is the way to go, could this class benefit from some comments right here about why this deprecated class is included in the uPortal codebase?

timlevett: Seconded

jameswennmacher: Agreed.  At this point it is a time constraint.  Gary is tied up on a project with scrum deadlines to deal with (you know how it goes).

pom.xml
```diff
@@ -214,7 +213,7 @@
         <resource-server.version>1.0.41</resource-server.version>
         <servlet-api.version>3.0.1</servlet-api.version>
         <slf4j.version>1.7.7</slf4j.version>
-        <spring-framework.version>3.1.4.RELEASE</spring-framework.version>
+        <spring-framework.version>4.0.5.RELEASE</spring-framework.version>
```
mmoayyed: Wouldnt you want to upgrade to 4.1.6 instead?

pom.xml
```diff
@@ -204,7 +203,7 @@
         <mockito.version>1.9.5</mockito.version>
         <net.oauth.version>20100527</net.oauth.version>
         <objenesis.version>2.1</objenesis.version>
-        <org.springframework.webflow.version>2.3.2.RELEASE</org.springframework.webflow.version>
+        <org.springframework.webflow.version>2.4.0.RELEASE</org.springframework.webflow.version>
```
mmoayyed: 2.4.1 maybe?

pom.xml
timlevett: Agreed

jameswennmacher: As Gary R mentioned in the PR description, there was a change made in Spring Framework that breaks later versions of Spring Framework 3.1, 4.0, and all 4.1.  See the link in the description.  This PR is an interim step toward newer versions of Spring Framework that would at least get us to Java 8.

jameswennmacher:
Cernunnos updated so remaining work on this would be to ideally remove JpaAccessor and JpaInterceptor in favor of EntityManager-based approaches.  Maybe we can get this done post 4.2.0.

bpowell:
Even though the Spring version has been updated, the XML files are still pointing to old Spring xsd files. For example in [applicationContext.xml](https://github.com/Jasig/uPortal/blob/38a5858c67c4440d1a4e8555c5fb141ded13a06d/uportal-war/src/main/resources/properties/contexts/applicationContext.xml#L28) on line 28, Spring version 3.1 is actually being used. All the XML files need to be updated to use at bare minimum [beans-4.0](http://www.springframework.org/schema/beans/spring-beans-4.0.xsd)

drewwills:
I added [WiP] to the title of this PR because...

* There are merge conflicts
* It's not complete

My sense is that -- rather than being merged directly -- this PR will act as a guide for upgrading to Spring 4 after we perform some significant refactoring with a view towards uPortal 5.

drewwills:
In accordance with the last comment (4 mo. ago), I'm going to close this PR.

Upgrading to Spring Framework v4 is something that we need, but it will need to happen through a fresh effort.

PR #526, opened 2015-03-09
[UP-4422]  Update the travis-ci env to try the new docker-based virtuali...

jhelmer-unicon:
...zation (faster, more memory) and also cache the local maven repo so that deps don't need to be redownloaded with every build.

-- Note:  Removed caching config.   After playing with it for a while, I could not see any real improvement.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,9 @@
 language: java
+sudo: false
+cache:
+    directory:
+        - $HOME/.m2/repositories
+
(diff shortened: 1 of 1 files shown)


jhelmer-unicon:
This is very low risk as it only affects the travis build.   Merging without comments to see if this improves our overall travis build stability.   Should solve a lot of the "error code 137" errors we keep seeing in our builds.

PR #528, opened 2015-03-12
UP-4414 handle guest access no cookie support in browser

jameswennmacher:
https://issues.jasig.org/browse/UP-4414

Found that disabling cookies in the browser when user agent is set for Googlebot gets into an infinite HTTP 302 loop.  This change allows for cookies disabled in browser when user-agent is Googlebot or other configured user agent that skips cookie checks to render the guest page, at least with the default data set on my local desktop.  TBD if this actually allows googlebot to work properly.


jameswennmacher:
Found Googlebot's user agent string is not just 'Googlebot'.  Adjusted to regex pattern for user agent comparison.  Merging.

PR #529, opened 2015-03-16
UP-4424: Changed footer to be sticky; added top padding to header; removed styles from skin.less

doodelicious:
https://issues.jasig.org/browse/UP-4424

A number of small UI changes:

* Made the footer sticky (so if content height is less than the height of the viewport, the footer will always remain at the bottom of the viewport) (screenshot below)
* Moved anchor styles from styles.less into navigation.less
* Added top padding to portal-header (to pull notifications icon down from the edge) 

screenshot 2015-03-16 16 47 22

### DIFF
--- a/uportal-war/src/main/resources/layout/theme/respondr/respondr.xsl
+++ b/uportal-war/src/main/resources/layout/theme/respondr/respondr.xsl
@@ -744,10 +744,12 @@
                     </div>
                 </div>
                 <chunk-point/> <!-- Performance Optimization, see ChunkPointPlaceholderEventSource -->
-                <xsl:call-template name="footer.nav" />
-                <xsl:call-template name="region.footer.second" />
-                <xsl:call-template name="region.page-bottom" />
-                <xsl:call-template name="region.hidden-bottom" />
+                <div id="footer-container">
+                    <xsl:call-template name="footer.nav" />
+                    <xsl:call-template name="region.footer.second" />
+                    <xsl:call-template name="region.page-bottom" />
+                    <xsl:call-template name="region.hidden-bottom" />
+                </div>
                 <chunk-point/> <!-- Performance Optimization, see ChunkPointPlaceholderEventSource -->
             </div>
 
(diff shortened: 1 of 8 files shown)


timlevett:
:+1: . We did this at first, but ran into issues on small screens. Basically it makes the header/footer take up majority of the screen.  I wrote up a simplified version of our solution on jsfiddle. http://jsfiddle.net/ys3n2nna/

If you want to see what happens when you have a lot of portlets just add a bunch of `.box` divs.

duggan-cindy:
Thanks!

We are having this problem with favorites in the sidebar, especially if
there are too many links there.

Cindy Duggan
Manhattan College

doodelicious:
I've added a pull request which addresses the issue with the sidebar. It seems that the Bootstrap collapsible navbar has a default max-height property set, which was causing a problem for the containing div not to resize when new links were added to the 'Favorites' portlet. You can find it here:

https://github.com/Jasig/uPortal/pull/531

jameswennmacher:
@doodelicious if the portal content is larger than the vertical viewport size, is the footer pushed down so the content takes up the viewport or does the footer always cover the bottom of the screen?

doodelicious:
@jameswennmacher There isn't a problem with the content being longer than the viewport, as the properties of the footer are bottom: 0; position: absolute; which pushes the footer at the bottom of the wrapper by the following properties: min-height: 100%; position: relative;

However, your comment brought up another issue (slightly unrelated). The height of the uPortal footer is varied based on how many tabs and how many portlets each tab contains. This is a problem, because as of this writing there is no reasonable variable height sticky footer solution that does not involve jQuery or a pure CSS solution that [supports IE8+] (https://wiki.jasig.org/display/UPM41/Supported+Web+Browsers). A variable sticky footer solution is available using [CSS flexbox] (http://stackoverflow.com/questions/9126777/keep-footer-with-variable-height-on-bottom), but at this time IE11 is the only IE browser that supports flexbox.

The solution I wrote is for a fixed-height footer only; meaning, the footer height can be set using the LESS variable @footer-height, which is also used to set padding-bottom to the wrapper, clearing the space needed for the footer content.

My suggestion is to reject this solution for now. I will document the Jira ticket with this issue and once we update the browser support list, this is something that can be implemented in master. If other adopters prefer a fixed-height footer, this is a solution they can use to implement one.

drewwills:
A lesser but uncomplicated/uncontroversial fix was distilled from this PR and merged out-of-band.

PR #532, opened 2015-03-25
UP-4429 Use BROWSE to determine portlet visibility in Customize Gallery, search, apis

jameswennmacher:
https://issues.jasig.org/browse/UP-4429

Suggest this goes into uPortal 4.2.0 release since it includes a change to /api/portletList.  **This change will also require uPortal 4.1 and prior to review their data entities to add BROWSE permissions when upgrading to uPortal 4.2.0**

Change behavior of uPortal so it consistently uses BROWSE permission to determine if a portlet is available to a user in the Customize Gallery, Search results, and the /api/portletList API (used by Customize Gallery).  Also changed behavior of /api/portletList to include portlets with no categories in a Uncategorized category.

See Jira issue.  Previously there were inconsistencies in determining whether a user should see a portlet.  Now the various system aspects (gallery, Marketplace, Search, and /api/portletList) are consistent.

# Changes

## New category 'Uncategorized' in Customize Gallery

Portlets not in any category are put into a new category called Uncategorized.  For demonstration purposes, I removed the category from About.com College Life.

new category uncategorized-user

## Customize gallery now uses BROWSE permission.  Doesn't matter if portlets have a category or not (previously no category hid portlets from customize gallery).

gallery matches marketplace

One consequence of this is administrative users will see portlets in the gallery they did not used to since administrative users have BROWSE on all portlets.

admin uncategorized portlets

## Portlet search and Marketplace Search now respect BROWSE permission

Users doing search no longer see portlets they don't have access to

## Default/quickstart data updated

Corrects a few issues with portlets that user's shouldn't see being visible.


apetro:
Some of the changes feel unrelated, so I kind of wish they were in separate commits that were in separate pull requests.

apetro:
Since we can now grant `BROWSE` within the `portlet-definition.xml` files, maybe we should drop the `BROWSE`-granting `permission-set` files and just grant this permission within the `portlet-definition` files, ala the way `SUBSCRIBE` is in practice granted?

uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/ChannelListController.java
```diff
@@ -79,42 +82,50 @@
...
-	 * @param asXml
 	 * @return
 	 */
 	@RequestMapping(method = RequestMethod.GET)
 	public ModelAndView listChannels(WebRequest webRequest, HttpServletRequest request, @RequestParam(value="type",required=false) String type) {
-        type = type != null ? type : TYPE_SUBSCRIBE;
-        if(TYPE_MANAGE.equals(type)) {
+        if(type != null && TYPE_MANAGE.equals(type)) {
             throw new UnsupportedOperationException("Moved to PortletRESTController under /api/portlets.json");
         }
```
apetro: Should there be similar handling for `type=all`?

apetro:
Is none of this changeset amenable to coverage with unit testing?

apetro:
:+1: -- this is what `BROWSE` should have meant when it was introduced, and it's quite good to do this cleanup.

jameswennmacher:
RE unit testing, yes some of the change is amenable.  Unfortunately I ran over on time working on this and am backlogged on other work I need to get done so I decided to unfortunately punt on adding unit tests.  Sadly I looked for some to update and there were none either. :disappointed:

uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/ChannelListController.java
jameswennmacher: All was removed in 4/15/10 for UP-2672.  It was so far in the past I just removed it since no one would have been using it at this point.

jameswennmacher:
RE dropping BROWSE in permission-set, I don't think you want to do that.  It is still used for portlet categories; e.g. https://github.com/Jasig/uPortal/blob/uportal-4.2.0-M1/uportal-war/src/main/data/quickstart_entities/permissions_set/Everyone_BROWSE_SpecificPortlets-permission-set.xml.  I think that is pretty useful.

jameswennmacher:
Will merge Thu 4/2 if no objections.

PR #535, opened 2015-03-27
UP-4432:  Switch the default PAGS implementation from XML file config to...

drewwills:
... the database (JPA entity) flavor;  @depricate the legacy version

https://issues.jasig.org/browse/UP-4432


jameswennmacher:
I suggest also increasing maxActive in uPortal's databaseContext.xml file from 75 to 100.  I don't think we really need a 25 increase just due to this feature, but from load tests I've seen the 75 value is a little small so I think going to 100 as a default is not unreasonable at all.

jameswennmacher:
+1

drewwills:
@jameswennmacher -- done.

drewwills:
12 days on the queue -- time to merge.

PR #537, opened 2015-03-31
UP-4273 Search replaces ${portlet.title} with portlet's title in search results

jameswennmacher:
https://issues.jasig.org/browse/UP-4273

Allows search results in the title from portlets to include SpEL expression to replace the search result title with values from the portlet definition.  Primary use case is https://issues.jasig.org/browse/CMSPLT-52 so search results in CMS portlet have a title of the portlet definition, not "cms" as it previously was.  Works with earlier versions of SimpleCMSPortlet but does title substitution with v1.2.0-M3.

selection_114

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/portlets/search/SearchPortletController.java
+++ b/uportal-war/src/main/java/org/jasig/portal/portlets/search/SearchPortletController.java
@@ -80,6 +80,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.expression.spel.SpelEvaluationException;
 import org.springframework.expression.spel.SpelParseException;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.RequestMapping;
@@ -732,8 +733,9 @@ public class SearchPortletController {
             final SpELEnvironmentRoot spelEnvironment = new SpELEnvironmentRoot(portletDefinition);
             try {
                 result.setTitle(spELService.getValue(result.getTitle(), spelEnvironment));
-            } catch (SpelParseException e) {
-                result.setTitle("Invalid Spring EL expression '" + result.getTitle()+"'");
+            } catch (SpelParseException | SpelEvaluationException e) {
+                result.setTitle("(Invalid portlet title) - see details in log file");
+                logger.error("Invalid Spring EL expression {} in search result portlet title", result.getTitle(), e);
             }
         }
     }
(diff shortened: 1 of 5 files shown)


jameswennmacher:
Will merge on Fri 4/3 if no objections

jameswennmacher:
Note: Includes improved IPortalSpELService method that makes it easier for uPortal to use SpEL with objects other than those defined by default in PortalSpELServiceImpl (e.g. request and person).  I hope to see more extensive use of this new method in other areas of uPortal.

uportal-war/src/main/java/org/jasig/portal/portlets/search/SearchPortletController.java
```diff
@@ -719,12 +723,18 @@ private void addSearchResults(SearchResults portletSearchResults, PortalSearchRe
...
+                    portletWindowId);
             final IPortletEntity portletEntity = portletWindow.getPortletEntity();
             final IPortletDefinition portletDefinition = portletEntity.getPortletDefinition();
-            String portletTitle = portletDefinition.getTitle();
-            result.setTitle(portletTitle);
+            final SpELEnvironmentRoot spelEnvironment = new SpELEnvironmentRoot(portletDefinition);
+            try {
+                result.setTitle(spELService.getValue(result.getTitle(), spelEnvironment));
+            } catch (SpelParseException e) {
+                result.setTitle("Invalid Spring EL expression '" + result.getTitle()+"'");
```
apetro: Hmm.  Should the failure case be to show this message to users as the title of a search result?  Seems like maybe the failure should be logged and the search result should fall back on some kind of best-under-the-circumstances title ("(Unknown content title)")?

uportal-war/src/main/java/org/jasig/portal/spring/spel/IPortalSpELService.java
```diff
@@ -81,4 +81,15 @@
...
+
+    /**
+     * Parses an expression string against a provided SpEL environment.  The environment object should be a simple
+     * class that has objects and getXXXX methods where XXXX is the name of the object in the expression.
+     * For instance, to handle SpEL expressions of ${portlet.title} and ${user.name} you pass in an object
+     * that has getPortlet() and getUser() methods, assuming ${} is your expression token prefix and suffix.
+     * @param expressionString SpEL expression string to parse
+     * @param spelEnvironment environment context object with getter methods matching SpEL expression object names
+     * @return value of parsed expression
+     */
```
apetro: Worth an `@since` to document when this method came into the API?

apetro:
Can parts of this change be unit tested?

uportal-war/src/main/java/org/jasig/portal/portlets/search/SearchPortletController.java
jameswennmacher: I had considered that.  I thought it might be more helpful for the admin to easily see the error and not have to figure out he has to dig into a stacktrace to figure out what happened.  Since the typical case is that the value is from a portlet.xml preference and is not likely to be set it seemed reasonable to me at the time.

uportal-war/src/main/java/org/jasig/portal/spring/spel/IPortalSpELService.java
jameswennmacher: Good catch.

uportal-war/src/main/java/org/jasig/portal/portlets/search/SearchPortletController.java
jameswennmacher: Decided to incorporate your suggestion.  Thanks!

PR #538, opened 2015-03-31
[UP-4435]  Better handling for bad layout owners.   Rather than blowing u...

jhelmer-unicon:
...p the render, it should now log the error and the skip that fragment.

https://issues.jasig.org/browse/UP-4435


drewwills:
I just reviewed this PR with @jhelmer-unicon.  It's pretty straightforward.

PR #539, opened 2015-04-10
UP-4436 Groovy Script to convert XML PAGS to entity PAGS files

jameswennmacher:
https://issues.jasig.org/browse/UP-4436

Groovy can probably be improved or made more idiomatic.  I'd love to hear feedback on improvements.

Can invoke script 2 ways:

1. using Groovy (requires installing Groovy and having it on path)

groovy bootstrap/utilities/scripts/xmlPagsToEntityPags.groovy uportal-war/src/main/resources/properties/groups/PAGSGroupStoreConfig.xml uportal-war/src/main/data/my_entities/pags-group

Output example at https://issues.jasig.org/secure/attachment/14738/groovyExecutionResults.txt

2. via ant (ugh - not as nice, made script less clear, but done on request)

ant -Dmaven.test.skip=true -Ddir=uportal-war/src/main/data/quickstart_entities/pags-group pagsXmlToEntity

Output similar to groovy execution but wrapped in ant execution.

# Note

* As mentioned at end of output example, after running the groovy script you still have to either initdb, or import the entity files and run a generated SQL script to update existing PAGS group references in the DB.
* I debated creating an ant task to allow running the generated SQL script, but I think low value for the effort.

### DIFF
--- a/utilities/scripts/xmlPagsToEntityPags.groovy
+++ b/utilities/scripts/xmlPagsToEntityPags.groovy
@@ -1,0 +1,199 @@
+/**
+ * Licensed to Apereo under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work
+ * for additional information regarding copyright ownership.
+ * Apereo licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License.  You may obtain a
+ * copy of the License at the following location:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * This file is a groovy script that parses the PAGSGroupStoreConfig.xml and creates entity-PAGS files that
+ * can be imported into an existing uPortal database, creates a SQL script to update the references to the
+ * XML PAGS groups to entity-PAGS groups, and provides some instructions to the user on updating
+ * compositeGroupService.xml to switch from XML PAGS to entity PAGS.
+ *
+ * This script is best IMHO executed directly by groovy as:
+ * groovy xmlPagsToEntityPags.groovy pathToPAGSGroupStoreConfig.xml directoryToWriteEntityFilesTo
+ *
+ * However it can also be invoked via ant as:
+ * ant -Dmaven.test.skip=true -Ddir=directoryToWriteEntityFilesTo data-import
+ * in which case to this groovy script it will look as if invoked as
+ * groovy -s pathToscript xmlPagsToEntityPags.groovy pathToPAGSGroupStoreConfig.xml directoryToWriteEntityFilesTo
+ *
+ * @since 4.2.0
+ */
+
+import groovy.xml.MarkupBuilder
+import org.xml.sax.EntityResolver
+import org.xml.sax.InputSource
+import org.xml.sax.SAXException
+
+// Extra blank output in case script is invoked through ant to segregate it's output better
+println()
+println()
+
+// If groovy script is invoked from ant through uPortal's PortalShell, there will be 2 arguments
+// at the front we can ignore: -s scriptPath
+if (args.length != 2 && args.length != 4) {
+    println "Usage:  groovy xmlPagsToEntityPags.groovy pathToPAGSGroupStoreConfig.xml directoryToWriteEntityFilesTo"
+    println()
+    println "Typically:"
+    println "  pathToPAGSGroupStoreConfig.xml = ../../uportal-war/src/main/resources/properties/groups/PAGSGroupStoreConfig.xml"
+    println "  directoryToWriteEntityFilesTo = ../../uportal-war/src/main/data/myInstitution/pags-group"
+    println "        where myInstitution is your modified copy of the quickstart data set"
+    println()
+    println "Some entities will exist in uportal-war/src/main/data/default_entities/pags-group"
+    println "but those created in your myInstitution will override them"
+    System.exit(1)
+}
+
+def pagsFile = args[args.length == 2 ? 0 : 2]
+def entityDir = args[args.length == 2 ? 1 : 3]
+
+def xmlSlurper = new XmlSlurper()
+
+// Ignore the dtd specified in PAGSGroupStoreConfig.xml.  For some reason I wasn't able to get XmlSlurper
+// to resolve the dtd even when specifying its directory in the classpath.
+xmlSlurper.setEntityResolver(new EntityResolver() {
+    @Override
+    public InputSource resolveEntity(String publicId, String systemId)
+            throws SAXException, IOException {
+        if (systemId.contains("PAGSGroupStore.dtd")) {
+            return new InputSource(new StringReader(""));
+        } else {
+            return null;
+        }
+    }
+})
+// For some reason, must disallow doctype declarations when script is run in intelliJ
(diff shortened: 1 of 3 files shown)


jameswennmacher:
Will invoke lazy consensus on Wed 4/15.

utilities/scripts/xmlPagsToEntityPags.groovy
```diff
@@ -0,0 +1,199 @@
+/**
```
drewwills: @jameswennmacher et al. -- what do you think about moving utilities/ under bootstrap/ so we're not adding a new top-level folder to the repo for a single file?

utilities/scripts/xmlPagsToEntityPags.groovy
```diff
@@ -0,0 +1,199 @@
...
+ */
+
+import groovy.xml.MarkupBuilder
+import org.xml.sax.EntityResolver
+import org.xml.sax.InputSource
+import org.xml.sax.SAXException
+
+// Extra blank output in case script is invoked through ant to segregate it's output better
+println()
+println()
```
drewwills: I would probably do this with one line -- like maybe `print('\n\n');` -- but this approach is fine.

utilities/scripts/xmlPagsToEntityPags.groovy
```diff
@@ -0,0 +1,199 @@
...
+if (args.length != 2 && args.length != 4) {
+    println "Usage:  groovy xmlPagsToEntityPags.groovy pathToPAGSGroupStoreConfig.xml directoryToWriteEntityFilesTo"
+    println()
+    println "Typically:"
+    println "  pathToPAGSGroupStoreConfig.xml = ../../uportal-war/src/main/resources/properties/groups/PAGSGroupStoreConfig.xml"
+    println "  directoryToWriteEntityFilesTo = ../../uportal-war/src/main/data/myInstitution/pags-group"
+    println "        where myInstitution is your modified copy of the quickstart data set"
+    println()
+    println "Some entities will exist in uportal-war/src/main/data/default_entities/pags-group"
+    println "but those created in your myInstitution will override them"
```
drewwills: You can do this with...

```Java
println """
Usage:  groovy xmlPagsToEntityPags.groovy pathToPAGSGroupStoreConfig.xml directoryToWriteEntityFilesTo

Typically:
pathToPAGSGroupStoreConfig.xml = ../../uportal-war/src/main/resources/properties/groups/PAGSGroupStoreConfig.xml
[...]  
"""
```

utilities/scripts/xmlPagsToEntityPags.groovy
```diff
@@ -0,0 +1,199 @@
...
+            throws SAXException, IOException {
+        if (systemId.contains("PAGSGroupStore.dtd")) {
+            return new InputSource(new StringReader(""));
+        } else {
+            return null;
+        }
+    }
+})
+// For some reason, must disallow doctype declarations when script is run in intelliJ
+xmlSlurper.setFeature("http://apache.org/xml/features/disallow-doctype-decl", false)
```
drewwills: The comment says "must disallow," but the code says `disallow-doctype-decl=false` -- is there a way to reword for clarity?

Do we need to include special treatment for running this script in a vendor IDE when we have 2 ways to run it from the shell?

utilities/scripts/xmlPagsToEntityPags.groovy
```diff
@@ -0,0 +1,199 @@
...
+
+def pagsXml = xmlSlurper.parse(new File(pagsFile))
+
+// Create a map of the XML Pags Groups key to name values since entity pages uses only the group-name
+def pagsGroupNames = new HashMap();
```
drewwills: FWIW I'm given to understand that the Groovy community has moved away from the `def` keyword when the runtime type of the variable is known and unchanging.  Looks like this one is just a Map.

utilities/scripts/xmlPagsToEntityPags.groovy
```diff
@@ -0,0 +1,199 @@
...
+                                    'tester-class'(testItem.'tester-class'.text()
+                                            == "org.jasig.portal.groups.pags.testers.RegexTester" ?
+                                            "org.jasig.portal.groups.pags.testers.PropertyRegexTester" :
+                                            "org.jasig.portal.groups.pags.testers.PropertyInvertedRegexTester")
+                                    'test-value'("org.jasig.portal.http.header.userAgent.mobile.regex.pattern")
+                                } else {
+                                    'attribute-name'(testItem.'attribute-name'.text())
+                                    'tester-class'(testItem.'tester-class'.text())
+                                    'test-value'(testItem.'test-value'.text())
+                                }                                          pagsXmlToEntity
```
drewwills: What is the role of these characters?  `pagsXmlToEntity`

utilities/scripts/xmlPagsToEntityPags.groovy
```diff
@@ -0,0 +1,199 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+    writer.close()
+}
+
+println()
```
drewwills: As above, Groovy multi-line Strings are cool.

utilities/scripts/xmlPagsToEntityPags.groovy
```diff
@@ -0,0 +1,199 @@
...
+sqlWriter.writeLine ""
+pagsGroupNames.each { k,v ->
+    sqlWriter.writeLine "update UP_PERMISSION p set p.PRINCIPAL_KEY='pags.${v}' where p.PRINCIPAL_KEY='pags.${k}';"
+}
+sqlWriter.close()
+
+// If invoked via ant, entityDir was 'groovy-safed' via a macro.  Clean it up for user display.
+def targetDir = entityDir.replaceAll("\\/", "/").replaceAll("\\\\","\\")
+
+println()
```
drewwills: As above, Groovy multi-line Strings are cool.

utilities/scripts/xmlPagsToEntityPags.groovy
jameswennmacher: It's not really bootstrap or items to support the build, though after you suggested an ant task to run it as opposed to just invoking using groovy directly I can see arguments for saying it is there to support the build.  I can also see arguments for putting it under uportal-war/src/utilities or similar since it is uportal-specific and not global.  

I guess now that the build invokes it as a target, putting it under bootstrap is fine though that still bothers me as somewhat incongruent with the items in there (right now only ant-related library jar files).

utilities/scripts/xmlPagsToEntityPags.groovy
jameswennmacher: Ahh yes. I should have thought of that.  Thanks for bumping up my groovyness.

utilities/scripts/xmlPagsToEntityPags.groovy
jameswennmacher: Running the script in the IDE made debugging MUCH easier, especially for someone who isn't as solid on groovy like me since I could set breakpoints and inspect objects and figure out what they were, etc.  If there are alternative ways to debug using command-line tools I'd love to become more aware of them.

utilities/scripts/xmlPagsToEntityPags.groovy
jameswennmacher: Good to know.

utilities/scripts/xmlPagsToEntityPags.groovy
jameswennmacher: Argh.  Thanks for catching.  I didn't see that.  Slow IDE.  Occassionally I do a search or something and focus doesn't always switch to the find dialog before I've typed in the characters and they show up in the file instead.

utilities/scripts/xmlPagsToEntityPags.groovy
jameswennmacher: Interesting note (to me at least), intelliJ's autoComplete feature works great with the def, but does not consider it a map without it.

PR #540, opened 2015-04-10
UP-4439 - Updated mobile background container to point to Respondr

asgrant:
The background changing portlet broke for mobile devices using Respondr, it was using older muniversality settings.


drewwills:
@asgrant -- nice catch, thanks!

PR #546, opened 2015-05-06
UP-4456:  PAGS (JPA) -- Add constructors to IPersonTester implementation...

drewwills:
...s that accept a single IPersonAttributesGroupTestDefinition;  deprecate the ones that accept String,String

https://issues.jasig.org/browse/UP-4456

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/EntityPersonAttributesGroupStore.java
+++ b/uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/EntityPersonAttributesGroupStore.java
@@ -317,7 +317,7 @@ public class EntityPersonAttributesGroupStore implements IEntityGroupStore, IEnt
             TestGroup tg = new TestGroup();
             Set<IPersonAttributesGroupTestDefinition> tests = testGroup.getTests();
             for(IPersonAttributesGroupTestDefinition test : tests) {
-                IPersonTester testerInst = initializeTester(test.getTesterClassName(), test.getAttributeName(), test.getTestValue());
+                IPersonTester testerInst = initializeTester(test);
                 tg.addTest(testerInst);
             }
             groupDef.addTestGroup(tg);
@@ -330,17 +330,19 @@ public class EntityPersonAttributesGroupStore implements IEntityGroupStore, IEnt
             groupDef.addMember(member.getName());
         }
     }
-    private IPersonTester initializeTester(String tester, String attribute, String value) {
-          try {
-             Class<?> testerClass = Class.forName(tester);
-             Constructor<?> c = testerClass.getConstructor(new Class[]{String.class, String.class});
-             Object o = c.newInstance(new Object[]{attribute, value});
-             return (IPersonTester)o;
-          } catch (Exception e) {
-             logger.error("Error in initializing tester class: {}", tester, e);
-             return null;
-          }
-       }
+
+    private IPersonTester initializeTester(IPersonAttributesGroupTestDefinition test) {
+        try {
+            Class<?> testerClass = Class.forName(test.getTesterClassName());
+            Constructor<?> c = testerClass.getConstructor(new Class[]{IPersonAttributesGroupTestDefinition.class});
+            Object o = c.newInstance(new Object[]{test});
+            return (IPersonTester) o;
+        } catch (Exception e) {
+            logger.error("Error in initializing tester class: {}", test.getTesterClassName(), e);
+            return null;
+        }
+   }
+
     private Set<IEntityGroup> getContainingGroups(String name, Set<IEntityGroup> groups) throws GroupsException
     {
         logger.debug("Looking up containing groups for {}", name);
--- a/uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/EntityPersonAttributesGroupStore.java
+++ b/uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/EntityPersonAttributesGroupStore.java
@@ -318,6 +318,17 @@ public class EntityPersonAttributesGroupStore implements IEntityGroupStore, IEnt
             Set<IPersonAttributesGroupTestDefinition> tests = testGroup.getTests();
             for(IPersonAttributesGroupTestDefinition test : tests) {
                 IPersonTester testerInst = initializeTester(test);
+                if (testerInst == null) {
+                    /*
+                     * A tester was intended that we cannot now recreate.  This
+                     * is a potentially dangerous situation, since tests in PAGS
+                     * are "or-ed" together;  a functioning group with a missing
+                     * test would have a wider membership, not narrower.  (And
+                     * remember -- permissions are tied to groups.)  We need to
+                     * play it safe and keep this group out of the mix.
+                     */
+                    return null;
+                }
                 tg.addTest(testerInst);
             }
             groupDef.addTestGroup(tg);
--- a/uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/BaseAttributeTester.java
+++ b/uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/BaseAttributeTester.java
@@ -32,6 +32,9 @@ public abstract class BaseAttributeTester implements IPersonTester {
     protected final String attributeName;
     protected final String testValue;
 
+    /**
+     * @since 4.3
+     */
     public BaseAttributeTester(IPersonAttributesGroupTestDefinition definition) {
         super();
         attributeName = definition.getAttributeName();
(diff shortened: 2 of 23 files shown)


uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/BaseAttributeTester.java
```diff
@@ -26,48 +28,60 @@
  * @author Dan Ellentuck
  * @version $Revision$
  */
-
 public abstract class BaseAttributeTester implements IPersonTester {
-    protected String attributeName;
-    protected String testValue;
+    protected final String attributeName;
+    protected final String testValue;
```
apetro: Looks like `BaseAttributeTester` stops being immutable as far as it goes?

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/EagerRegexTester.java
```diff
@@ -55,7 +57,15 @@
  * @see RegexTester
  */
 public class EagerRegexTester extends RegexTester {
-    
+
+    public EagerRegexTester(IPersonAttributesGroupTestDefinition definition) {
```
apetro: Worth an `@since` for inline documentation of when this constructor entered the API?

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/BaseAttributeTester.java
drewwills: Stops being immutable or mutable?

I'm not sure subclasses should or will need to change the values of those member variables to do their work.  I tend to put `final` where I can.  But this part of the change is just something I threw in because I noticed they were never being re-set by subclasses... it's not a critical part of it.

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/EagerRegexTester.java
drewwills: Yep -- sounds good.  I intend to run through both this PR and the next one today looking for places to add @Since, @Deprecated (the annotation, not merely the javadoc bit), and more comments where appropriate.

uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/EntityPersonAttributesGroupStore.java
```diff
@@ -317,7 +317,7 @@ private GroupDefinition initGroupDef(IPersonAttributesGroupDefinition group) {
             TestGroup tg = new TestGroup();
             Set<IPersonAttributesGroupTestDefinition> tests = testGroup.getTests();
             for(IPersonAttributesGroupTestDefinition test : tests) {
-                IPersonTester testerInst = initializeTester(test.getTesterClassName(), test.getAttributeName(), test.getTestValue());
+                IPersonTester testerInst = initializeTester(test);
                 tg.addTest(testerInst);
```
jameswennmacher: Minor, but if testerInst is null because of an exception in initializeTester, the tester instance should not be added to tg.  It will cause a NPE in GroupDefinition.test.

drewwills: If initializeTester cannot complete successfully, I would think the whole group is invalid... and perhaps the portal should not operate at all.

There's probably some improvement we could make in the way this situation is communicated (viz. by an error).

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/PropertyInvertedRegexTester.java
```diff
@@ -37,10 +38,25 @@
...
+        final String propertyName = definition.getTestValue();
+        String regexExpression = PropertiesManager.getProperty
+                (propertyName, "");
+        if (StringUtils.isBlank(regexExpression)) {
+            logger.error("Unable to find property name {} in portal.properties or has empty value."
+                    + " PAGS PropertyInvertedRegexTester will always return true for attribute {}",
+                    propertyName, definition.getAttributeName());
+        }
+        setPattern(regexExpression);
+    }
```
jameswennmacher: Minor, but rather than duplicating lines 44 - 51, perhaps put them in a private method invoked by both constructors.

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/PropertyRegexTester.java
```diff
@@ -36,10 +37,25 @@
...
+        final String propertyName = definition.getTestValue();
+        String regexExpression = PropertiesManager.getProperty
+                (propertyName, "");
+        if (StringUtils.isBlank(regexExpression)) {
+            logger.error("Unable to find property name {} in portal.properties or has empty value."
+                + " PAGS PropertyRegexTester will always return false for attribute {}", propertyName,
+                definition.getAttributeName());
+        }
+        setPattern(regexExpression);
+    }
```
jameswennmacher: Same as PropertyInvertedRegexTester in regards to private method to avoid code duplication.

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/ThemeNameEqualsIgnoreCaseTester.java
```diff
@@ -48,6 +51,21 @@
...
+    public ThemeNameEqualsIgnoreCaseTester(IPersonAttributesGroupTestDefinition definition) {
+        final String themeTestValue = definition.getTestValue();
+        assert StringUtils.isNotBlank(themeTestValue);
+        this.themeTestValue = themeTestValue;
+        this.applicationContext = ApplicationContextLocator.getApplicationContext();
+        this.portalRequestUtils = applicationContext.getBean(IPortalRequestUtils.class);
+        this.profileMapper = applicationContext.getBean("profileMapper", IProfileMapper.class);
+        this.userLayoutStore = applicationContext.getBean("userLayoutStore", IUserLayoutStore.class);
+        this.stylesheetDescriptorDao = applicationContext.getBean("stylesheetDescriptorDao", IStylesheetDescriptorDao.class);
+    }
```
jameswennmacher: Again suggest private method or constructor chaining to avoid duplicating code.

jameswennmacher:
Overall, looks good.  I'm glad this was handled separate from https://github.com/Jasig/uPortal/pull/547 to keep it more digestible.  Good call!

uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/EntityPersonAttributesGroupStore.java
apetro: > perhaps the portal should not operate at all.

Eeep. Please no.  Shrill logging about something being horribly wrong, sure, by all means.  Causing the whole portal to fail, that's aggressively unfriendly.  Seldom is a production outage preferable to uptime of a gracefully degraded service.

timlevett: I think I agree with @apetro here. Check for null, log properly if is null, continue. Worst case is they are not part of a group they should be so they don't see something they should have access to.

drewwills: Okay @apetro & @timlevett  -- production outage may be too severe here.

But I still feel that the group should not be in service.  In fact the "worst case" -- in this case -- would be that suddenly you *are* a member of a group you shouldn't be.  Test groups are "or-ed" together, but tests are "and-ed" together.  If a test drops out of a test group that had multiple tests, the membership gets wider (not narrower).

jameswennmacher: Excellent point.  I agree with taking the entire group out of service.

bjagg: Perhaps a try-catch block around the code that defines the test group (lines 317 to 323)?

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/ThemeNameEqualsIgnoreCaseTester.java
drewwills: I updated the other 2 places where you made this observation, but I'm going to pass on this one.

* The other 2 places were doing some actual work;  this one is just assigning values to member variables, which is pretty natural in a constructor.
* Those fields are marked final -- and not even by me -- so i'd have to "un-final" them (which I'd prefer not to do) in order to set them in an init() method or somesuch.

jameswennmacher: I think the easiest thing to do is have ThemeNameEqualsIgnoreCaseTester(IPersonAttributesGroupTestDefinition definition) invoke ThemeNameEqualsIgnoreCaseTester(null, definition.getTestValue())

PR #547, opened 2015-05-07
Expand the nature of the data than can be defined in a PAGS <test> element...

drewwills:
with Set<String> includes and Set<String> excludes

https://issues.jasig.org/browse/UP-4458

This enhancement relates to [Pull Request #546](https://github.com/Jasig/uPortal/pull/546) and to Benito's work on the AdHocGroupsTester.


uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/jpa/PersonAttributesGroupTestDefinitionImpl.java
```diff
@@ -46,14 +51,14 @@
  * @author Shawn Connolly, sconnolly@unicon.net
  */
 @Entity
-@Table(name = "UP_PAGS_TEST")
+@Table(name = JpaPersonAttributesGroupTestDefinitionDao.TABLENAME_PREFIX)
```
apetro: Is it weird that the name of the table would be the TABLENAME  **PREFIX**  (instead of being the TABLENAME ?)

apetro:
No unit test coverage for this changeset?

uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/jpa/PersonAttributesGroupTestDefinitionImpl.java
drewwills: Yes, it's a little weird.

But the table is pre-existing, and I imagine it's safer to keep the name as-is.  I could make the prefix simply `UP_PAGS_`, but then all of the table names would start with `TEST`.

drewwills:
@apetro I could probably organize a sensible unit test for the postProcessPortalDataKey on the PortalDataKey impl that's a part of this changeset.  Most of the rest of it is getters & setters.

uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/jpa/PersonAttributesGroupTestDefinitionImpl.java
jameswennmacher: FWIW I would probably have the prefix UP_PAGS and use it for group, member, test, etc., but this works too.

drewwills: :+1:  I'm all for that, I just wasn't into that many files for this enhancement.  (And I didn't want to add several files to the delta for something so small.)  I support this change, FWIW.

uportal-war/src/main/resources/properties/contexts/importExportContext.xml
```diff
@@ -600,20 +615,6 @@
...
-                <entry key="SqlAttributes.DATA_SOURCE" value-ref="PortalDb" />
-                <entry key="SqlAttributes.TRANSACTION_MANAGER" value-ref="transactionManager" />
-                <entry key="isUserIncluded" value-ref="isUserIncluded" />
-                <entry key="jpaPersonAttributesGroupDefinitionDao" value-ref="jpaPersonAttributesGroupDefinitionDao" />
-                <entry key="jpaPersonAttributesGroupTestDefinitionDao" value-ref="jpaPersonAttributesGroupTestDefinitionDao" />
-                <entry key="jpaPersonAttributesGroupTestGroupDefinitionDao" value-ref="jpaPersonAttributesGroupTestGroupDefinitionDao" />
-            </util:map>
-        </property>
-    </bean>
-
```
jameswennmacher: Does this deletion mean we aren't doing export of pags-groups?

drewwills: Not at all.

There are 2 approaches in current use for implementing an export;  Shawn provided them both.  I'm deleting the unused (and "older") solution.

drewwills:
@apetro -- Unit test added

PR #548, opened 2015-05-08
UP-4443 adhoc pags groups

bjagg:
New PAGS Ad hoc group tester with an example group.


uportal-war/src/main/data/quickstart_entities/pags-group/Member_of_Students.pags-group.xml
```diff
@@ -0,0 +1,40 @@
...
+<pags-group script="classpath://org/jasig/portal/io/import-pags-group_v4-1.crn">
+  <name>Member of Students (PAGS)</name>
+  <description>Member of Students group</description>
+  <selection-test>
+    <test-group>
+      <test>
+          <tester-class>org.jasig.portal.groups.pags.testers.AdHocGroupTester</tester-class>
+          <includes>Students</includes>
+          <excludes>BannedStudents</excludes>
+          <excludes>SummerSessionStudents</excludes>
```
drewwills: I don't think we can check in (effectively) foreign-key references to groups that don't exist.  It would be confusing, even if it didn't actually break.

uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/jpa/PersonAttributesGroupTestDefinitionImpl.java
```diff
@@ -55,12 +56,12 @@
 @Table(name = JpaPersonAttributesGroupTestDefinitionDao.TABLENAME_PREFIX)
 @SequenceGenerator(
         name=JpaPersonAttributesGroupTestDefinitionDao.TABLENAME_PREFIX + "_GEN",
-        sequenceName="UP_PAGS_TEST_SEQ",
+        sequenceName=JpaPersonAttributesGroupTestDefinitionDao.TABLENAME_PREFIX + "_SEQ",
```
drewwills: Looks like this PR has my changes in it as well.  We can re-roll it with some Git magic to take out this "noise."

uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/jpa/PersonAttributesGroupTestDefinitionImpl.java
```diff
@@ -103,11 +108,11 @@ public PersonAttributesGroupTestDefinitionImpl(
 
     @ElementCollection(fetch=FetchType.EAGER)
     @CollectionTable(name=JpaPersonAttributesGroupTestDefinitionDao.TABLENAME_PREFIX + "_INCLUDES")
-    private Set<String> includes = Collections.emptySet();
+    private Set<String> includes = new HashSet<>(0);
```
drewwills: Good call.  This approach to defining `includes` and `excludes` is better than what I had -- works much better with the setters.

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * Immutable PAGS Tester for inclusive/exclusive membership in sets of groups. To avoid infinite recursion,
+ * calls to {@code test()} are tracked by parameters plus thread ID.
+ *
+ * @author  Benito J. Gonzalez <bgonzalez@unicon.net>
+ * @see     org.jasig.portal.groups.pags.IPersonTester
+ * @see     org.jasig.portal.groups.pags.dao.EntityPersonAttributesGroupStore
+ * @see     org.jasig.portal.groups.pags.dao.IPersonAttributesGroupTestGroupDefinition
+ * @since   4.3
```
drewwills: Nice job here.

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
```diff
@@ -0,0 +1,172 @@
...
+ */
+public final class AdHocGroupTester implements IPersonTester {
+
+    private static final Logger logger = LoggerFactory.getLogger(AdHocGroupTester.class);
```
drewwills: I'm not sure if slf4j (specifically) has weighed in on this topic, but generally -- iirc -- it has been deemed to be better to use non-static logging objects...
```
    private final Logger logger = LoggerFactory.getLogger(getClass());
```

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
```diff
@@ -0,0 +1,172 @@
...
+    private static final Set<String> currentPersons = new ConcurrentSkipListSet<>();
+    private final Set<IEntityGroup> includes = new HashSet<>();
+    private final Set<IEntityGroup> excludes = new HashSet<>();
+    private final String hashcode;
+
+    public AdHocGroupTester(IPersonAttributesGroupTestDefinition definition) {
+        Validate.notNull(definition.getIncludes());
+        Validate.notNull(definition.getExcludes());
+        collectGroups(definition.getIncludes(), this.includes, true);
+        collectGroups(definition.getExcludes(), this.excludes, false);
```
drewwills: For my money, I'd say...
```
    this.includes = collectGroups(definition.getIncludes(), true);
    this. excludes = collectGroups(definition.getExcludes(), false);
```
Is more idiomatic and readable.

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
```diff
@@ -0,0 +1,172 @@
...
+        hash.append("^");
+        for (IEntityGroup group : excludes) {
+            hash.append(group.getKey());
+        }
+        hash.append("_#");
+        return hash.toString();
+    }
+
+    @Override
+    public boolean test(IPerson person) {
```
drewwills: Suggest moving public methods above non-public in the source file.

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
```diff
@@ -0,0 +1,172 @@
...
+            }
+        }
+    }
+
+    /**
+     * Create a hashcode based on the includes/excludes groups.
+     *
+     * @return      hashcode for this instance
+     */
+    private String calcHashcode() {
```
drewwills: I suggest a different name for this concept, given that the Object class already has a `hashCode()` method -- this name could lead to confusion.

How about just using a UUID for this purpose?  You wouldn't need this method at all, just...
```
    this.uuid = UUID.randomUUID();
```
in the constructor.

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
```diff
@@ -0,0 +1,172 @@
...
+ * @author  Benito J. Gonzalez <bgonzalez@unicon.net>
+ * @see     org.jasig.portal.groups.pags.IPersonTester
+ * @see     org.jasig.portal.groups.pags.dao.EntityPersonAttributesGroupStore
+ * @see     org.jasig.portal.groups.pags.dao.IPersonAttributesGroupTestGroupDefinition
+ * @since   4.3
+ */
+public final class AdHocGroupTester implements IPersonTester {
+
+    private static final Logger logger = LoggerFactory.getLogger(AdHocGroupTester.class);
+    private static final Set<String> currentPersons = new ConcurrentSkipListSet<>();
```
drewwills: I suspect this collection doesn't want to be static.  I think you want to prevent re-entry on the pre-tester instance basis (not a global basis).

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     * Find {@link IEntityGroup} from group name.
+     *
+     * @param groupName     name of group to search from {@code GroupService}
+     * @return              {@code IEntityGroup} with given name or null if no group with given name found
+     * @see                 org.jasig.portal.services.GroupService
+     */
+    private IEntityGroup findGroupByName(String groupName) {
+        EntityIdentifier[] identifiers = GroupService.searchForGroups(groupName, GroupService.IS, EntityEnum.GROUP.getClazz());
```
drewwills: I suspect the 3rd argument should be `IPerson.class` -- or, alternately `EntityEnum.PERSON`

I'd rather not refer here to something (`EntityEnum`) in a package called `org.jasig.portal.portlets.groupselector` (to me it feels like coloring-out-of-the-lines) but I see a lot of code all over that does that already.  (Perhaps the enum should be moved.)

uportal-war/src/main/data/quickstart_entities/pags-group/Member_of_Students.pags-group.xml
bjagg: Good catch!

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
bjagg: The original design did not use a static collection. It assumed that the tester objects were reused. However, the current implementation of the group store creates a tester every time one is needed, so I changed this collection to static.

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
bjagg: Agree!

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
drewwills: That's surprising -- I would have thought PAGS group objects were fully-created (including testers) and held in memory... at least for a period.

This may be a good place to use EhCache (instead of a private collection).  We have more management options, including the ability to give entries a "time to live" (TTL), which could help us clean up entries that escaped removal for some reason.  (Untimely exception?)

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
bjagg: I will change this method to something like "calcGroupsHash".

Testing membership person membership will create several instances of testers that represent the same test.  When the thread recurses to the same AdHocGroupTest condition, it will create a second object that is different than others in the call stack. 

Once the PAGS group store is refactored to reuse testers, this would work.

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
bjagg: Nice catch! I pulled this snippet of code from another class and did not notice the package of EntityEnum.

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
bjagg: The method that creates all these instances is initializeTester in EntityPersonAttributesGroupStore. During my tests, tester classes are continuously being created. Using EhCache here would possibly speed up performance given all the object creation. Testers appear to be immutable, so good candidates for caching. Further, we should consider caching at the group level, but we still need to account for using the same AdHocGroupTester object when the parameters are exactly the same.

bjagg: Correction: if we cache the group, we do not need to special handling for AdHocGroupTester objects as they will be localized to their group. Having the same test object for a particular group membership test localizes the recursion issue to a single instance of AdHocGroupTester.

drewwills: That's a great find -- that these testers are continuously being created.  I feel like we should take advantage of this discovery by refactoring the surrounding code until they *are not* continuously created.  Smells like a good oportunity for improvement.

uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/jpa/PersonAttributesGroupTestDefinitionImpl.java
```diff
@@ -37,6 +37,7 @@
 import javax.persistence.TableGenerator;
 import javax.persistence.Version;
 
+import org.apache.commons.lang.StringUtils;
```
apetro: Prefer `lang3` when introducing the new import.  Sooner or later uPortal will stop depending upon commons lang < v3.

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
```diff
@@ -0,0 +1,172 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.jasig.portal.groups.pags.testers;
+
+import org.apache.commons.lang.Validate;
```
apetro: Prefer `lang3` for the new import.

apetro:
Any of this change set amenable to JUnit testing?

uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/jpa/PersonAttributesGroupTestDefinitionImpl.java
bjagg: Good to know -- will update

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
bjagg: Got it!

bjagg:
Not as it currently exists. I was reviewing the class for interesting code to test. The findX methods originated from other classes. (Good candidates for refactoring out a Group Service method?) A majority of the code is iterating the returned collections and adding them to a new collection or testing membership.

The calcGroupsHash method is a good candidate for a unit test as it performs a simple, testable action -- it can be refactored into a pure function. 

An interesting unit test is one that exercises recursion through the calls to isDeepMemberOf in the main test method. This will require a bit of mocking. I will check into this.

uportal-war/src/main/data/quickstart_entities/pags-group/Member_of_Students.pags-group.xml
```diff
@@ -0,0 +1,40 @@
...
+    under the License.
+
+-->
+<!--
+  Group member_of_students: "Member of Students (PAGS)"
+  All IPersons that are in the Students Group are members of this group
+-->
+<pags-group script="classpath://org/jasig/portal/io/import-pags-group_v4-1.crn">
+  <name>Member of Students (PAGS)</name>
+  <description>Member of Students group</description>
```
jameswennmacher: I suggest updating the comment, name, and description to be more clear; e.g. Member of Student that are not staff or access the site by mobile device.  This is really just to help non-developers understand this example.

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
```diff
@@ -0,0 +1,208 @@
...
+     *
+     * Also, if the person object is cached and used twice, then the group key with the ad hoc tester is not added to
+     * the containing group keys during the recursion but is added (or not) after the test call returns positive.
+     */
+    @Override
+    public boolean test(IPerson person) {
+        String personHash = person.getEntityIdentifier().getKey() + groupsHash + Thread.currentThread().getId();
+        logger.debug("Entering test() for {}", personHash);
+        if (currentTests.getQuiet(personHash) != null) {
+            logger.warn("Returning from test() for {} due to recusion for person = {}", personHash, person.toString());
```
jameswennmacher: Should this be a warning or info?  Does this mean the groups were setup recursively (user error which warrants a warning and more information to explain how the recursion occurred) or that is a natural aspect of this group tester having include groups (an info or debug because it is not a user error)?

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
```diff
@@ -0,0 +1,208 @@
...
+        IGroupMember gmPerson = findPersonAsGroupMember(person);
+        boolean pass = true;
+
+        Iterator<IEntityGroup> groups = includes.iterator();
+        while (pass && groups.hasNext()) {
+            IEntityGroup group = groups.next();
+            logger.debug("Checking if {} is a member of {}", person.getUserName(), group.getName());
+            if (!gmPerson.isDeepMemberOf(group)) {
+                logger.debug("!! {} is not a member of {}", person.getUserName(), group.getDescription());
+                pass = false;
```
jameswennmacher: I think this logic is wrong.  If person is not a member of 1st group, it won't go on to check the 2nd group.  I suggest default pass = false, loop through includes and if isDeepMemberOf=true, set pass to true and break out of loop.  If exit loop with pass=false, don't bother checking excludes.

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
```diff
@@ -0,0 +1,208 @@
...
+                logger.debug("{} is a member of {}", person.getUserName(), group.getDescription());
+            }
+        }
+        groups = excludes.iterator();
+        while (pass && groups.hasNext()) {
+            IEntityGroup group = groups.next();
+            logger.debug("Checking if {} is NOT a member of {}", person.getUserName(), group.getName());
+            if (gmPerson.isDeepMemberOf(group)) {
+                logger.debug("!! {} is a member of {}", person.getUserName(), group.getDescription());
+                pass = false;
```
jameswennmacher: For efficiency, can break at this point.  Additional exclude groups don't matter.

jameswennmacher: Actually, for greater efficiency I think you can check excludes list before includes.  If member of excludes list, can simply return false.  You don't need to check the includes list.  I assume excludes has priority over includes.

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
bjagg: Changed this to `debug` since it is expected during normal operations. Good catch, James!

uportal-war/src/main/data/quickstart_entities/pags-group/Member_of_Students.pags-group.xml
bjagg: Good points! I will clean this up to provide more clarity. Maybe a second example is also needed.

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
bjagg: The goal for a positive test result is for the person to be a member of every includes group and not a member of every excludes group. By starting with pass = true, the first time a membership check fails, the while loops are exited/skipped. The excludes while loop checks `pass`, so if an includes fails it is skipped.

I will add documentation.

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
bjagg: The `pass` in the while condition is the means to avoid further testing once a failed membership ship is found. The reason I fall through the rest of the method is to hit the cache remove call before returning.

uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/AdHocGroupTester.java
jameswennmacher: Got it.  That makes sense.  The code matches that definition.

jameswennmacher:
Thought I'd record this for general posterity and in case it triggers something I hadn't thought of by someone else.

I had some concern about caching implications because due to recursion, the 2nd time through the PAGS tester (the recursive call) returning false might have gone through a set of code that caches the result (with the PAGS tester returning false) that was incorrect.  True that as the recursion unwinds the cache would get overridden with with the correct result (the PAGS tester returning whatever is the right result).  However if caching was present and was a shared cache, there is a window of time where another thread might potentially read the 'incorrect' cached value.

I spent a good deal of time looking at the code.  In summary, I think this change does not cause caching problems. For details, continue reading.

The below are areas where caching is used internal to the method and the method I think can be in the calling chain.  I might be concerned with these caching the wrong result. 

## GroupMemberImpl.initializeContainingGroupKeys (and its subclasses)

This group member (person or group) cache is a set of the groups this member is in.  If you allow any nesting with ad-hoc groups (either in PAGS groups or via local groups), set up scenarios where you have nested groups to insure that for a Person A that the cache is not initialized with incorrect information.

I think this method is OK because the cache is local to the GroupMemberImpl instance.  I think code creates separate instances of these GroupMember subclass instances.

### Potential future optimizations

The above observation makes me think that uPortal might improve performance by using some form of shared cache instead of instance-based cache in this method.  I believe '''getCompositeGroupService().findContainingGroups(this)''' is probably an expensive call and making it multiple times seems quite negative.

However care would need to be taken to using a shared cache because there might be use cases where a member might have different results based on context.  An example of this is https://issues.jasig.org/browse/UP-3860 and some changes being considered for uMobile to communicate to uPortal for that session the user is accessing uPortal as a uMobile device which would affect PAGS groups.  Wrote this observation and idea up as https://issues.jasig.org/browse/UP-4465.

## EntityGroupImpl.contains(IGroupMember)

I think this cache is OK as it is only caching the group's direct group and person members without concern for nesting.

## ReferenceIndividualGroupService.findContainingGroups(IGroupMember gm)

Caches results from getGroupStore().findContainingGroups(gm).  I think the cache in this method is OK since it is adding groups to a global cache.  Even if an add was incorrectly false the first time through (the recursive call), it would be added later on.  If another thread had executed when the add had not occurred, it would do its own evaluation and add later on so both threads would get the correct results.

## ReferenceGroupService.findContainingGroups(IGroupMember gm)

Caches results from getGroupStore().findContainingGroups(gm).  I think the cache in this method is OK since it is adding groups to a global cache.  Follows the same rules as ReferenceIndividualGroupService.findContainingGroups(IGroupMember gm) above.

## AuthorizationImpl.primGetPermissionsForPrincipal(IAuthorizationPrincipal principal, String owner, String activity, String target)

For any permissions target (may be an iPerson, group, portlet, category, or various synthetic ALL permissions targets), caches the list of groups that contain that target.  This is a shared cache.

I don't think this would be a problem because the cache is not populated until the return from the call to  targetEntity.getAllContainingGroups(); e.g. this method is not internal to the calling chain.

### Potential optimization

It occurs to me that:

1. The cached data for a target such as a Person is exactly the same type of information GroupMemberImpl.initializeContainingGroupKeys and potentially other places in the code might benefit from having.  If it is being calculated multiple times and would yield the same information, caching it in one place would make a lot of sense.  Further it would help in terms of invalidating the cache easier when group information changes.

2. It might make sense if the groups code cached Person or Group to list of containing groups, and AuthorizationImpl (or perhaps a utility class it uses to assist with caching) cache the permissions-related information, but delegate to the groups caches for the person or Group targets.
Included this idea in https://issues.jasig.org/browse/UP-4465.

uportal-war/src/main/java/org/jasig/portal/portlets/groupadmin/AdHocPagsForm.java
```diff
@@ -0,0 +1,136 @@
...
+ * @author Benito J. Gonzalez <bgonzalez@unicon.net>
+ * @see     org.jasig.portal.groups.pags.dao.EntityPersonAttributesGroupStore
+ * @see     org.jasig.portal.groups.pags.dao.IPersonAttributesGroupTestGroupDefinition
+ * @see     org.jasig.portal.groups.pags.testers.AdHocGroupTester
+ * @since   4.3
+ */
+public final class AdHocPagsForm implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+    private static final String DESCRIPTION_BRANCH = "Collection of ad hoc groups";
```
drewwills: Relates to functionality that is no longer a part of the enhancement.

uportal-war/src/main/webapp/WEB-INF/flows/entity-selector/selectEntities.jsp
```diff
@@ -139,180 +126,280 @@
...
+
     <script type="text/javascript">
+        var setGroupDescription = function() {
+            var includes, excludes, description;
+            descInput = $("#groupDesc");
+            includes = $("li", "#${n}dataIncludesList").map(function () { return $(this).text(); }).get();
+            excludes = $("li", "#${n}dataExcludesList").map(function () { return $(this).text(); }).get();
+
+            if (includes.length + excludes.length == 0) {
+                $("#groupDesc").val("Collection of ad hoc groups");
```
drewwills: I guess maybe this should just be blank;  we're not managing container groups with this UI.

uportal-war/src/main/webapp/WEB-INF/flows/entity-selector/selectEntities.jsp
```diff
@@ -139,180 +126,280 @@
...
+            var tests = [];
+            $.each(groupNames, function (i, name) {
+                tests.push( { "testValue": name, "attributeName": testAttr,
+                              "testerClassName" : "org.jasig.portal.groups.pags.testers.AdHocGroupTester" } );
+            });
+            return tests;
+        };
+
+        var resetAdHocDialog = function () {
+            console.log("resetting ad hoc dialog");
```
drewwills: does console.log work on all browsers?

uportal-war/src/main/webapp/WEB-INF/flows/entity-selector/selectEntities.jsp
bjagg: Yep.

uportal-war/src/main/webapp/WEB-INF/flows/entity-selector/selectEntities.jsp
bjagg: Not sure. There was one in the original browseEntity() function. These should be stripped out.

drewwills:
Thanks @bjagg!

PR #549, opened 2015-05-14
UP-4460: Fixing contains() check in EntityPersonAttributeGroupStore#getContain…

bjagg:
…ingGroups() plus some refactoring.

1. Refactored out a method named getPagsGroupDefByName(String name) since the same code was used a three times. Also added better handling if return set was empty from dao call.

2. My IDE found a few lines of code where casting and such was not needed -- updated.

3. Reduced return statement code in findContainingGroupsForGroup() since a set object was always returned. No need to check for null.

4. Main change was to use the same object used for a contains() test as was potentially placed in the collection in the GetContainingGroups() method.

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/EntityPersonAttributesGroupStore.java
+++ b/uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/EntityPersonAttributesGroupStore.java
@@ -59,7 +59,7 @@ import org.springframework.context.ApplicationContext;
  * The Person Attributes Group Store uses attributes stored in the IPerson object to determine
  * group membership.  It can use attributes from any data source supported by the PersonDirectory
  * service.
- * 
+ *
  * @author Shawn Connolly, sconnolly@unicon.net
  * @since 4.1
  */
@@ -81,14 +81,14 @@ public class EntityPersonAttributesGroupStore implements IEntityGroupStore, IEnt
     public boolean contains(IEntityGroup group, IGroupMember member) {
         logger.debug("Checking if group {} contains member {}/{}", group.getName(), member.getKey(), member.getEntityType().getSimpleName());
         PagsGroup groupDef = convertEntityToGroupDef(group);
-        if (member.isGroup()) 
+        if (member.isGroup())
         {
            String key = ((IEntityGroup)member).getLocalKey();
            return groupDef.hasMember(key);
-        } 
-        else 
+        }
+        else
         {
-           if (member.getEntityType() != IPERSON_CLASS) 
+           if (member.getEntityType() != IPERSON_CLASS)
                { return false; }
            IPerson person = null;
            try {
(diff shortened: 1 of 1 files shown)


uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/EntityPersonAttributesGroupStore.java
```diff
@@ -354,20 +350,37 @@ private IPersonTester initializeTester(IPersonAttributesGroupTestDefinition test
...
+     *
+     * @param name      group name used to search for group definition
+     * @return          {@code IPersonAttributesGroupDefinition} of named group or null
+     * @see             IPersonAttributesGroupDefinitionDao#getPersonAttributesGroupDefinitionByName(String)
+     * @see             IPersonAttributesGroupDefinition
+     */
+    private IPersonAttributesGroupDefinition getPagsGroupDefByName(String name) {
+        Set<IPersonAttributesGroupDefinition> pagsGroups = personAttributesGroupDefinitionDao.getPersonAttributesGroupDefinitionByName(name);
+        if (pagsGroups.size() > 1) {
+            logger.warn("More than one PAGS group with name {} found.", name);
```
apetro: `warn` seems insufficiently dire for a violation of a fundamental assumption of the groups system.  `error` at the least?

I'm not sure returning what is effectively a random selection among the illegally multiple groups with the same name is the right behavior.  I think I'd rather it fail with an exception.  Not an exception that results in the entire portal falling over dead, of course, but an exception that at least fails the specific group operation being computed.  Multiple groups with the same name is, alas, not okay.

bjagg: I will test out what happens if a illegal state exception is thrown. If it does not break the portal, we can go with that change. Thanks Andrew!

uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/EntityPersonAttributesGroupStore.java
```diff
@@ -354,20 +350,37 @@ private IPersonTester initializeTester(IPersonAttributesGroupTestDefinition test
...
-        IPersonAttributesGroupDefinition pagsGroup = pagsGroups.iterator().next();
+        IPersonAttributesGroupDefinition pagsGroup = getPagsGroupDefByName(name);
         Set<IPersonAttributesGroupDefinition> pagsParentGroups = personAttributesGroupDefinitionDao.getParentPersonAttributesGroupDefinitions(pagsGroup);
-        for (IPersonAttributesGroupDefinition parent : pagsParentGroups) {
+        for (IPersonAttributesGroupDefinition pagsParent : pagsParentGroups) {
+            IEntityGroup parent = convertPagsGroupToEntity(pagsParent);
             if (!groups.contains(parent)) {
-                groups.add(convertPagsGroupToEntity(parent));
-                getContainingGroups(parent.getName(), groups);
+                groups.add(parent);
```
jameswennmacher: Ahh!  intelliJ highlighted the old code as suspicious!  I love intelliJ!  Good catch!

uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/EntityPersonAttributesGroupStore.java
bjagg: Looks like throwing an exception breaks the portal. I will change this to `error`.

jameswennmacher:
+1 on merge

timlevett:
:+1: Looks like merge conflicts tho

PR #552, opened 2015-05-22
UP-4473: Fix exclude filters and add component-scans where needed to …

bjagg:
…compensate

https://issues.jasig.org/browse/UP-4473

### DIFF
--- a/uportal-war/src/main/resources/properties/contexts/applicationContext.xml
+++ b/uportal-war/src/main/resources/properties/contexts/applicationContext.xml
@@ -34,16 +34,22 @@
     
     <context:annotation-config/>
     <context:component-scan base-package="org.jasig.portal">
-        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.portlets\."/>
-        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.security\.mvc\."/>
-        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.json\.rendering\."/>
-        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.layout\.dlm\.remoting\."/>
-        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.security\.remoting\."/>
-        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.rest\."/>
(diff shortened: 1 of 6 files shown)


jameswennmacher:
Thanks @bjagg.  In the future, please include a link to the Jira ticket in the description to make it easier to find.

drewwills:
This PR is still open, but the ticket linked at the top is marked "Resolved" -- we need to get on the same page.

bjagg:
I will re-open this. At the time, I did not know that Resolved means the change is merged into master in Apereo projects. (In my previous projects, Resolved meant ready for testing.)

uportal-war/src/main/resources/properties/contexts/applicationContext.xml
```diff
@@ -34,16 +34,22 @@
...
+        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.json\.rendering\..*"/>
+        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.layout\.dlm\.remoting\..*"/>
+        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.security\.remoting\..*"/>
+        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.rest\..*"/>
+        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.redirect\..*"/>
+        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.rendering\..*"/>
     </context:component-scan>
-    
+
+    <!-- Required for TenantService -->
```
drewwills: Ha!  I see what's happening here, and it adds up... uPortal is in need of some "services-layer" rework.

In many cases the "services" we have grew organically out of administrative portlets we were building.  We cobbled together an API inside an o.j.p.portlets.* package (impl package for an admin portlet), and it fell into use by other areas of the portal.  The result is that now we can't "fix" the bug in <context:component-scan>.

I think we should look briefly at where we can set up proper services in short order (maybe nowhere), and provide a comment explaining why we're crossing these boundaries everywhere else.

drewwills:
Closing this PR because a different version of this fix was committed here:  https://github.com/Jasig/uPortal/commit/22ffa571ecdaa4891508293ff29248d8c46f253d

That commit even references the same JIRA ticket, which is marked _Resolved_.

PR #553, opened 2015-05-28
Menu icon works with larger icons, doesn't collide

dcfreema:
Prevents the menu icon from moving over the uPortal logo on small mobile view.
The menu also works if a larger logo is chosen.
#UP-4475

oldmenu
oldmenu2
updatedmenu
updatedmenu2


timlevett:
:+1:

PR #554, opened 2015-06-01
Centered the uPortal Logo. Also separated logo and search bar

dmartign:
uPortal logo and search bar are each in its own row. Logo is centered.


funayman:
I think the title looks good when the browser is smaller and the title is center, but I think the title should be to the left when the browser window is bigger.

dmartign:
Closing in favor of #556

PR #555, opened 2015-06-04
UP-4483:  Add caching to CriteriaQuery usages inside JPA PAGS that we…

drewwills:
…re previously missing it

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/jpa/JpaPersonAttributesGroupDefinitionDao.java
+++ b/uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/jpa/JpaPersonAttributesGroupDefinitionDao.java
@@ -16,6 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+
 package org.jasig.portal.groups.pags.dao.jpa;
 
 import java.util.HashSet;
@@ -30,6 +31,7 @@ import javax.persistence.criteria.ParameterExpression;
 import javax.persistence.criteria.Root;
 
 import com.google.common.base.Function;
+
 import org.apache.commons.lang.Validate;
 import org.jasig.portal.groups.pags.dao.IPersonAttributesGroupDefinition;
 import org.jasig.portal.groups.pags.dao.IPersonAttributesGroupDefinitionDao;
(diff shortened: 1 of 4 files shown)


drewwills:
DO NOT MERGE YET!

There are a couple items I still want to look at when I get back from the conference.
  - Need to declare new caches for these queries
  - Look into use of @PortalTransactionalReadOnly and whether it would provide any benefit here

uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/jpa/JpaPersonAttributesGroupDefinitionDao.java
```diff
@@ -86,64 +117,37 @@ public void deletePersonAttributesGroupDefinition(IPersonAttributesGroupDefiniti
...
-        Set<IPersonAttributesGroupDefinition> parents = new HashSet<IPersonAttributesGroupDefinition>();
-        for (IPersonAttributesGroupDefinition parent : query.getResultList()) {
-            parents.add(parent);
-        }
-        return parents;
+        Set<IPersonAttributesGroupDefinition> rslt = new HashSet<IPersonAttributesGroupDefinition>(query.getResultList());
+        return rslt;
     }
 
-    @PortalTransactional
```
drewwills: Do we need @PortalTransactionalReadOnly here?  (I'm pretty sure we don't need @PortalTransactional).

uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/jpa/JpaPersonAttributesGroupDefinitionDao.java
```diff
@@ -86,64 +117,37 @@ public void deletePersonAttributesGroupDefinition(IPersonAttributesGroupDefiniti
...
-            groups.add(group);
-        }
+        Set<IPersonAttributesGroupDefinition> groups = new HashSet<IPersonAttributesGroupDefinition>(query.getResultList());
         if(groups.size() > 1) {
             logger.error("More than one PAGS Group found for name: {}", name);
         }
         return groups;
     }
 
-    @PortalTransactional
```
drewwills: Do we need @PortalTransactionalReadOnly here?  (I'm pretty sure we don't need @PortalTransactional).

uportal-war/src/main/java/org/jasig/portal/groups/pags/dao/jpa/JpaPersonAttributesGroupDefinitionDao.java
```diff
@@ -86,64 +117,37 @@ public void deletePersonAttributesGroupDefinition(IPersonAttributesGroupDefiniti
         entityManager.remove(persistentDefinition);
     }
 
-    @PortalTransactional
```
drewwills: Do we need @PortalTransactionalReadOnly here?  (I'm pretty sure we don't need @PortalTransactional).

PR #557, opened 2015-06-09
'You are signed in as' message doesn't appear on mobile view

dmartign:


drewwills:
@dmartign -- this issue seems to have sorted itself out.  I can see the greeting on every display size.

PR #560, opened 2015-06-16
UP-4492: Node data loading limited to grandchildren. Remove bug where …

bjagg:
…non-group children were loaded sometimes. Add group name validation.

I am not thrilled with the light red background when the group name is invalid. Any suggestions on a different indication to the user when the group name is invalid?

Also, should there be some text under group name label or a tool tip for the input?

### DIFF
--- a/uportal-war/src/main/webapp/WEB-INF/flows/entity-selector/selectEntities.jsp
+++ b/uportal-war/src/main/webapp/WEB-INF/flows/entity-selector/selectEntities.jsp
@@ -315,6 +315,19 @@
          * Ad Hoc Groups (jsTree)
          */
 
+        $("#groupName").on('input', function () {
+            var nameRegex = /^[\w ]{5,500}$/;
+            if (nameRegex.test($(this).val())) {
+                $("#${n}saveAdHocButton").removeAttr('disabled');
+                $(this).removeClass("invalid");
+            } else {
+                $("#${n}saveAdHocButton").attr('disabled', 'disabled');
+                $(this).addClass("invalid");
+            }
+        });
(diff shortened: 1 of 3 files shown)


uportal-war/src/main/webapp/WEB-INF/flows/entity-selector/selectEntities.jsp
```diff
@@ -315,6 +315,19 @@
...
          */
 
+        $("#groupName").on('input', function () {
+            var nameRegex = /^[\w ]{5,500}$/;
+            if (nameRegex.test($(this).val())) {
+                $("#${n}saveAdHocButton").removeAttr('disabled');
+                $(this).removeClass("invalid");
+            } else {
+                $("#${n}saveAdHocButton").attr('disabled', 'disabled');
+                $(this).addClass("invalid");
```
drewwills: Instead of defining the custom CSS class "invalid" (and associated CSS rules for it), why not use Bootstrap's `.has-error` class:  http://getbootstrap.com/css/#forms-control-validation

This approach would tie into (we hope) the rest of the portal's visual clues for web forms, and can be themed.

uportal-war/src/main/webapp/media/skins/common/less/entity-selector.less
```diff
@@ -366,4 +366,10 @@
         }
     }
 
+    .modal {
+        .invalid {
+            background-color: #fdd;
```
drewwills: See above.

drewwills:
Suggestion implemented;  merged out-of-band.

PR #561, opened 2015-06-19
UP-4497:  Rev the version of the /portletList REST API and add an opt…

drewwills:
…ional 'categoryId' parameter

https://issues.jasig.org/browse/UP-4497

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/ChannelListController.java
+++ b/uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/ChannelListController.java
@@ -157,11 +157,8 @@ public class ChannelListController {
      * specified category, including all descendants, and <em>excluding</em>
      * uncategorized portlets.  If no 'categoryId' is provided, this method
      * returns the portlet registry beginning with 'All Categories' (the root)
-     * and <em>including</em> uncategorized portlets.
-     * 
-     * Always returns the entire contents
-     * of the Portlet Registry, including uncategorized portlets,  to which the
-     * user has access.  Access is based on the SUBSCRIBE permission.
+     * and <em>including</em> uncategorized portlets.  Access is based on the
+     * SUBSCRIBE permission.
      *
      * @since 4.3
      */
--- a/uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/ChannelListController.java
+++ b/uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/ChannelListController.java
@@ -162,7 +162,7 @@ public class ChannelListController {
      *
      * @since 4.3
      */
-    @RequestMapping(value="/v4-3/dlm/portletRegistry", method=RequestMethod.GET)
+    @RequestMapping(value="/v4-3/dlm/portletRegistry.json", method=RequestMethod.GET)
     public ModelAndView getPortletRegistry(WebRequest webRequest, HttpServletRequest request,
             @RequestParam(value="categoryId", required=false) String categoryId) {
 
--- a/uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/ChannelListController.java
+++ b/uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/ChannelListController.java
@@ -20,7 +20,6 @@
 package org.jasig.portal.layout.dlm.remoting;
 
 import java.util.HashSet;
-import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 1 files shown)


uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/ChannelListController.java
```diff
@@ -131,78 +129,124 @@ public void setAuthorizationService(IAuthorizationService authorizationService)
...
+     */
+    @RequestMapping(value="/v4-3/dlm/portletRegistry.json", method=RequestMethod.GET)
+    public ModelAndView getPortletRegistry(WebRequest webRequest, HttpServletRequest request,
+            @RequestParam(value="categoryId", required=false) String categoryId) {
 
-        // get a list of all channels 
-        List<IPortletDefinition> allChannels = portletDefinitionRegistry.getAllPortletDefinitions();
-        Set<IPortletDefinition> uncategorizedPortlets = new HashSet<>(allChannels);
+        final PortletCategory rootCategory = categoryId != null
+                ? portletCategoryRegistry.getPortletCategory(categoryId)
```
apetro: Should this API return something interesting in the case where no category `categoryId` is found?  `404 Not Found`?

uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/ChannelListController.java
```diff
@@ -131,78 +129,124 @@ public void setAuthorizationService(IAuthorizationService authorizationService)
...
+     * uncategorized portlets.  If no 'categoryId' is provided, this method
+     * returns the portlet registry beginning with 'All Categories' (the root)
+     * and <em>including</em> uncategorized portlets.  Access is based on the
+     * SUBSCRIBE permission.
+     *
+     * @since 4.3
+     */
+    @RequestMapping(value="/v4-3/dlm/portletRegistry.json", method=RequestMethod.GET)
+    public ModelAndView getPortletRegistry(WebRequest webRequest, HttpServletRequest request,
+            @RequestParam(value="categoryId", required=false) String categoryId) {
```
apetro: Would there be an advantage in representing the `categoryId` in the path rather than as a request parameter?

`/v4-3/portletRegistry/{categoryId}/`, perhaps?

uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/ChannelListController.java
```diff
@@ -131,78 +129,124 @@ public void setAuthorizationService(IAuthorizationService authorizationService)
...
+     * If provided, this URL will return the portlet registry beginning with the
+     * specified category, including all descendants, and <em>excluding</em>
+     * uncategorized portlets.  If no 'categoryId' is provided, this method
+     * returns the portlet registry beginning with 'All Categories' (the root)
+     * and <em>including</em> uncategorized portlets.  Access is based on the
+     * SUBSCRIBE permission.
+     *
+     * @since 4.3
+     */
+    @RequestMapping(value="/v4-3/dlm/portletRegistry.json", method=RequestMethod.GET)
```
apetro: What is `dlm` about the registry of portlets?  Doesn't seem to be in the domain of layout management?

uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/ChannelListController.java
```diff
@@ -131,78 +129,124 @@ public void setAuthorizationService(IAuthorizationService authorizationService)
...
 
-    private Map<String,SortedSet<?>> getRegistry(WebRequest request, IPerson user) {
+    /**
+     * Updated version of this API.  Supports an optional 'categoryId' parameter.
+     * If provided, this URL will return the portlet registry beginning with the
+     * specified category, including all descendants, and <em>excluding</em>
+     * uncategorized portlets.  If no 'categoryId' is provided, this method
+     * returns the portlet registry beginning with 'All Categories' (the root)
+     * and <em>including</em> uncategorized portlets.  Access is based on the
+     * SUBSCRIBE permission.
```
apetro: Probably worth explicitly calling out here that if no `categoryId` is provided, portlets that the user has `BROWSE` on are represented in the response of this web service as if they were un-categorized portlets that the user does have `SUBSCRIBE` on even if the user does not actually have `SUBSCRIBE` on them.  That is, `Access is based on the SUBSCRIBE permission.` isn't the whole story.

PR #562, opened 2015-06-24
UP-4478:  Create api endpoint to return portlet content like exclusiv…

drewwills:
…e window state URL did

https://issues.jasig.org/browse/UP-4478

Also see:  http://jasig.275507.n4.nabble.com/uPortal-API-endpoint-for-returning-portlet-markup-aka-exclusive-window-state-alternative-for-Rich-JSe-td4665470.html

This PR doesn't have all the options discussed in the ticket or in the emails, but it has the basic feature.  I would be delighted for some of those additional capabilities to be added, but could use some collaboration to bring that about.

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/rest/PortletsRESTController.java
+++ b/uportal-war/src/main/java/org/jasig/portal/rest/PortletsRESTController.java
@@ -16,6 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+
 package org.jasig.portal.rest;
 
 import java.io.Serializable;
@@ -30,19 +31,25 @@ import javax.servlet.http.HttpServletResponse;
 import org.apache.commons.lang.StringUtils;
 import org.jasig.portal.EntityIdentifier;
 import org.jasig.portal.portlet.om.IPortletDefinition;
+import org.jasig.portal.portlet.om.IPortletWindow;
 import org.jasig.portal.portlet.om.PortletCategory;
 import org.jasig.portal.portlet.registry.IPortletCategoryRegistry;
 import org.jasig.portal.portlet.registry.IPortletDefinitionRegistry;
+import org.jasig.portal.portlet.registry.IPortletWindowRegistry;
+import org.jasig.portal.portlet.rendering.IPortletExecutionManager;
 import org.jasig.portal.rest.layout.LayoutPortlet;
 import org.jasig.portal.security.IAuthorizationPrincipal;
 import org.jasig.portal.security.IPerson;
 import org.jasig.portal.security.IPersonManager;
 import org.jasig.portal.services.AuthorizationService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.ResponseBody;
 import org.springframework.web.servlet.ModelAndView;
 
 /**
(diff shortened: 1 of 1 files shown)


uportal-war/src/main/java/org/jasig/portal/rest/PortletsRESTController.java
```diff
@@ -55,24 +62,22 @@
...
 
+    @Autowired
     private IPortletDefinitionRegistry portletDefinitionRegistry;
+
+    @Autowired
     private IPortletCategoryRegistry portletCategoryRegistry;
-    private IPersonManager personManager;
 
     @Autowired
-    public void setPortletDefinitionRegistry(IPortletDefinitionRegistry portletDefinitionRegistry) {
```
apetro: Worth keeping the public setter method in the spirit of maintaining the exposed public API of this public class?

uportal-war/src/main/java/org/jasig/portal/rest/PortletsRESTController.java
```diff
@@ -55,24 +62,22 @@
...
-        this.portletDefinitionRegistry = portletDefinitionRegistry;
-    }
+    private IPersonManager personManager;
 
     @Autowired
-    public void setPersonManager(IPersonManager personManager) {
```
apetro: Worth keeping the public setter method in the spirit of maintaining the exposed public API of this public class?

uportal-war/src/main/java/org/jasig/portal/rest/PortletsRESTController.java
```diff
@@ -55,24 +62,22 @@
...
-    }
+    private IPortletWindowRegistry portletWindowRegistry;
 
     @Autowired
-    public void setPortletCategoryRegistry(IPortletCategoryRegistry portletCategoryRegistry) {
-        this.portletCategoryRegistry = portletCategoryRegistry;
-    }
+    private IPortletExecutionManager portletExecutionManager;
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
```
apetro: Class is not `final`, so I'd go for `protected` here to make this `Logger` available to putative sub-classes.

apetro:
:+1:

uportal-war/src/main/java/org/jasig/portal/rest/PortletsRESTController.java
```diff
@@ -104,17 +107,60 @@ public ModelAndView getManageablePortlets(HttpServletRequest request, HttpServle
...
+            response.setStatus(HttpServletResponse.SC_FORBIDDEN);
+            return new ModelAndView("json");
+        }
+    }
+
+    /**
+     * Provides a single, fully-rendered portlet.  NOTE:  Access to this API
+     * enpoint requires only <code>IPermission.PORTAL_SUBSCRIBE</code> permission.
+     */
+    @RequestMapping(value="/v4-3/portlet/{fname}.html", method = RequestMethod.GET)
```
timlevett: Since this is the most recent version, thoughts on utilzing multiple URLS?

`@RequestMapping(value={"/v4-3/portlet/{fname}.html", "/portlet/{fname}.html"})` for example.  Then when we get another version we can just switch the base one to that one along with v4-10 or something.

Also not sure i like the dash in the url, but that is just an opinion.

uportal-war/src/main/java/org/jasig/portal/rest/PortletsRESTController.java
drewwills: That's a good question.  I get the sentiment or rationale, but I'd say it isn't.

 * This is a @Controller class;  it probably doesn't have direct callers/clients in uPortal (probably shouldn't)
 * The bean-style setters are noise in the file -- they're what I have to scroll through to see code I care about
 * One of the great lessons-learned in the evolution from channels to portlets was that it's tremendously difficult and limiting to hold out every single type and method signature in uPortal's own class loader as some kind of public API with a commitment to backwards compatibility;  we are at a point where if you have Java code that directly calls/implements/inherits from Java code distributed with uPortal under uportal-war/src/main/java, it should probably be contributed.  (Exceptions might be the pluggable interface-based abstractions like IPersonAttributesDAO.)

uportal-war/src/main/java/org/jasig/portal/rest/PortletsRESTController.java
drewwills: Gotcha!  I forgot to mark the class as final.  :smile:

uportal-war/src/main/java/org/jasig/portal/rest/PortletsRESTController.java
drewwills: I think I get what you're saying -- so there would always be an unversioned endpoint that points to the most recent incarnation of an API?

drewwills:
Thanks very much @apetro and @timlevett for pitching in with this process!

uportal-war/src/main/java/org/jasig/portal/rest/PortletsRESTController.java
timlevett: That was my thought, but just a thought, for the clean URL approach.

uportal-war/src/main/java/org/jasig/portal/rest/PortletsRESTController.java
apetro: When you go to write the JUnit test for this controller, are you going to wish there were these `public` setter methods intact so that you can inject Mockito mocks for the portlet definition registry and the person manager?

uportal-war/src/main/java/org/jasig/portal/rest/PortletsRESTController.java
drewwills: @timlevett -- I think the thought is kinda cool, but I'm concerned that it would produce unexpected issues in practice.  We're versioning & rev-ing these URLs precisely because the accepted parameters are changing and the outputs are changing (albeit subtly).  I think it would be very difficult to write a client of one of these APIs that would be resilient to these kinds of changes.  I think you want to tie your client to a specific, known set of inputs & outputs.

PR #563, opened 2015-06-24
added cas-filter-webapp to allow directories/files to be protected us…

prashanti19:
…ing CAS Authentication.


cas-filter-webapp/README.md
```diff
@@ -0,0 +1,10 @@
+#  CASifying directories/files
+
+This web application can be used to protect files/folders using CAS Authentication.
+The folder/files can be either be copied to the cas-filter-webapp or create a nfs mount or similar in teh deployed folder
```
apetro: Trivial : typo : `teh`

cas-filter-webapp/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+        <groupId>org.jasig.portal</groupId>
+        <artifactId>uportal-parent</artifactId>
+        <version>4.3.0-SNAPSHOT</version>
+    </parent>
+
+    <modelVersion>4.0.0</modelVersion>
+	<artifactId>cas-filter-webapp</artifactId>
+	<packaging>war</packaging>
+	<name>CAS Filter Web App</name>
+	<description>A CAS Client Application for Inside/TRAINING folder</description>
```
apetro: This `<description>` looks suspect.

jameswennmacher:
Overview.  This cas filter webapp was added by Texas A&M to have an external network share holding sets of documents they want protected by authentication so guests cannot access them, but no other security such as having some documents for students, others for faculty, others for staff.  Typically have  links to the documents within the portal.  I suggested contributing it as an optional add-on available to sites.

drewwills:
Love culture of sharing.  How many adopters do you think would use a feature like this one?  For something ancillary & optional, it feels a bit heavy-weight (own subproject, changes to build.xml).

drewwills:
Folks -- I'm going to go ahead and close this PR as out-of-scope.  It's a very reasonable thing to do, but it feels more like a module you add to your service, rather than a core component of uPortal.

PR #567, opened 2015-07-08
UP-4509: Casting groupId as a String

jonathanmtran:


timlevett:
:+1:

jameswennmacher:
+1

A deferred question is whether the GROUP_ID column should be a varchar. Certainly it should at least match the size of the same column in the GROUP table.

PR #568, opened 2015-07-16
UP-4512

chasegawa:
* Added method to the API for addFolder
* Added method to the API for updatingAttributes with JSON arrays in the request body of the REST call. Also supports setting structure attributes.
* Updated logic for move element (formerly move portlet) to allow inserting into a folder (nested folders).

https://issues.jasig.org/browse/UP-4512
Modify UpdatePreferenceServlet.java to support adding folders via the API.
Additionally, UI would should be able to effectively edit the structure-attributes associated with a folder; e.g. in particular
```
<structure-attribute>
<name>display</name>
<value>row</value>
</structure-attribute>
```
Create API method to allow setting structure-attributes on a specified id (it should be done automatically during the above create as well). Along with that, the ability to set all the folder attributes should be handled as well (separately rather than trying to take one list and figure out which ones belong to which attribute set).


uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/UpdatePreferencesServlet.java
```diff
@@ -308,30 +311,49 @@ public ModelAndView movePortletAjax(HttpServletRequest request,
...
      *
-     * @param ulm
      * @param request
      * @param response
      * @throws IOException
      * @throws PortalException
+     * @deprecated - replaced by the method/action "moveElement"
      */
     @RequestMapping(method = RequestMethod.POST, params = "action=movePortlet")
+    @Deprecated
```
timlevett: :+1: to deprecate vs change

timlevett:
:+1:

uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/UpdatePreferencesServlet.java
```diff
@@ -863,7 +882,133 @@ public ModelAndView addTab(HttpServletRequest request, HttpServletResponse respo
...
+    
+    /**
+     * Add a new folder to the layout.
+     * 
+     * @param request
+     * @param response
+     * @param targetId - id of the node to add the new folder to. By default, the folder will be inserted after other
+     *                   existing items in the node unless a siblingId is provided.
+     * @param siblingId - if set, insert new folder prior to the node with this id, otherwise simple insert at the end of the list.
+     * @param attributes - if included, parse the JSON name-value pairs in the body as the attributes of the folder.
```
jameswennmacher: Suggest indicating these override the default attributes placed on a folder.

uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/UpdatePreferencesServlet.java
```diff
@@ -863,7 +882,133 @@ public ModelAndView addTab(HttpServletRequest request, HttpServletResponse respo
 
         return new ModelAndView("jsonView", Collections.singletonMap("tabId", tabId));
     }
+    
+    /**
+     * Add a new folder to the layout.
+     * 
+     * @param request
+     * @param response
+     * @param targetId - id of the node to add the new folder to. By default, the folder will be inserted after other
```
jameswennmacher: For clarity: suggest indicating 'id of the folder node to add'...

uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/UpdatePreferencesServlet.java
```diff
@@ -863,7 +882,133 @@ public ModelAndView addTab(HttpServletRequest request, HttpServletResponse respo
...
+     * @param siblingId - if set, insert new folder prior to the node with this id, otherwise simple insert at the end of the list.
+     * @param attributes - if included, parse the JSON name-value pairs in the body as the attributes of the folder.
+     * e.g. :
+     * {   
+     *      "structureAttributes" : {"display" : "row", "other" : "another" },
+     *      "attributes" : {"hidden": "true", "type" : "header-top" }
+     * }
+     * 
+     * @return
+     */
```
jameswennmacher: This method's documentation is MUCH better than the others.  Thanks!!!

uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/UpdatePreferencesServlet.java
```diff
@@ -863,7 +882,133 @@ public ModelAndView addTab(HttpServletRequest request, HttpServletResponse respo
...
+            ulm.saveUserLayout();
+        } catch (Exception e) {
+            log.warn("Error saving layout", e);
+            return new ModelAndView("jsonView", Collections.singletonMap("error", getMessage("error.persisting.layout.change.folder", "Unable to add a new folder", locale)));
+        }        
+
+        Map<String, Object> model = new HashMap<>();
+        model.put("response", getMessage("success.add.folder", "Added a new folder", locale));
+        model.put("folderId", newFolder.getId());
+        model.put("immutable", newFolder.isImmutable());
```
jameswennmacher: You don't need to change the code, but FYI the immutable and unremovable attributes are somewhat obsolete.  I did note that the Respondr theme still uses them somewhat, but they have been replaced by the dlm:addChildAllowed, dlm:deleteAllowed, dlm:editAllowed, and dlm:moveAllowed . See https://wiki.jasig.org/display/UPM42/Setup+a+fragment+layout+by+modifying+fragment-layout+XML+files.

uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/UpdatePreferencesServlet.java
```diff
@@ -863,7 +882,133 @@ public ModelAndView addTab(HttpServletRequest request, HttpServletResponse respo
...
+            }
+        }
+    }
+
+    /**
+     * Update the attributes for the node. Unrecognized attributes will log a warning, but are otherwise ignored.
+     * 
+     * @param request
+     * @param response
+     * @param targetId
```
jameswennmacher: Please document what targetId can be; e.g. a folder node

uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/UpdatePreferencesServlet.java
```diff
@@ -863,7 +882,133 @@ public ModelAndView addTab(HttpServletRequest request, HttpServletResponse respo
...
+    public ModelAndView updateAttributes(HttpServletRequest request, 
+                                         HttpServletResponse response, 
+                                         @RequestParam("targetId") String targetId, 
+                                         @RequestBody Map<String, Map<String, String>> attributes) {
+        IUserLayoutManager ulm = userInstanceManager.getUserInstance(request).getPreferencesManager().getUserLayoutManager();
+
+        if (!ulm.getNode(targetId).isAddChildAllowed()) {
+            response.setStatus(403);
+            return null;
+        }
```
jameswennmacher: Is addChildAllowed relevant to updateAttributes?  I think checking permission dlm:editAllowed makes much more sense.

uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/UpdatePreferencesServlet.java
```diff
@@ -863,7 +882,133 @@ public ModelAndView addTab(HttpServletRequest request, HttpServletResponse respo
...
+    }
+
+    /**
+     * Update the attributes for the node. Unrecognized attributes will log a warning, but are otherwise ignored.
+     * 
+     * @param request
+     * @param response
+     * @param targetId
+     * @param attributes - parse the JSON name-value pairs in the body as the attributes of the element.
```
jameswennmacher: Please clarify as you did in addFolder that this is both attributes and structureAttributes and what the request body should look like (and mime type if application/json is needed to get the object mapping to work correctly)

uportal-war/src/main/java/org/jasig/portal/layout/dlm/remoting/UpdatePreferencesServlet.java
chasegawa: Technically the reflection doesn't care - all we really care about is that the layout manager gives us a node object. If the attributes don't match the type, a warning is logged, but no bad behavior happens.

jameswennmacher:
+1. Merging

PR #571, opened 2015-07-17
UP-4438 layout json v2 endpoint

jameswennmacher:
https://issues.jasig.org/browse/UP-4438

This PR replaces https://github.com/Jasig/uPortal/pull/541:
* changes endpoint to /uPortal/api/v4-3/dlm/layout.json
* Makes regions, favorites, and favoriteGroups consistent with tabs; e.g. a list of any combinations of portlets or folders of arbitrary depth
* Supports structure-attribute of display to cue UI for alternate layouts other than column; e.g. free-flowing row
* Supports content area with multiple rows, each having different layouts (e.g. one free-flowing row, one row of 2 columns, one row of 6 columns, one free-flowing row) for JS Respondr UI
* does NOT invoke the portlet rendering for portlets flowing through the JSON rendering pipeline (eliminates many stacktraces).  This fix applies to json layout v1 also (/uPortal/layout.json)

See Jira issue for examples of output, such as admin output at https://issues.jasig.org/secure/attachment/14845/layoutv43Admin.json.


jameswennmacher:
The only thing I see not right is when you become a fragment owner. Need to communicate that in a safe way to the JS UI.  Will do as separate commit or more likely separate Jira issue.

jameswennmacher:
Added fragmentAdmin flag, hostname, and authenticated flag.  To date only known issue is userImpersonating flag is always false.  Not sure it matters.

jameswennmacher:
Since this has been tested fairly well and there was some review on the previous PR, I'll merge this on Thu 7/23 if no other feedback.

timlevett:
:+1:

PR #574, opened 2015-07-30
UP-4525 : Handle errors around profile selection event publication.

apetro:
Fire the profile selection event safely, catching, logging, and ignoring exceptions that arise in response to the profile selection event, figuring that failing the entire login is an inappropriate escalation of a failure in honoring the user's expressed desire to use a particular profile.

(Safety added in response to an issue in MyUW where a glitch in the profile selection persistence code broke portal login for affected users, when those users would have been better off being able to log in even if not being able to switch among available profiles.)

See also [portal-dev@ thread](http://thread.gmane.org/gmane.comp.java.jasig.uportal.devel/6221).


timlevett:
:+1:

vertein:
:+1:

PR #578, opened 2015-08-27
UP-4537:  Improve performance of the Marketplace subsystem by REMOVIN…

drewwills:
…G the (additonal) permissions check for MANAGE and by creating far fewer AuthorizationPrincipal objects

ATTENTION WISCONSON:  This proposed change will alter the permissions necessary to view a portlet in the Marketplace.  The current implementation makes a second permissions check -- for MANAGE -- for every BROWSE check.  It will improve performance greatly if we can remove that.


vertein:
:+1:

PR #579, opened 2015-08-27
Fix for HTTP 415 (unsupported content type) issue with 'updateAttributes' API call

groybal:
This fix is related to changes made for UP-4512, where new API calls were added to UpdatePreferencesServlet.  For the updateAttributes(...) method, the HTTP request body is JSON that needs to be converted to a Map (@RequestBody Map<String, Map<String, String>> attributes).  This conversion was causing an HTTP 415 'unsupported content type' exception/response due to not having a JSON message converter configured.


jameswennmacher:
+1. Gary has tested with RespondrJS (JS-based Respondr-like front end). This fixes an error in the original enhancements.

jhelmer-unicon:
:+1:  Barring down-votes, will most likely try to deal with merge conflicts and merge this shortly.

jhelmer-unicon:
Resolved conflict and merged via directly

PR #582, opened 2015-09-09
UP-4517 Add documentation for Angular strategy

andrewstuart:
Add documentation and generator link for the angular uPortal strategy.


jameswennmacher:
Thanks for submitting this PR.  This is a strategy we are using to allow AngularJS portlets to cohabitate in uPortal.  It is nice to have the documentation in the repo instead of in confluence.

One suggestion before I +1:  the ngportal theme is not part of uPortal yet (that is the name Andrew uses for a Respondr-like theme in development that uses AngularJS for the rendering of most of the page content and some page regions).  Since this approach is not dependent upon having ngportal (it works with the regular Respondr theme), may I suggest moving it up one level in the directory hierarchy?

jameswennmacher:
Univ of Wisconsin-Madison team FYI on this PR.  We should probably re-engage on discussing the two AngularJS approaches to see if we can collaborate more effectively and insure AngularJS portlets work with both the approach you are using and the ngportal (or RespondrJS) theme or whatever we call it.

jameswennmacher:
Hmmm... at the uPortal root there is a docs folder.  Perhaps instead put this in there, maybe in a portletDevelopment directory.  It would be great to start transitioning some of the documentation into project markdown.

timlevett:
Would love to discuss @jameswennmacher . This looks promising for Angularified portlets.

andrewstuart:
@jameswennmacher, Alright, I moved it to the docs folder and renamed it so it's a little clearer.

One more option for in-repo documentation would be to make a jasig.github.io/uportal, using [Github Pages](https://pages.github.com/). It's as easy as making a static html site on a gh-pages branch (usually created with `git checkout --orphan gh-pages`).

Here's an example [site](https://andrewstuart.github.io/angular-drag-drop/#/pages/angular-drag-drop.directive:ngDrag) for ng-drag and its [gh-pages branch](https://github.com/andrewstuart/angular-drag-drop/tree/gh-pages).

timlevett:
I would second the approach @andrewstuart is suggesting with a gh page.

andrewstuart:
Alright, I made some changes after looking at some of the other implementations we've done.

One of the specific things that needed addressing was namespacing, and how to make sure everything works well if you want to use external javascript files.

Really, thus far the best thing to check out is [the generator](https://github.com/andrewstuart/generator-ng-portlet). I've tried it in both respondr and our respondrjs angularized version to make sure that the portlets it generates are compliant and don't trample each other.

I've also tried to make sure the documentation is updated for everything I've discovered/realized while writing it, and from looking at (for example) the Survey Portlet.

Let me know if the updates make sense.

timlevett:
:+1: good work! I like the generator. Neat stuff.

apetro:
If I give only one item of feedback, it's this:

I had a hard time understanding the bigger picture of this strategy from this documentation so far.

This looks like a document about tactics and technical practices for non-collision.  Tactics are very important.  Don't delete it.  But prepend something about strategy, about bigger picture, what approach is this trying to enable.

I think the strategy here is something like enabling JSR-286 portlets from various sources to use AngularJS directives and expressions and so forth in the markup they generate, and enabling the theme transform to also use AngularJS, and do this without fundamentally changing what makes up uPortal (the theme, and content from JSR-286 portlets interpolated into placeholders left by the theme.)  This would be stronger documentation if it first stated the strategy clearly, and then articulated the tactics and techniques that implement the strategy.

andrewstuart:
@timlevett, Thanks for the kind words :-)

@apetro Thanks for the feedback, and I definitely agree I could be clearer about the purpose. I've added a specific Abstract section to address this. I left out anything specific to JSR-286 because I'm still not very familiar with the full specification and so I'm not sure if this would cause any portlets to be non-compliant, though a brief text search of the document doesn't seem to say much regarding javascript.

PR #583, opened 2015-09-11
UP-4543 Upgrade to SimpleContentPortlet 2.0.1 (WARNING must export/import if have existing attachments)

jameswennmacher:
https://issues.jasig.org/browse/UP-4543

If you have existing attachments in the database, you must export them with v1.2.x and import them with the latest version. See https://wiki.jasig.org/display/PLT/Upgrading+from+1.2.x+or+prior+to+2.0.0 for instructions.


timlevett:
Since we are still bundling portlets, would the be a switch to 4.3.0 for the uPortal release (breaking API, requires big change, etc...)

timlevett:
:+1:

jameswennmacher:
Yes I was thinking this would be a 4.3.0 item.

PR #585, opened 2015-09-15
UP-4545: Update Maven dependency for portlet-api to newer 2.1 jar.

bjagg:


pom.xml
```diff
@@ -209,7 +209,7 @@
         <persistence-api.version>1.0</persistence-api.version>
         <person-directory.version>1.7.1-M1</person-directory.version>
         <pluto.version>2.1.0-M3</pluto.version>
-        <portlet-api.version>2.0</portlet-api.version>
+        <portlet-api.version>1.0</portlet-api.version>
```
jameswennmacher: Calling it portlet-api.version is misleading since it is artifact 1.0 but NOT portlet-api 1.0.  I suggest to rename this property to portlet-2-1-api.version or in this case perhaps remove it.

uportal-portlets-overlay/CalendarPortlet/pom.xml
```diff
@@ -49,9 +49,9 @@
         </dependency>
 
         <dependency>
-            <groupId>javax.portlet</groupId>
-            <artifactId>portlet-api</artifactId>
-            <version>2.0</version>
+            <groupId>org.apache.portals</groupId>
+            <artifactId>portlet-api_2.1.0_spec</artifactId>
+            <version>1.0</version>
```
jameswennmacher: Interesting.  I wonder if this is still needed for import.  I don't see it passed along to the data-import execution.

Likewise, I can't comment on it for some reason but I wonder if Announcements portlet really needs portlet-api as a dependency for its import/export.

pom.xml
bjagg: Great idea. Clarity is important.

uportal-portlets-overlay/CalendarPortlet/pom.xml
bjagg: I will check.

bjagg: Announcements does have a dependency on the artifact. Calendar did not, so removing that reference.

apetro:
:+1:

pom.xml
```diff
@@ -518,9 +517,9 @@
                 <version>${mail.version}</version>
             </dependency>
             <dependency>
-                <groupId>javax.portlet</groupId>
-                <artifactId>portlet-api</artifactId>
-                <version>${portlet-api.version}</version>
+                <groupId>org.apache.portals</groupId>
+                <artifactId>portlet-api_2.1.0_spec</artifactId>
+                <version>1.0</version>
```
drewwills: Strange choice, IMHO, on the part of the folks releasing this new API jar.  Since the meaning is stripped from the <version> number, I support the notion of taking it out of the <properties> section;  putting 1.0 there would be confusing.

PR #586, opened 2015-09-21
Add a REST endpoint for the current user details

andrewstuart:
This addition is something we needed in order to easily get details about the currently-logged-in user, and we thought it might be useful for the rest of the community as well. :-)


uportal-war/src/main/java/org/jasig/portal/rest/PeopleRESTController.java
```diff
@@ -102,4 +102,18 @@ public ModelAndView getPerson(@PathVariable String username,
...
+        final IPerson me = personManager.getPerson(request);
+
+        if ( me == null ) {
+            //If null, this person is not logged in.
+            response.setStatus(401);
+            return null;
+        }
+
+        return new ModelAndView("json")
+                .addObject("person", me.getAttributeMap());
```
jameswennmacher: Looks good.  For consistency with the other API responses in this class, consider changing to
`.addObject("person", new NamedPersonImpl(person.getUsername(), person.getAttributeMap());`
True it only adds an additional username attribute that returns your username which is probably not that necessary and nests the attributes one level deeper, but I am thinking it would be good for all of these methods to return an IPersonAttributes for consistency.

andrewstuart:
Alright, I made your suggested changes and actually threw in a simplified (for consumers) /people/v2 endpoint that has the same functionality but should greatly simplify client use.

A couple highlights:
- No more repeating your query parameter keys in searchTerms[]
- All /v2 endpoints return the object representing the resource requested, or an array if the resource requested has multiple values (not under {people: []})

Hopefully this is helpful :-)

uportal-war/src/main/java/org/jasig/portal/rest/PeopleRESTControllerV2.java
```diff
@@ -0,0 +1,118 @@
...
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+@Controller
+@RequestMapping("/people/v2")
```
jameswennmacher: Making the APIs easier to consume is a good thing.  +1 on merging and for your insight.  I have three minor requests first though:
* change the url to /v4-3/people to be consistent with how we are naming the endpoints (see https://github.com/Jasig/uPortal/blob/master/uportal-war/src/main/java/org/jasig/portal/rest/layout/LayoutJsonV43RenderingController.java#L70 as an example)
* Add a @since 4.3 to this class and to the added method above to clarify their entry into the API
* Add a bit of javadoc to the methods describing the output, perhaps a sample output (our APIs are sorely lacking javadocs explaining what the output looks like).

andrewstuart:
Do you think I should use javadoc or something more api-specific like JSONDoc? It seems like jsondoc would generate better API documentation but would mean adding another dependency. There is precedent for it, though, as it's used in the Survey Portlet.

jameswennmacher:
I think documenting the uPortal API with JSONDoc would be excellent and is an outstanding suggestion.  It's certainly easier and better than finding the Java class file and reading the (rarely present) comments. It would be a great improvement IMHO to the project to start moving us in that direction.  I'm interested in whether others in the community would have a different opinion.

I suggest publishing to an endpoint of /apidoc or something similar.

jameswennmacher:
BTW you'd add the JSONDoc bean to properties/contexts/servlet/mvcServletContext.xml and have a basePath of /doc.  That should give you /api/doc as the base path.  It may not work this way, but I'd try it.

uportal-war/src/main/java/org/jasig/portal/rest/PeopleRESTController.java
```diff
@@ -102,4 +103,18 @@ public ModelAndView getPerson(@PathVariable String username,
         return mv;
     }
 
+    @RequestMapping(value = "/people/me.json", method = RequestMethod.GET)
```
drewwills: Two questions:

* I have seen the word `self` used in REST APIs so commonly that I suspect it's idiomatic; do you have an objection to changing to `/people/self.json`?
* Do we really need to add this endpoint to both controllers?  Would it be enough to add it to the `PeopleRESTControllerV43`?

timlevett: I agree with @drewwills , this call should only exist in one spot.

uportal-war/src/main/java/org/jasig/portal/rest/PeopleRESTControllerV43.java
```diff
@@ -0,0 +1,124 @@
...
+        final IPersonAttributes person = lookupHelper.findPerson(searcher, username);
+        if (person == null) {
+            response.setStatus(HttpServletResponse.SC_NOT_FOUND);
+            return null;
+        }
+
+        return new ModelAndView("json", person.getAttributes());
+    }
+
+    @RequestMapping(value = "/me", method = RequestMethod.GET)
```
timlevett: Can we end this in `.json`?

andrewstuart: My vote is definitely no, for these reasons:
- .json is metadata about the resource and thus should go in the request headers; there's an explicit `Accept:` well-defined HTTP header for that.
- It would break the conventions in the rest of this controller; if I see `/me.json` and try `/andrewstuart.json`, I'm going to get a 404.
- We only return JSON anyway. Unless we add xml, yaml, or some other text format, adding `json` in such an obvious place is a bit misleading in my experience.

uportal-war/src/main/java/org/jasig/portal/rest/PeopleRESTController.java
andrewstuart: Pulled out of pre-4-3 now. `self` is a common idiom in REST APIs for a fully-qualified URL to the current resource. In most of the popular APIs I've used, `me` is the idiom I've seen used ([reddit](https://www.reddit.com/dev/api/) for one example; probably the first place I saw this).

I've just finished looking around at a few more popular REST apis. None of them (aside from Reddit) had either `me` or `self` endpoints. Nearly all of them were OAuth-authenticated now anyway, so of course you get that information from the oauth provider's `user_info` endpoint. But since uPortal isn't an OAuth provider (yet 😄) we should probably avoid that to avoid confusion.

Just give me a holler if you want me to change it to `/self`, but my vote is for `/me` given that it's already in use and won my ad-hoc popularity search by a paltry 1-0.

uportal-war/src/main/java/org/jasig/portal/rest/PeopleRESTControllerV43.java
drewwills: I don't find this argument compelling...

> It would break the conventions in the rest of this controller; if I see /me.json and try /andrewstuart.json, I'm going to get a 404.

because (1) the controller is entirely new with this commit (and we have the opportunity simply to add `.json` across the board);  and (2) omitting the `.json` is actually a "break" from the pattern we've been following with REST APIs.

BUT!  I do think the first point holds water, and (inspired by this discussion) I've been doing some googling & reading on the topic of "restful API naming conventions" and "rest api naming best practices."  My research suggests we're in the minority... perhaps an extremely slim minority.. when we add `.json` to our URIs.

When we began that practice, I think there was a real sense that we -- and other typical API designers -- would commonly offer a "/people/drewwills.json" side-by-side with a "/people/drewwills.xml."  That hasn't panned out in uPortal, or the industry more broadly.

So I feel I could easily be persuaded to flip this rule;  but I'd like for more folks to chime in, and -- if we change -- I'd believe we should **communicate** and **document** this choice effectively.

bjagg: Spring @RestController supports .json and .xml extensions dynamically. If Jackson annotations are applied to the data model served by the controller, then Spring handles the marshalling to the proper format. But it likely also supports ACCEPT: as well.

bjagg: After some contemplation, I suggest we start dropping the extensions, like ".json". There are a number of formats coming out which may challenge JSON in the future. I prefer our APIs  with well known URIs to respond intelligently rather than muddy them with variations for the same endpoint.

jhelmer-unicon: If it were my call, I'd prefer to omit the extension too.  Since we don't tend to use the \@RequestBody annotation, we can't fully take advantage of the spring goodness (eg. handle encoding based on Accept header and/or extension) it would be good to move in that direction, IMO.

vertein: :+1:

drewwills:
Since we reached a consensus, I...

* Created a JIRA (https://issues.jasig.org/browse/UP-4720)
* Updated the version numbers (4.3 is already out)
* Merged out-of-band

PR #588, opened 2015-10-01
Pass through -Denv on `ant data-import`

andrewstuart:
I've run into this a few times and it makes it quite painful to run a single import with a separate *.properties file. This should fix at least the data-import task.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -671,6 +671,7 @@
         </echo>
 
         <antcall target="up-shell">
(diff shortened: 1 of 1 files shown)


jameswennmacher:
This work should be obsoleted by https://issues.jasig.org/browse/UP-4591

andrewstuart:
Yeah, that's definitely a better solution. :-)

PR #589, opened 2015-10-13
UP-4459 browse permission ui

bjagg:
Change Portlet admin UI to support management of BROWSE permission.


bjagg:
https://issues.jasig.org/browse/UP-4459

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
```diff
@@ -95,13 +95,7 @@
...
 import org.jasig.portal.portlets.groupselector.EntityEnum;
 import org.jasig.portal.portlets.portletadmin.xmlsupport.IChannelPublishingDefinitionDao;
-import org.jasig.portal.security.AuthorizationPrincipalHelper;
-import org.jasig.portal.security.IAuthorizationPrincipal;
-import org.jasig.portal.security.IAuthorizationService;
-import org.jasig.portal.security.IPermission;
-import org.jasig.portal.security.IPermissionManager;
-import org.jasig.portal.security.IPerson;
-import org.jasig.portal.security.PermissionHelper;
+import org.jasig.portal.security.*;
```
drewwills: Normally we prefer not to use * imports, IIRC.  But I don't think the topic has come up recently in this dev community.

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
```diff
@@ -118,13 +112,14 @@
...
- * 
+ *
  * @author Jen Bourey, jbourey@unicon.net
  * @revision $Revision$
  */
 @Service
 public class PortletAdministrationHelper implements ServletContextAware {
-	protected final Log logger = LogFactory.getLog(PortletAdministrationHelper.class);
+
+    protected final Log logger = LogFactory.getLog(PortletAdministrationHelper.class);
```
drewwills: If the logger is non-static, it's convenient to use getClass().  (I realize this isn't a line that was added or changed -- other than whitespace -- in this delta.)

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
```diff
@@ -158,87 +153,48 @@
...
+                form.addCategory(new JsonEntityBean(cat));
+            }
 
-        // otherwise, if this is a new portlet, pre-populate the categories
-        // and groups with some reasonable defaults
+            try {
+                getGroupPermissions(def, form, IPermission.PORTLET_SUBSCRIBER_ACTIVITY);
+                getGroupPermissions(def, form, IPermission.PORTLET_BROWSE_ACTIVITY);
+            } catch (GroupsException e) {
+                e.printStackTrace();
```
drewwills: This catch {} block looks like what the IDE generates.  I think we're going to want to log this circumstance, and probably add a custom message that describes what was happening and what wen't wrong.

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
```diff
@@ -253,13 +209,71 @@ public PortletDefinitionForm getPortletDefinitionForm(IPerson person, String por
         return form;
     }
 
+    /*
+     * Add groups and permissions to form.
+     */
+    private void getGroupPermissions(IPortletDefinition def, PortletDefinitionForm form, String activity) {
```
drewwills: I suggest a new name for this method.  A method that starts with "get" should return something.  Also "get" is so intimately associated with beans & properties that (IMHO) it's probably better not to use "get" to name a method that has any real behavior.  Maybe something like "addGroupPermissionsToForm."

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
```diff
@@ -253,13 +209,71 @@ public PortletDefinitionForm getPortletDefinitionForm(IPerson person, String por
...
+        final AuthorizationService authService = AuthorizationService.instance();
```
drewwills: Looks like AuthorizationService is a Spring-managed bean dependency.  I suggest a member variable with @Autowired.

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
```diff
@@ -253,13 +209,71 @@ public PortletDefinitionForm getPortletDefinitionForm(IPerson person, String por
...
+
+            // first assume this is a group
+            IEntityGroup group = GroupService.findGroup(principal.getKey());
+            if (group != null) {
+                bean = new JsonEntityBean(group, EntityEnum.GROUP);
+            }
+
+            // if a matching group can't be found, try to find a matching
+            // non-group entity
+            else {
```
drewwills: Arrangement of whitespace & comments make these _appear to be_ 2 different if {} clauses.

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
```diff
@@ -253,13 +209,71 @@ public PortletDefinitionForm getPortletDefinitionForm(IPerson person, String por
...
+                IGroupMember member = authService.getGroupMember(principal);
+                bean = new JsonEntityBean(member, EntityEnum.PERSON);
+                String name = groupListHelper.lookupEntityName(bean);
+                bean.setName(name);
+            }
+
+            /* Make sure we capture the principal */
+            if (!form.getGroups().contains(bean)) {
```
drewwills: If the form can handle both groups & people (as principals), perhaps we should rename the getter & setter for the collection of principals it manages.

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
```diff
@@ -253,13 +209,71 @@ public PortletDefinitionForm getPortletDefinitionForm(IPerson person, String por
...
+                form.addGroup(bean);
+            }
+
+            /* Assumption: group names are unique in this context */
```
drewwills: What happens if we have a group and a person with the same name?  (Unlikely, but maybe we should cover it.)

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
```diff
@@ -253,13 +209,71 @@ public PortletDefinitionForm getPortletDefinitionForm(IPerson person, String por
...
+            form.addPermission(bean.getName() + "_" + activity);
+        }
+    }
+
+    /*
+     * Create a {@code PortletDefinitionForm} and pre-populate it with default categories and group permissions.
+     */
+    private PortletDefinitionForm createNewPortletDefinitionForm() {
```
drewwills: It looks like this behavior was pulled from existing code and refactored into this method (good idea).  But it doesn't look like the way the existing interface actually behaves... does it?  Categories & groups are empty, correct?

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
jameswennmacher: Actually it has.  There was discussion of adopting the Google Code standards.  One thing IMHO we might not is the actual order of the imports (Google Code standards have a different order than most of our files), but the Google Code standards do suggest never importing *.

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
```diff
@@ -253,13 +209,71 @@ public PortletDefinitionForm getPortletDefinitionForm(IPerson person, String por
...
+        PortletDefinitionForm form = new PortletDefinitionForm();
+
+        // pre-populate with top-level category
+        final IEntityGroup portletCategoriesGroup = GroupService.getDistinguishedGroup(GroupService.PORTLET_CATEGORIES);
+        form.addCategory(new JsonEntityBean(portletCategoriesGroup, groupListHelper.getEntityType(portletCategoriesGroup)));
+
+        // pre-populate with top-level group
+        final IEntityGroup everyoneGroup = GroupService.getDistinguishedGroup(GroupService.EVERYONE);
+        form.addGroup(new JsonEntityBean(everyoneGroup, groupListHelper.getEntityType(everyoneGroup)));
+        form.addPermission(everyoneGroup.getName() + "_" + IPermission.PORTLET_SUBSCRIBER_ACTIVITY);
```
drewwills: Not clear yet on why there is both addGroup() -- the old way -- and addPermission(PORTLET_SUBSCRIBER_ACTIVITY)  -- the new way.  These appear to cover the same information (same part of the form).  Why not addSubscribeGroup() & addBrowseGroup()?

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
```diff
@@ -295,99 +309,142 @@ public PortletDefinitionForm savePortletRegistration(IPerson publisher,
...
-				groupMembers[i] = GroupService.findGroup(bean.getId());
-			} else {
-            	groupMembers[i] = GroupService.getGroupMember(bean.getId(), entityEnum.getClazz());
-				
-			}
-		}
+        // create the group array from the form's group list -- only groups with permissions
+        final List<IGroupMember> groupList = new ArrayList<>(form.getGroups().size());
+        final List<IGroupMember> subscribeList = new ArrayList<>(form.getGroups().size());
+        final List<IGroupMember> browseList = new ArrayList<>(form.getGroups().size());
```
drewwills: Not sure these are inherently ordered, so perhaps a Set is better for these.

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
timlevett: This catch block also eats the exception, do we want that behavior or a rethrow?

drewwills: Probably a re-throw;  if this is one of those rare circumstances where we don't, it's worth a comment.

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
bjagg: I have updated my IDE since this PR to not collapse imports. Will revert to individual import statements.

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
bjagg: Agreed. Also, added final to the class as this is a helper class that is not intended to be subclassed.

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
bjagg: Changed name to addSubscribePermissionsToForm() and updated the comment

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
bjagg: It was already autowired, just not used. Updated the methods to use the autowired one.

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
bjagg: It was grated on me as well. Cleaned this up.

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
bjagg: Group is used to match the Label in the UI.

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
bjagg: There are no checked exceptions thrown from getGroupPermissions() -- removing the try-catch block.

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
bjagg: They are initially empty. The contains check is to prevent adding duplicate principals while looping through the permission set where a principal is likely to have both SUBSCRIBE and BROWSE permissions.

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
bjagg: Changed to Sets -- good catch!

uportal-war/src/main/java/org/jasig/portal/portlets/portletadmin/PortletAdministrationHelper.java
bjagg: The group collection drives the HTML table in the form while the permissions are referenced only to set the checkbox state. The permissions list alone could be used but at the cost of more parsing and logic in the JSP. 

One thing I am investigating is if the groups and categories could be populated with just the names instead of JSON objects. I did not want to remove groups (as JSON objects) in case they were used somewhere else in the code. Looking at this simplification now.

jameswennmacher:
@bjagg I have been testing and found a few issues.  Form submission fails under 2 cases:
1. Selecting a PAGS group with a space in the name, such as "Authenticated Users" fails because the form field name contains a space
2. Selecting a user such as Amy Admin fails. Can't find the property to set.

jameswennmacher:
One other minor thing ... I had made an improvement to the tool tips I suggest incorporating.  See https://github.com/jameswennmacher/uPortal/commit/4801f1c5d145ea74877100300fcdbc5358ccc69c

jameswennmacher:
@bjagg I forgot to suggest one other thing.  I think the user experience would be much better if when you click on the button to add a new principal if the default is that both Browse and Subscribe are checked (defaulted to active).  That is the normal behavior most users would want so it saves them from checking the checkboxes, or forgetting to check them and losing the principal they just attempted to add.

bjagg:
Thanks for the feedback, James. I addressed your three comments.

jameswennmacher:
Looking very good.  I did notice that if you add a person with a dash in the name, such as 'student-lo' it fails when the portlet form is posted.

jameswennmacher:
+1 Merging with minor fix to handle other characters that might appear in principal names that SpEL complains about (dash and parenthesis).

PR #593, opened 2015-10-23
UP-4561: set attribute portalCookieAlwaysSecure on PortletCookieServi…

bjagg:
…ceImpl from portal.properties. This allows for stronger security in HTTPS environments.

https://issues.jasig.org/browse/UP-4561

### DIFF
--- a/filters/local.properties
+++ b/filters/local.properties
@@ -91,3 +91,5 @@ environment.build.ldap.url=
 environment.build.ldap.baseDn=
 environment.build.ldap.userName=
 environment.build.ldap.password=
+
+secureSessionCookie=false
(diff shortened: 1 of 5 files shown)


filters/local.properties
```diff
@@ -91,3 +91,5 @@ environment.build.ldap.url=
 environment.build.ldap.baseDn=
 environment.build.ldap.userName=
 environment.build.ldap.password=
+
+secureSessionCookie=false
```
jameswennmacher: You should probably set a value in each filters/*.properties file, not just local.properties so it is more likely someone sees it and changes it if need be, and they can be set differently (I assume prod would default to true).

bjagg: Good point. Also should add a comment to point out the other related property in portal.properties

drewwills:
Adjusted per comments and merged out-of-band.

PR #595, opened 2015-11-11
UP-4570:  Portlet Marketplace -- make it possible to publish the Mark…

drewwills:
…etplace portlet multiple times, targeting different portlets each time

https://issues.jasig.org/browse/UP-4570

This is an optional setting for the Marketplace portlet.

This enhancement covers 2 Marketplace scenarios:
* I want to publish 2 or more copies of the Marketplace portlet, each with it's own collection of portlets (based on category)
* I want to publish the Marketplace portlet 1 time, but I want to limit the portlets that appear in it (based on category)

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/portlet/marketplace/MarketplaceService.java
+++ b/uportal-war/src/main/java/org/jasig/portal/portlet/marketplace/MarketplaceService.java
@@ -16,6 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+
 package org.jasig.portal.portlet.marketplace;
 
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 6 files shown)


uportal-war/src/main/java/org/jasig/portal/portlet/marketplace/MarketplaceService.java
```diff
@@ -110,23 +130,63 @@ public void onApplicationEvent(LoginEvent loginEvent) {
...
+                    boolean includeIt = false;  // default
+                    for (final PortletCategory parent : parents) {
+                        if (allSpecifiedAndDecendantCategories.contains(parent)) {
+                            includeIt = true;
+                            break;
+                        }
+                    }
+                    if (includeIt) {
+                        filteredPortletDefinitions.add(portletDefinition);
+                        break;
```
drewwills: There's a bug here -- fix incoming.

PR #596, opened 2015-11-13
Nojira add offcanvas stickynav flyout

cousquer:
Add sticky nav like mySail but with notification and greeting on mobile
Restore optional Fly-out menu bootstrapized. (the menu toggle is on the left of each tab because it won't interfere with delete tab link)
Add a new optional off-canvas menu for the main nav.
in navigation.xsl  you just need to set true for these param
<xsl:param name="USE_FLYOUT_MENUS" select="'false'" /> 
<xsl:param name="USE_OFFCANVAS_MENU" select="'true'" />


cousquer:
I've pushed a new pull request

PR #601, opened 2015-11-19
UP-4583: Add ability to use SpEL expressions in stylesheet attribute …

groybal:
…values.

https://issues.jasig.org/browse/UP-4583

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/rendering/StructureAttributeSource.java
+++ b/uportal-war/src/main/java/org/jasig/portal/rendering/StructureAttributeSource.java
@@ -22,6 +22,8 @@ package org.jasig.portal.rendering;
 import javax.servlet.http.HttpServletRequest;
 
 import org.jasig.portal.layout.IStylesheetUserPreferencesService.PreferencesScope;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Returns structure stylesheet descriptor and preferences data
@@ -30,8 +32,15 @@ import org.jasig.portal.layout.IStylesheetUserPreferencesService.PreferencesScop
  * @version $Revision$
  */
 public class StructureAttributeSource extends StylesheetAttributeSource {
+    private static final Logger LOGGER = LoggerFactory.getLogger(StructureAttributeSource.class);
+
     @Override
     public PreferencesScope getStylesheetPreferencesScope(HttpServletRequest request) {
         return PreferencesScope.STRUCTURE;
+    }
+
+    @Override
+    protected Logger getLogger() {
+        return LOGGER;
     }
 }
--- a/uportal-war/src/main/java/org/jasig/portal/rendering/StylesheetAttributeSource.java
+++ b/uportal-war/src/main/java/org/jasig/portal/rendering/StylesheetAttributeSource.java
@@ -30,8 +30,6 @@ import javax.xml.stream.XMLEventFactory;
 import javax.xml.stream.events.Attribute;
 import javax.xml.stream.events.StartElement;
 
-import org.jasig.portal.IUserPreferencesManager;
-import org.jasig.portal.IUserProfile;
 import org.jasig.portal.layout.IStylesheetUserPreferencesService;
 import org.jasig.portal.layout.IStylesheetUserPreferencesService.PreferencesScope;
 import org.jasig.portal.layout.IUserLayoutManager;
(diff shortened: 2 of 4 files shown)


uportal-war/src/main/java/org/jasig/portal/rendering/StructureAttributeSource.java
```diff
@@ -30,8 +32,15 @@
  * @version $Revision$
  */
 public class StructureAttributeSource extends StylesheetAttributeSource {
+    private static final Logger LOGGER = LoggerFactory.getLogger(StructureAttributeSource.class);
```
jameswennmacher: This is good.  As a personal preference I suggest in the future using a non-static so you don't have to specify the class because it is an opportunity for an error as code is copied or classes change; e.g. something like
private final Logger LOGGER = LoggerFactory.getLogger(this.getClass());

timlevett: Even further, since the class is not final, I suggest doing `protected` so that children classes can use it :)

uportal-war/src/main/java/org/jasig/portal/rendering/StylesheetAttributeSource.java
```diff
@@ -133,4 +150,35 @@ public IStylesheetDescriptor getStylesheetDescriptor(HttpServletRequest request)
     }
     
     public abstract PreferencesScope getStylesheetPreferencesScope(HttpServletRequest request);
+
+    protected abstract Logger getLogger();
+
+    protected boolean shouldDoSpelEvaluationForAttributeValue(final String attributeValue) {
+        return this.portalSpELService != null && attributeValue != null && attributeValue.startsWith("${") && attributeValue.endsWith("}");
```
jameswennmacher: Might have been cool to allow an expression such as student-${person.attributeMap['college']}-apps, but  you can handle that with SpEL so this approach is sufficient and probably cleaner.

uportal-war/src/main/java/org/jasig/portal/rendering/StylesheetAttributeSource.java
```diff
@@ -133,4 +150,35 @@ public IStylesheetDescriptor getStylesheetDescriptor(HttpServletRequest request)
...
+            return attributeValue;
+        }
+    }
+
+    protected String doSpelEvaluationForAttributeValue(final WebRequest request, final String attributeValue) {
+        String result;
+        try {
+            final String valueForEvaluation = this.getValueForSpelEvaluation(attributeValue);
+            final Expression expression = this.portalSpELService.parseExpression(valueForEvaluation);
+            result = this.portalSpELService.getValue(expression, request, String.class);
```
jameswennmacher: Minor but the above 3 lines could be simplified to
```
return this.portalSpELService.getValue(getValueForSpelEvaluation(attributeValue), request, String.class);
```
It would also allow you to avoid the try/catch block.

uportal-war/src/main/java/org/jasig/portal/rendering/StructureAttributeSource.java
jameswennmacher: Since you are exposing a logger for subclasses with getLogger(), you could probably just have done
protected final Logger LOGGER = LoggerFactory.getLogger(this.getClass());
instead.

uportal-war/src/test/java/org/jasig/portal/rendering/StylesheetAttributeSourceForSpELVariablesTest.java
```diff
@@ -0,0 +1,261 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.expression.EvaluationException;
+import org.springframework.expression.Expression;
+import org.springframework.expression.ParseException;
+import org.springframework.web.context.request.WebRequest;
+
+/**
+ * JUnit test class for StyleAttributeSource tests with SpEL variables.
+ */
+public class StylesheetAttributeSourceForSpELVariablesTest {
```
jameswennmacher: Unit tests.  I love it!

jameswennmacher:
Overall +1 for merge.  I prefer the suggested minor simplifications were incorporated first, but if they don't get done let's merge this in anyway.

uportal-war/src/main/java/org/jasig/portal/rendering/StructureAttributeSource.java
groybal: I tend to prefer static loggers, but it's not a big deal to me.  It can go either way.  SLF4J does not recommend one over the other:  http://www.slf4j.org/faq.html#declared_static

Child classes can access the logger using the protected getLogger() method, but I would recommend that child classes define their own logger and override getLogger() to return it.

getLogger() is not so much to expose the logger to the subclasses, but rather for the superclass to use the same logger as the child so the log messages will show the subclass name.

uportal-war/src/main/java/org/jasig/portal/rendering/StylesheetAttributeSource.java
groybal: I used the separate parseExpression(...) and getValue(expression, ..) calls simply because I copied over code from elsewhere and that is how it was done there.  I wasn't really clear on the differences, if any, between getValue(Expression...) and getValue(String...).  Note that I also could have left out the try..catch in my code, but the desired behavior is to not let the exception bubble up, but rather to just return the expression as the value.

jhelmer-unicon:
:+1:   This PR has been in-use for CCC for a while now and has been working well for them.  Baring downvotes, will most likely try to merge this change sometime in the next 24 hours.

jhelmer-unicon:
I pulled this down locally, dealt with merge conflicts and did some basic testing to ensure that I hadn't introduced any oddness with my merge.   Hearing no objections to merge this, merging manually.

PR #602, opened 2015-11-20
UP-4586:  Resource Aggregation is broken in Respondr

drewwills:
https://issues.jasig.org/browse/UP-4586

It looks as though a critical piece of the Resource Aggregation setup was omitted in Respondr: you have to use (apparently) the resource-server-plugin to aggregate the resources in the build process.
I have prepared a patch (will create a PR) – and have demonstrated that the patch makes aggregation in Respondr work as well as it did in Universality – but I'm confused as to why some resources get aggregated while others don't.
The uPortal JS in skins/common/javascript/uportal/ DOES get aggregated, but none of the other JS libs (e.g. jQuery, etc.) do, and none of the CSS does.

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/portlet/rendering/PortletEventCoordinatationService.java
+++ b/uportal-war/src/main/java/org/jasig/portal/portlet/rendering/PortletEventCoordinatationService.java
@@ -60,6 +60,7 @@ import org.jasig.portal.layout.IUserLayoutManager;
 import org.jasig.portal.portlet.container.EventImpl;
 import org.jasig.portal.portlet.om.IPortletDefinition;
 import org.jasig.portal.portlet.om.IPortletDefinitionId;
+import org.jasig.portal.portlet.om.IPortletDefinitionParameter;
 import org.jasig.portal.portlet.om.IPortletEntity;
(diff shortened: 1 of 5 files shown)


uportal-war/src/main/java/org/jasig/portal/portlet/rendering/PortletEventCoordinatationService.java
```diff
@@ -275,7 +276,16 @@ public void resolvePortletEvents(HttpServletRequest request, PortletEventQueue p
                 	if (defaultPortletEntity.equals(portletEntity)) {
                 		portletDefinitions.remove(portletDefinition);
                 	}
-                	
+
+                    // Is this portlet permitted to receive events?  (Or is it disablePortletEvents=true?)
+                    IPortletDefinitionParameter disablePortletEvents = portletDefinition.getParameter(
```
drewwills: Ack!  This delta slipped in from another PR... maybe merging it will remove it?

drewwills:
I'm going to re-roll this PR to take out the deltas that aren't a part of it.

drewwills:
Merged out-of-band

PR #603, opened 2015-11-30
UP-4408:  Returning from DETACHED state goes to 1st tab instead of mo…

drewwills:
…st recent

https://issues.jasig.org/browse/UP-4408

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/url/UrlSyntaxProviderImpl.java
+++ b/uportal-war/src/main/java/org/jasig/portal/url/UrlSyntaxProviderImpl.java
@@ -38,6 +38,7 @@ import java.util.regex.Pattern;
 import javax.portlet.PortletMode;
 import javax.portlet.WindowState;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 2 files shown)


jameswennmacher:
See updated comment in UP-4408.  I can still duplicate the original issue reported, and this change should not resolve the original issue reported.  

Other than that, this looks like a righteous improvement that's well implemented and documented.   +1

drewwills:
@jameswennmacher -- I'm not sure about well-documented, but there are comments at least.

You're right about UP-4408.  Interestingly, the sponsoring client's use case needs a bit of extra "help" to benefit from this enhancement... the same help would solve UP-4408.

The difference is this (afaik):  the URL in your use case DOES specify a folder... but the one it specifies doesn't actually qualify because it's not actually a tab.  The specified folder trumps the remembered one (which is ignored), but the structure XSL realizes that it's not a tab, and therefore sends you to the first tab.

drewwills:
Closing this PR because the issue has been solved in a much simpler way.

PR #604, opened 2015-12-01
UP-4518 Add Lifecycle to portlet definition, rev portlet definition

jameswennmacher:
https://issues.jasig.org/browse/UP-4518

Add support for optional lifecycle element.  In a portlet definition, you can add a lifecycle element:

Bumped portlet-definition to 4.3.  Added optional lifecycle element:

```xml
    <lifecycle>
        <approved>2015-11-06T12:48:19.717-07:00</approved>
        <published user="admin">2015-11-06T12:48:19.717-07:00</published>
        <expiration user="admin">2015-12-16T12:48:19.717-07:00</expiration>
    </lifecycle>
```
* lifecycle with no children: created
* user is optional, defaults to system
* expiration can be in the past (expired) or future (scheduled expiration)
* prior elements not necessary; e.g. can specify just published and approved will be defaulted appropriately

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/UserInstance.java
+++ b/uportal-war/src/main/java/org/jasig/portal/UserInstance.java
@@ -26,7 +26,7 @@ import org.jasig.portal.user.IUserInstance;
 
 
 /**
- * A class handling holding all user state information. The class is also reponsible for
+ * A class handling holding all user state information. The class is also responsible for
  * request processing and orchestrating the entire rendering procedure.
  * (this is a replacement for the good old LayoutBean class)
  * @author Peter Kharchenko  {@link <a href="mailto:pkharchenko@interactivebusiness.com"">pkharchenko@interactivebusiness.com"</a>}
--- a/uportal-war/src/main/java/org/jasig/portal/io/xml/portlet/PortletDefinitionImporterExporter.java
+++ b/uportal-war/src/main/java/org/jasig/portal/io/xml/portlet/PortletDefinitionImporterExporter.java
@@ -368,7 +368,7 @@ public class PortletDefinitionImporterExporter
             if (username != null) {
                 return username;
             }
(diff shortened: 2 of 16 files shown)


uportal-war/src/main/java/org/jasig/portal/UserInstance.java
```diff
@@ -26,7 +26,7 @@
 
 
 /**
- * A class handling holding all user state information. The class is also reponsible for
+ * A class handling holding all user state information. The class is also responsible for
```
apetro: I wish this unrelated change were in a different commit.

apetro:
Love the feature.  Love the thoughtful, graceful implementation of defaults.  Love that there's test coverage over some of this. Love that there's an upgrade path and that the new entity element is eminently optional.

I'd quibble with some of the implementation details. The private methods look procedural where a more object-oriented approach might be more amenable to test coverage.

:+1:.

uportal-war/src/main/java/org/jasig/portal/io/xml/portlet/PortletDefinitionImporterExporter.java
```diff
@@ -256,14 +265,116 @@ public void importData(ExternalPortletDefinition portletRep) {
...
+    }
+
+    // Returns the username for a valid userId, else the system username
+    private String getUsernameForUserId(int id) {
+        if (id > 0) {
+            String username = userIdentityStore.getPortalUserName(id);
+            if (username != null) {
+                return username;
+            }
+            logger.warn("Invalid userID {} found when exporting a portlet; return system username instead");
```
apetro: Placeholder `{}` without placeholder value?

jameswennmacher: Excellent catch.  Thanks.

PR #605, opened 2015-12-17
UP-4599: Add permission owner UP_SYSTEM for MAX_INTERVAL and skip eva…

bjagg:
…luation of permission after firsrt time per session.

This will allow changes to this permission by importing data files. Currently, the permission is not updated due to the missing permission owner. Once it was set, it could not be changes without direct database updates.

https://issues.jasig.org/browse/UP-4599


jameswennmacher:
+1.  Best with PR 614

PR #606, opened 2015-12-17
UP-4600:  DistributedLayoutManager may incorrectly return the same ca…

drewwills:
…cheKey for users who have different layouts

https://issues.jasig.org/browse/UP-4600

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/layout/IUserLayoutManager.java
+++ b/uportal-war/src/main/java/org/jasig/portal/layout/IUserLayoutManager.java
@@ -247,7 +247,7 @@ public interface IUserLayoutManager {
     public String getPreviousSiblingId(String nodeId) throws PortalException;
 
     /**
-     * Return a cache key, uniqly corresponding to the composition and the structure of the user layout.
+     * Return a cache key, uniquely corresponding to the composition and the structure of the user layout.
      *
      * @return a <code>String</code> value
      * @exception PortalException if an error occurs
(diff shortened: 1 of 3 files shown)


apetro:
:+1:  Less broken is less broken.

Are there any circumstances in which there's rightly a cache hit, or, given that the cache keys are simply random numbers, should layout caching be removed outright as a never-implemented idea?

jameswennmacher:
Actually the random number approach is generally fine (though technically a chance two instances can get the same random number).  It results in some extra processing where you might have the same layout (multiple guest users, multiple instances of the same layout because users did not do any layout customizations) and could potentially generate the same key instead of a random number.  

The layout caching is used fairly extensively through out the rendering pipeline for performance benefits.  See invocations of DistributedLayoutManger.getCacheKey() and in particular UserLayoutStoreComponent.getCacheKey().

jameswennmacher:
+1

jameswennmacher:
To clarify RE the caching, when DLM loads and processes a user's layout, it generates a random number to use to cache the item in (specifically, DLM callers use that random number in their cache key calculations).  When a user makes layout changes, the random number is re-generated thus invalidating the calling instances caches.  So the random number approach, though not as good as a UUID calculation since you could get two random numbers that are identical within a cache expiration time period, is sufficient.  It is just not optimally sufficient (two users with effectively identical layouts would typically have two different random numbers even though they could share the same cached instance, assuming the caller's overall cache key algorithm that includes the DLM.getCacheKey() value would not be unique per user, session, etc.

drewwills:
@apetro there are many circumstances with appropriate cache hits with this implementation;  they are all, however, user-scoped.  Before my change -- as far as I can tell -- correct cache hits were always for the same user who cached it in the first place.  That is still true with my change.  The difference is that this change removes a chance to have an incorrect cache hit (across different users).

Once you reach the next stage of the rendering pipeline, the results of the first stage will be cached for you.  The cached results will be replaced if you change your layout.

PR #607, opened 2015-12-22
UP-4566: Limit additional portlet parameters to when copy render para…

bjagg:
…ms is found. Performed some testing, but more could be used. 
https://issues.jasig.org/browse/UP-4566


jameswennmacher:
Simple, elegant change.  Thanks Benito!  It will be great to have this in when the unit tests are working.

PR #608, opened 2015-12-28
UP-4603: Updated to allow Tenant Manager to parse templates that cont…

groybal:
…ain CDATA elements.

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/io/xml/SpELDataTemplatingStrategy.java
+++ b/uportal-war/src/main/java/org/jasig/portal/io/xml/SpELDataTemplatingStrategy.java
@@ -28,7 +28,6 @@ import javax.xml.transform.sax.SAXSource;
 import org.dom4j.Attribute;
 import org.dom4j.Document;
 import org.dom4j.Node;
-import org.dom4j.Text;
 import org.dom4j.io.DocumentSource;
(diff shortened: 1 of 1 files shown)


uportal-war/src/main/java/org/jasig/portal/io/xml/SpELDataTemplatingStrategy.java
```diff
@@ -80,11 +79,11 @@ public Source processTemplates(Document data, String filename) {
                         }
                         break;
                     case org.w3c.dom.Node.TEXT_NODE:
-                        Text t = (Text) n;
-                        inpt = t.getText();
```
jameswennmacher: Gary is TEXT_NODE a subset of CDATA_SECTION_NODE?  E.g. will a text node get processed within the CDATA_SECTION_NODE case?

groybal: Since there is no "break;" the code will run for both CDATA and TEXT nodes.

jameswennmacher: Ahh. Didn't catch that.  Thanks.

jameswennmacher:
+1

PR #610, opened 2015-12-31
UP-4582: Fix issue with moving portlets into empty columns.

bjagg:
https://issues.jasig.org/browse/UP-4582


jameswennmacher:
+1

PR #612, opened 2016-01-11
UP-4529: Replace form text input with select in Skin Manager pop-up.

bjagg:
https://issues.jasig.org/browse/UP-4529


drewwills:
@bjagg this looks like a good change... is it working?

bjagg:
@drewwills Yes, from the limited testing I did it was working quite well.

uportal-war/src/main/java/org/jasig/portal/portlets/dynamicskin/FileSystemDynamicSkinService.java
```diff
@@ -249,4 +250,33 @@ public String calculateTokenForCurrentSkin(PortletRequest request) {
...
+    @Override
+    public SortedSet<String> getSkinNames(PortletRequest request) {
+        // Context to access the filesystem
+        PortletContext ctx = request.getPortletSession().getPortletContext();
+
+        // Determine the full path to the skins directory
+        String skinsFilepath = ctx.getRealPath(rootFolder);
+
+        // Create File object to access the filesystem
+        File skinsDir = new File(skinsFilepath);
```
drewwills: Sooo...

I'm not sure it matters enough to bother with now, but I wonder if we'd be better served by getting the list of skins from the skinList.xml file.  (This file normally controls which skins are selectable in the personalization-gallery UI.)

Technically Respondr skins can work just swimmingly without an entry in that file, but...
- We presumably have a routine somewhere for reading that file and figuring out what skins there are;  it would be attractive to do that one way, and perhaps even enhance it to be more powerful/flexible
- This path would give us a way to specify which skins we want to offer;  scanning the directory will not allow us to exclude defaultSkin.less as a potential choice

uportal-war/src/main/java/org/jasig/portal/portlets/dynamicskin/FileSystemDynamicSkinService.java
```diff
@@ -47,6 +31,23 @@
...
+import java.io.IOException;
+import java.net.URL;
+import java.text.MessageFormat;
+import java.util.Enumeration;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeSet;
+import java.util.concurrent.CopyOnWriteArraySet;
```
jameswennmacher: FWIW it would be good to update your intelliJ to use the same order as eclipse which is the uPortal standard to avoid this noise; e.g. see 

intellijimportsettings

uportal-war/src/main/java/org/jasig/portal/portlets/dynamicskin/FileSystemDynamicSkinService.java
jameswennmacher: That is a great idea @drewwills. Perhaps a later enhancement as you say so we can capture this value for now.

jameswennmacher:
+1

uportal-war/src/main/java/org/jasig/portal/portlets/dynamicskin/FileSystemDynamicSkinService.java
jameswennmacher: I just looked into it.  The javascript (up-skin-selector.js) requests the XML file and parses it itself.  The back end does not do any parsing of the file.

jameswennmacher: So for what it is worth, we now parse the skinList.xml and provide the options listed there.  That was a good idea @drewwills

jameswennmacher: @drewwills BTW I assumed the skin-key was the filename and I returned that.  I assumed the skin-name was the friendly user name.  If that is not the correct convention, let me know but it looked like the javascript code was honoring that convention.

PR #614, opened 2016-01-15
UP-4614: Add special handling of MAX_INACTIVE permission for changing …

bjagg:
…value.

See https://issues.jasig.org/browse/UP-4614


jameswennmacher:
+1

PR #615, opened 2016-01-19
UP-4566: search fails lots of portlets

bjagg:
Add special handling of Search request to minimize additional portlet parameters.

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/url/UrlSyntaxProviderImpl.java
+++ b/uportal-war/src/main/java/org/jasig/portal/url/UrlSyntaxProviderImpl.java
@@ -59,6 +59,7 @@ import org.jasig.portal.xml.xpath.XPathOperations;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 import org.springframework.web.util.UrlPathHelper;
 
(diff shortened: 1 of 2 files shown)


uportal-war/src/main/java/org/jasig/portal/url/UrlSyntaxProviderImpl.java
```diff
@@ -1091,7 +1099,11 @@ protected void addPortletUrlData(
             final String portletWindowIdStr = portletWindowId.toString();
             prefixedPortletWindowId = SEPARATOR + portletWindowIdStr;
             suffixedPortletWindowId = portletWindowIdStr + SEPARATOR;
-            url.addParameter(PARAM_ADDITIONAL_PORTLET, portletWindowIdStr);
+
+            // handle URI (i.e. search) that requires skipping additional portlet info
+            if (!request.getRequestURI().contains(this.compactParamsForUriFragment)) {
+                url.addParameter(PARAM_ADDITIONAL_PORTLET, portletWindowIdStr);
+            }
```
jameswennmacher: This approach is specific to search.  Is there not a way to determine if there are actual parameter values that would be sent to the additional portlet and if not then skip adding it?  As I recall I thought the purpose of this additionalPortlet flag is to add a string that is also added in the windowId of the additional parameters to aid in later parsing.

My thought is that if the switch statement and if statements below in this method do not call url.addParameter() for this portlet, then don't invoke url.addParameter(PARAM_ADDITIONAL_PORTLET, portletWindowIdStr); e.g. add this parameter later if we decide we actually add any parameters for this portlet to the URL.

This would also eliminate passing in the compactParamsForUriFragment property.

bjagg: The original approach failed three unit tests in UrlSyntaxProviderImplTest. The tests checked for the additional parameters. I am reviewing those for commonality. All three explicitly set "action=dashboard".

So simply checking if portletUrlBuilder.getCopyCurrentRenderParameters() returned a value was insufficient. I will add a check for "action=dashboard" and test.

jameswennmacher:
Closing in favor of https://github.com/Jasig/uPortal/pull/617

PR #616, opened 2016-01-22
UP-4617: clean up data set permissions

bjagg:
Reducing default permissions for developers an fragment owners.


uportal-war/src/main/data/default_entities/permission_set/Portal_Administrators__ALL_PERMISSIONS__UP_SYSTEM.permission-set.xml
```diff
@@ -20,13 +20,13 @@
 
 -->
 <permission-set script="classpath://org/jasig/portal/io/import-permission_set_v3-1.crn">
-  <owner>UP_PORTLET_PUBLISH</owner>
+  <owner>UP_SYSTEM</owner>
```
drewwills: Is this change correct?  It really _should_ be owned by UP_SYSTEM?

uportal-war/src/main/java/org/jasig/portal/security/provider/DefaultPermissionPolicy.java
```diff
@@ -1,139 +0,0 @@
...
-
-/**
- * Implements a strategy for answering the basic authorization question: does the
- * <code>principal</code> have permission to perform the <code>activity</code> on
- * the <code>target</code>.
- *
- * @author Dan Ellentuck (de3@columbia.edu)
- * @version $Revision$
- */
-public class DefaultPermissionPolicy implements IPermissionPolicy {
```
drewwills: Why are we getting rid of this file?  (I can imagine some good reasons... but I don't see a reason in the ticket or the PR.)

bjagg: The reason to remove in this PR is thin -- I found it confusing to have an unused class when checking the policy code.

PR #621, opened 2016-02-16
UP-4615 tenants in quickstart

jameswennmacher:
Improve tenant information in quickstart data set

* creates default tenant admins with password=admin (useful in quickstart and demo)
* default tenant admin experience
* gives tenant fragment admins the Fragment Exit
* assumes tenants might have multiple fragment layouts so have a respondr fragment layout plus additional
* renames tenant sample data set from tenants/data to tenants/sampledata to make it more clear you should copy it and customize it
    
Assumes that the Groups BROWSE permission assignment change (https://github.com/Jasig/uPortal/pull/589) will be incorporated to allow non-admin tenants to see portlets created by the tenant admin

tenant admin default experience


gcarroll:
+1👍

drewwills:
Is it a good idea to add tenant-owned data to the quickstart_entities before we have Import/Export support for ITenant objects?  (Does this PR do that?... I didn't look at all of it.)

+1 for improving the quickstart_entities & default_entities as far as what permissions are given to the 'Tenant Administraters' group and in other ways not related to tenant-owned data.

jameswennmacher:
The PR doesn't do anything for ITenant objects.  Perhaps you could clarify with me what you are thinking.

drewwills:
`The PR doesn't do anything for ITenant objects. Perhaps you could clarify with me what you are thinking.`

Does it include any data (in src/main/data) that would be generated by creating a tenant in the Tenant Manager?

jameswennmacher:
RE the ITenant data.  No it does not.

PR #622, opened 2016-02-19
NPE when no results. getSearchResults should return a empty list when…

acolebourne:
… no results are found.


vertein:
:+1:

jameswennmacher:
+1

drewwills:
Nice @acolebourne, thank you!

PR #624, opened 2016-02-24
Upgrade DataTables and include responsive plugin using CDN. Added loc…

acolebourne:
…al copy of upgraded ColumnFilterWidgets plugin. Refactor all uses of DataTables to use the new API syntax

NOTE: Marketplace has not been updated because it continues to function using the deprecated APIs. I will (re)submit a separate pull request for additional Marketplace UI improvements.


acolebourne:
I've created a Jira for this, see https://issues.jasig.org/browse/UP-4653

auxepaul:
It uses CDNs instead of the resource-server : https://github.com/acolebourne/uPortal/blob/responsive-datatables-pull/uportal-war/src/main/webapp/media/skins/respondr/common/common_skin.xml#L30 and https://github.com/acolebourne/uPortal/blob/responsive-datatables-pull/uportal-war/src/main/webapp/media/skins/respondr/common/common_skin.xml#L91
Why?

@see https://wiki.jasig.org/display/PLT/JavaScript+Best+Practices

acolebourne:
The way datatables 1.9.4 had been added to the resource server caused multiple requests, if we are to add version 1.10 to the resource server I suggest we use a url scheme similar to that of the DataTabes CDN site e.g. /t/bs/dt-1.10.11,r-2.0.2/ 

We need to be clear what features are included/combined without requiring multiple requests. We loose some flexibility for change, but gain off-line working. I'm happy to make a PR on resource-server.

drewwills:
@acolebourne -- James has some really cool ideas around enhancing the ResourceServer to use CDN URLs (instead of internal) when permitted and when they are available.  I guess I'd like to preserve offline functionality as much as possible.

jameswennmacher:
@acolebourne  see https://issues.jasig.org/browse/RESSERV-82.  I've been trying to get to this for a while and haven't been able to carve out the time.  Shouldn't be too difficult to do.  I know you are passionate about parallelizing URL requests by using CDNs.  If you have time to work on it before me, I think it would be great!

acolebourne:
@drewwills it would be good to get this upgrade into 4.3.0 so I'll resubmit using ResourceServer, I'll also make the corresponding PR on the ResourceServer repo. Hopefully before Thursday.

acolebourne:
### update - Switched to using ResourceServer ###

**Note**: The branch has been rebased
**Note**: This PR is dependent on [PR #44](https://github.com/Jasig/resource-server/pull/44) on resource server and would require an upgrade to the bundled resource-server version (not included in the PR)

drewwills:
It looks like this item just needs some busywork.  I'll try to get to it.

drewwills:
This is a change that we want, but the conflicts and aging of this PR take it beyond what we want to merge as-is.  I hope we address it effectively in uP5.

PR #626, opened 2016-02-25
UP-4638:  DLM -- HTTP 403 Forbidden from /api/layout is ignored by po…

drewwills:
…rtlet reordering JavaScript

https://issues.jasig.org/browse/UP-4638

up-4638

### DIFF
--- a/uportal-war/src/main/resources/properties/i18n/Messages.properties
+++ b/uportal-war/src/main/resources/properties/i18n/Messages.properties
@@ -391,6 +391,7 @@ most.frequently.added=Most Frequently Added
 move=Move
 move.this.column=Move this column
 move.this.portlet=Move this portlet
+move.this.portlet.error=This change was disallowed by the server.  The portlet will return to its original position.  Reload the page now?
 move.this.tab=Move this tab
 my.account=My Account
 my.account.details=My Account Details
--- a/uportal-war/src/main/resources/layout/theme/respondr/respondr.xsl
+++ b/uportal-war/src/main/resources/layout/theme/respondr/respondr.xsl
@@ -558,6 +558,7 @@
                 messages: {
                     confirmRemoveTab: '<xsl:value-of select="upMsg:getMessage('are.you.sure.remove.tab', $USER_LANG)"/>',
                     confirmRemovePortlet: '<xsl:value-of select="upMsg:getMessage('are.you.sure.remove.portlet', $USER_LANG)"/>',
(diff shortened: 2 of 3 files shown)


uportal-war/src/main/resources/properties/i18n/Messages.properties
```diff
@@ -391,6 +391,7 @@ most.frequently.added=Most Frequently Added
 move=Move
 move.this.column=Move this column
 move.this.portlet=Move this portlet
+move.this.portlet.error=This change was disallowed by the server.  The portlet will return to its original position.  Reload the page now?
```
timlevett: This is neat, curious if the typical users would get confused by this. Maybe "You cannot move this portlet"? What would be really neat if then it moved it back, or even better if it didn't allow it to move in the first place. There could be something passed in from XSL saying whether or not it is movable. I remember being able to set something in the fragment layout. The move back option would be just capturing what was before it initially, then if you get the forbidden, then just run an "undo" dom change.

jameswennmacher: The goal is that an invalid move would not be allowed in the first place and the markup generated has classes to help the javascript in the UI know this.  However since the logic is implemented in two places they can (and obviously from this PR have) been out of sync.  This message captures when the UI thinks it can do something, but the back end fails the change.  Drew had mentioned that the current implementation does the move then calls the back end.  You have a point that it would be nice if that order of operations was reversed if possible.

drewwills: @timlevett -- James has the essence of it...
- The theme is already attempting to prevent illegal moves, and does a pretty good job.  As we identify cracks in that logic, we should address them as well.  I want to get this change in because I noticed that (a) the REST API was already sending 403 Forbidden whenever it detected an illegal move, but (b) the front end was not reacting to the non-200 HTTP status in any way.
- The portlet reorderer uses this fluid component:  https://wiki.fluidproject.org/display/docs/Layout+Reorderer+API.  My first choice would be to prevent/undo the move in these cases, but the event model of this widget does not make that easy.  (Widget only allows you to understand & deal with the move after it's done;  provides no "cancel" mechanism.) I don't we can put off other priorities long enough to make this fix that small amount better.

And actually -- this fix does "undo" the move, just at the cost of a page re-draw.

drewwills: And thanks @timlevett for jumping in... really appreciate the dialog.

timlevett: happy to help @drewwills . Thanks for the clarification.

timlevett:
:+1:

PR #627, opened 2016-02-29
UP-4640: Add support for specifying multiple files to import for data…

groybal:
…-import ant task.

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/shell/PortalShellBuildHelperImpl.java
+++ b/uportal-war/src/main/java/org/jasig/portal/shell/PortalShellBuildHelperImpl.java
@@ -18,11 +18,16 @@
  */
 package org.jasig.portal.shell;
 
+import java.io.IOException;
 import java.io.File;
+import java.nio.charset.Charset;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
 
+import com.google.common.io.Files;
+import org.apache.commons.io.Charsets;
 import org.apache.commons.lang.StringUtils;
 import org.jasig.portal.IUserIdentityStore;
 import org.jasig.portal.io.xml.IPortalData;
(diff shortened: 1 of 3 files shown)


jameswennmacher:
I think it might be nicer to take in a file that has a list of files in it. My thought being you could maintain and check in a filelist into git that applies to a release to track the files that need to be imported for that release.  That might make it easier from a dev/ops perspective since the file list is associated with that particular git commit and not maintained separately.

What you have above looks like it works, though I think the latter would be more useful.  What are your thoughts?

groybal:
Sounds like a good idea, I'll work on that today.  Is it worth keeping the -Dfiles option?

jameswennmacher:
Sure.  It's done :smile:

uportal-war/src/main/java/org/jasig/portal/shell/PortalShellBuildHelperImpl.java
```diff
@@ -367,4 +387,21 @@ private ISchemaExport getSchemaExport(String persistenceUnit) {
...
     }
+
+    @Override
+    public String getFilesListFromFile(String filePathsFile) throws IOException {
+        String filepaths = "";
+        List<String> lines = Files.readLines(new File(filePathsFile), Charsets.UTF_8);
+        boolean isFirst = true;
+        for (String line : lines) {
+            if (isFirst) {
+                isFirst = false;
```
jameswennmacher: isFirst isn't really doing anything and can be deleted.

jameswennmacher: Scratch that.  I just realized what it is doing.

uportal-war/src/main/java/org/jasig/portal/shell/PortalShellBuildHelperImpl.java
```diff
@@ -367,4 +387,21 @@ private ISchemaExport getSchemaExport(String persistenceUnit) {
         
         throw new IllegalArgumentException("No ISchemaExport bean found for persistence unit: '" + persistenceUnit + "'");
     }
+
+    @Override
+    public String getFilesListFromFile(String filePathsFile) throws IOException {
+        String filepaths = "";
```
jameswennmacher: Minor but since you are doing a lot of string concatenation it would be more efficient to use StringBuffer instead of String.

jameswennmacher:
+1 on merge.  Only saw one very minor performance improvement.

PR #628, opened 2016-03-02
UP-4549: DLM - Allow Fragment Owners to alter fragments that have DLM restrictions

drewwills:
https://issues.jasig.org/browse/UP-4549

This update applies to the following restrictions:

- dlm:moveAllowed
- dlm:deleteAllowed
- dlm:editAllowed
- dlm:addChildAllowed

Historically a fragment admin would have to remove these restrictions, make changes, then reinstate them.  That's non-intuitive, so this update allows fragment owners to ignore restrictions when changing fragment layouts.

### DIFF
--- a/uportal-war/src/main/resources/layout/theme/respondr/content.xsl
+++ b/uportal-war/src/main/resources/layout/theme/respondr/content.xsl
@@ -160,15 +160,6 @@
       </xsl:choose>
     </xsl:variable>
 
-    <!-- Applied in the case of DLM fragment owners only;  tells the UI to
-         permit the fragment admin to manage content even when it it locked. -->
-    <xsl:variable name="FRAGMENT_OWNER_CSS">
-      <xsl:choose>
-        <xsl:when test="$IS_FRAGMENT_ADMIN_MODE='true'">up-fragment-admin</xsl:when>
-        <xsl:otherwise></xsl:otherwise>
-      </xsl:choose>
-    </xsl:variable>
-
     <xsl:variable name="PORTLET_CHROME"> <!-- Test to determine if the portlet has been given the highlight flag. -->
       <xsl:choose>
         <xsl:when test="./parameter[@name='chromeStyle']/@value='no-chrome'">no-chrome</xsl:when>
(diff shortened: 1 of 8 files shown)


uportal-war/src/main/resources/layout/theme/respondr/content.xsl
```diff
@@ -151,6 +160,15 @@
...
 
+    <!-- Applied in the case of DLM fragment owners only;  tells the UI to
+         permit the fragment admin to manage content even when it it locked. -->
+    <xsl:variable name="FRAGMENT_OWNER_CSS">
+      <xsl:choose>
+        <xsl:when test="$IS_FRAGMENT_ADMIN_MODE='true'">up-fragment-admin</xsl:when>
+        <xsl:otherwise></xsl:otherwise>
+      </xsl:choose>
+    </xsl:variable>
+
```
jameswennmacher: Rather than duplicating the variable and conditional check, I suggest defining it once in respondr.xsl near the other fragment admin variable (IS_FRAGMENT_ADMIN_MODE) so it can be used anywhere.

uportal-war/src/main/resources/layout/theme/respondr/navigation.xsl
```diff
@@ -179,17 +179,25 @@
...
       </xsl:choose>
     </xsl:variable>
+    <!-- Applied in the case of DLM fragment owners only;  tells the UI to
+         permit the fragment admin to manage content even when it it locked. -->
+    <xsl:variable name="FRAGMENT_OWNER_CSS">
+      <xsl:choose>
+        <xsl:when test="$IS_FRAGMENT_ADMIN_MODE='true'">up-fragment-admin</xsl:when>
+        <xsl:otherwise></xsl:otherwise>
+      </xsl:choose>
+    </xsl:variable>
```
jameswennmacher: See previous comment on defining FRAGMENT_OWNER_CSS once in respondr.xsl

jameswennmacher:
+1 Excellent set of changes and I like the overall approach for implementation (very clean).  Just the one minor DRY comment.

PR #630, opened 2016-03-08
UP-4645:  Fail-fast when Authentication and person-directory attempt …

drewwills:
…to create a user without a username

https://issues.jasig.org/browse/UP-4645


bjagg:
+1

PR #633, opened 2016-03-13
UP-4646:  Add missing tenant-related caches to ehcache.xml

drewwills:
https://issues.jasig.org/browse/UP-4646


jameswennmacher:
+1

PR #634, opened 2016-03-13
UP-4647:  Suppress noisy WARN messages for optional properties files

drewwills:
https://issues.jasig.org/browse/UP-4647

### DIFF
--- a/uportal-war/src/main/java/org/springframework/context/support/ExtendedPropertySourcesPlaceholderConfigurer.java
+++ b/uportal-war/src/main/java/org/springframework/context/support/ExtendedPropertySourcesPlaceholderConfigurer.java
@@ -19,7 +19,11 @@
 package org.springframework.context.support;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanInitializationException;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
@@ -28,6 +32,7 @@ import org.springframework.core.env.PropertiesPropertySource;
 import org.springframework.core.env.PropertyResolver;
 import org.springframework.core.env.PropertySource;
 import org.springframework.core.env.PropertySourcesPropertyResolver;
+import org.springframework.core.io.Resource;
 
 
 /**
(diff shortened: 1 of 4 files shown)


uportal-war/src/main/java/org/springframework/context/support/ExtendedPropertySourcesPlaceholderConfigurer.java
```diff
@@ -37,6 +42,30 @@
...
+    @Override
+    public void setLocations(Resource[] locations) {
+        final List<Resource> list = new ArrayList<>();
+        for (Resource r : locations) {
+            if (r.exists()) {
+                list.add(r);
+            } else {
+                // In our case this event is worth a DEBUG note.
+                logger.debug("The following Resource was not present (it may be "
+                        + "optional, or it's absence may lead to issues):  ", r);
```
jameswennmacher: I suggest checking the 'ignoreResourceNotFound' property which we set to true in applicationContext.xml and if that property is set to false (the default), call the superclass method.  This code effectively removes the missing resource and Spring will not fail as it normally would in PropertiesLoaderSupport.loadProperties().  Only ignore the resource and issue a debug message if that property is set to true.  That maintains the spirit of the Spring contract.

jameswennmacher: And to add to above, since you'd only be logging if ignoreResourceNotFound=true, the log message could be changed to something like

The following Resource was not present.  It is flagged as optional so its absence should not lead to issues:

drewwills:
@jameswennmacher -- That's a good suggestion, and not one that I thought of before.  It may be feasible (and we should do it if it is), but I'm concerned that it may not work.  This change is to the setLocations() method, which is a bean setter method.  I'm not sure the order in which properties are set is known or reliable.  Even with ignoreResourceNotFound=true in the xml, I'm not sure we'll know it in time.

And I'm not sure we have the opportunity to refactor the code to a @PostConstruct method.  It may be too late, and I' not sure about what the superclass may do with @PostConstruct.

jameswennmacher:
That's a good point about the property order.  

At the moment the only place this class is used is applicationContext which has ignoreResourceNotFound=true so it doesn't matter for our current use but it would be confusing and painfully misleading if someone else uses it without setting that flag.

I think you could instead override loadProperties(Properties props) to call super if ignoreResourceNotFound=false but if ignoreResourceNotFound=true (and this is a bit ugly) filter the locations (which means you are effectively post-processing the locations property) and then call the super method.  I'm thinking this is the best solution.

If you don't do that or something similar, at minimum I'd place a warning comment in the javadocs for ExtendedPropertySourcesPlaceholderConfigurer making it clear that there is a hazard if you don't set ignoreResourceNotFound=true (a missing file would be ignored instead of failing bean creation) and also place a warning in its current use in applicationContext.xml.  That of course wouldn't be sufficient, but at least there would be some visibility/acknowledgement of the issue.  By the same token if we end up on this path, I'd rename ExtendedPropertySourcesPlaceholderConfigurer to ExtendedPropertySourcesPlaceholderConfigurerIgnoresMissingResources so it is more clear.

drewwills:
@jameswennmacher -- Documentation of loadProperties is sparse;  I had to look at the source to understand it's role in the class:  http://grepcode.com/file/repository.springsource.com/org.springframework/org.springframework.core/3.2.2/org/springframework/core/io/support/PropertiesLoaderSupport.java#PropertiesLoaderSupport.loadProperties%28java.util.Properties%29

There is an issue:  the locations member variable is private, and there is no getLocations() method.  I'd have to overload setLocations(), hold onto the collection, then overload loadProperties() and call setLocations _within that_ before calling super.loadProperties().  This approach has a certain code smell.

Better, I think, is to take ignoreResourceNotFound out of the xml entirely and call setIgnoreResourceNotFound(true) inside the constructor... and document that this impl is ALWAYS ignoreResourceNotFound=true.

jameswennmacher:
I see.  Sounds good.  I do suggest renaming the class then to make it painfully obvious. :smile:

drewwills:
I renamed it, but I thought the example name was even more painful than obvious (though it was obvious).  I added a big, descriptive javadoc comment at the class level.

bjagg:
+1 for descriptive javadocs!

jameswennmacher:
@drewwills what can I say.  I have a greater tolerance/love for long names than most people. :smile:

jameswennmacher:
+1 on the recent changes.

PR #635, opened 2016-03-15
UP-4648:  Improve performance of EntityPersonAttributesGroupStore by …

drewwills:
…caching the results of evaluating membership

https://issues.jasig.org/browse/UP-4648

@jameswennmacher -- need to discuss cache keys for guests.


jameswennmacher:
Overall looks good. You are correct that the cache key calculation for the guest account will be a problem if we continue down the path of trying to have different memberships for guest based on dynamic attribute values (passed in on the URL for having a different guest experience based on request attribute values; e.g. allow guest with cccMisCode=xxx or guest with nativeClient=true to support native mobile client access).

drewwills:
@jameswennmacher - I was hoping you'd maybe make some suggestions for key generation that would work with that use case.  I think the EntityIdentifier for the parent group is fine, even for CCC.

jameswennmacher:
I do have an idea which I've partially implemented in another branch. The gist of it is a class which compares the name to the current user the request is executing for and if a match uses a method of the person that takes into account the person's attribute values to calculate a hash and if not just calculates the hash based on the username.  That code is not in Apereo yet.  

I figured when I got back to trying to handle that situation, hopefully in a week or so, I could incorporate that into your submission.

PR #639, opened 2016-03-22
UP-4650 Add in originalUsername field

timlevett:
Adds in an originalUsername to the `/uPortal/api/session.json` call in the event you are swapping to another user so that you can do some validation on the client.

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/rest/SessionRESTController.java
+++ b/uportal-war/src/main/java/org/jasig/portal/rest/SessionRESTController.java
@@ -30,6 +30,7 @@ import javax.servlet.http.HttpSession;
 import org.jasig.portal.events.IPortalEventFactory;
 import org.jasig.portal.security.IPerson;
 import org.jasig.portal.security.IPersonManager;
+import org.jasig.portal.security.IdentitySwapperManager;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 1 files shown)


uportal-war/src/main/java/org/jasig/portal/rest/SessionRESTController.java
```diff
@@ -81,6 +88,13 @@ public ModelAndView isAuthenticated(HttpServletRequest request, HttpServletRespo
             attributes.put("displayName", person.isGuest() ? "Guest" : person.getFullName());
             attributes.put("sessionKey", person.isGuest() ? null: key); //only provide keys to non guest users
             attributes.put("version", uPortalVersion);
+            if(swapperManager != null) {
```
drewwills: Doesn't this approach create a race condition?  IIRC you can't count on the order in which dependencies are injected into beans by the app context.  Would it be better to handle this task in a @PostConstruct initialization method?

drewwills:
How about a JIRA ticket?

jameswennmacher:
+1.  Also agree with the benefit of a Jira ticket. :smile:

timlevett:
Jira added, thanks for the reminder

uportal-war/src/main/java/org/jasig/portal/rest/SessionRESTController.java
timlevett: This if statement `should` always be true because of the `required = true` in the annotation. More of a paranoid programmer issue :).

uportal-war/src/main/java/org/jasig/portal/rest/SessionRESTController.java
```diff
@@ -58,6 +60,11 @@ public void setPersonManager(IPersonManager personManager) {
         this.personManager = personManager;
     }
     
+    @Autowired(required = true)
```
apetro: I haven't looked -- is an `IdentitySwapperManager` always available in all uPortal implementations, or is the way I disable identity swapping to fail to instantiate an `IdentitySwapperManager`?

drewwills:
:+1:

PR #640, opened 2016-03-22
UP-4649:  Remove dead/unused code identified using tools (UCDetector)

drewwills:
https://issues.jasig.org/browse/UP-4649

## Trimming Old, Unused Code

I'd like to identify & remove uPortal code that is no longer in use.  Given the 15-year history of the project, I suspect there's more than a little of it.

The presence of this (unnecessary) code has the potential to impose a maintenance burden over time, and it has the potential to inhibit desired refactoring.

UCDetector identifies a large number of files, but the list is 90% false-positives.  As you may imagine, it's difficult for UCDetector to see how classes might be used in SpEL expressions, ant tasks, groovy sources, etc.  I am eliminating these possibilities (afaik) before attempting to remove a file.  I believe UCDetector can also detect intra-class unused code... I just haven't gotten that far.  (I have now... see below.)

I'd like to use this PR as a public place to discuss the files I can identify, as well as best practices around removing whatever we decide to remove.

### Deprecation?

Here's an important question related to this work:  is there any of this (removed code) that should be deprecated instead of removed?  I'd love to hear the thoughts of others, but (for my part) I don't think many -- if any -- of the items here should be deprecated.  I think we should simply remove them.  

Deprecation would be appropriate if we were concerned about custom local code referencing these items.  That would have been an immense concern in the days of IChannels, but (IMHO) it's been a long time since uPortal adopters did much custom Java code inside uportal-war.

Besides, the items identified by this process are (for the most part) dangerous to keep.  They contain code that hasn't been used or maintained in years -- there's every chance they don't function properly in 2016.  None of the removed code so far, moreover, has had any unit test coverage.


timlevett:
Neat, interested to learn more about this. I would caution with the REST controllers as they are used heavily by us, but this tool would probably flag them as unused.

drewwills:
@timlevett -- It does.  Not just the REST controllers, even the ones for Portlet MVC.  So far I have ignored any result that ends in "Controller."

I also invest extra sleuthing time into classes with annotations (especially) and classes that implement interfaces.

jameswennmacher:
Great!  I'd love to see this evolve.

uportal-war/src/main/java/org/jasig/portal/portlets/account/selfedit/AccountPasswordFormValidator.java
```diff
@@ -1,100 +0,0 @@
...
-import org.springframework.binding.message.MessageContext;
-import org.springframework.stereotype.Component;
-
-/**
- * 
- * @author Jen Bourey, jbourey@unicon.net
- * @version $Revision$
- */
-@Component("accountPasswordFormValidator")
-public class AccountPasswordFormValidator {
```
drewwills: I have some anxieties about this one.  I dimly remember that Spring form binding & validation support may include some kind of magic for finding (and invoking) validators.  But there's one issue with that, at least:  the HTML form for updating a password doesn't even include (currently) a box for the old password.  Also, this class doesn't implement any Spring interfaces... how would the magic know which method to call?  Anything starting with "validate*"?

uportal-war/src/main/java/org/jasig/portal/portlets/permissionsadmin/PermissionDefinitionFormValidator.java
```diff
@@ -1,44 +0,0 @@
...
-import org.springframework.stereotype.Component;
-
-/**
- * Validator for the permission editing subflow.
- * 
- * @author Jen Bourey, jbourey@unicon.net
- * @version $Revision$
- */
-@Component
-public class PermissionDefinitionFormValidator {
```
drewwills: See comment WRT AccountPasswordFormValidator.java.

drewwills:
## Update

UCDetector is producing 4 kinds of warning:

* 0 references
* Test only
* Visibility:  Use default (wants you to change class/method/filed visibility to default)
* Visibility:  Use private (wants you to change class/method/filed visibility to private)

Within each of these, there are sub-lists for classes, fields, enums, methods, constants, etc.

I have made it completely through the list of Classes with 0 references.  The vast majority were false-positives (e.g. controller classes and other String-managed beans, things used in webflow, things used in Import/Export, etc.)

I have scratched the surface of the list of methods with 0 references.  There are loads of these, and few of them are false-positives.  I don't think I can make it entirely through this sub-list -- or the warnings generated by UCDetector generally -- in one go.   I think we should approach this task iteratively.

This pull request is already very large, and difficult to review.  I may add a bit more, but I think we should concentrate on finalizing this round at about where it is.

bjagg:
Excellent! Thanks Drew for this extra effort.

I agree with your deprecation assessment. This code just needs to be removed.

uportal-war/src/main/java/org/jasig/portal/io/PortletPreferenceOverridePhrase.java
```diff
@@ -1,62 +0,0 @@
...
-import org.danann.cernunnos.Formula;
-import org.danann.cernunnos.Phrase;
-import org.danann.cernunnos.Reagent;
-import org.danann.cernunnos.ReagentType;
-import org.danann.cernunnos.SimpleFormula;
-import org.danann.cernunnos.SimpleReagent;
-import org.danann.cernunnos.TaskRequest;
-import org.danann.cernunnos.TaskResponse;
-
-public class PortletPreferenceOverridePhrase implements Phrase {
```
apetro: Should I be concerned that this is a phrase that could be used in some `.crn` script that would not have been caught by the dead-code-detection tooling?

jameswennmacher:
I took a quick look.  +1 on the merge.

uportal-war/src/main/java/org/jasig/portal/io/PortletPreferenceOverridePhrase.java
drewwills: @apetro -- Sort of.  I'm glad you asked, because I want this part to be broadly known & abundantly clear...

Before ANYTHING was removed, multiple layers of diligence were applied to verify that the item is genuinely unused.  The *first* of these layers was almost always a '$grep -R "{item}" uportal-war/src/main'.  Even that was often followed by a '$mvn clean' plus '$grep -R "{item}" .'.  Any Cernunnos Phrase that is in use would have generated a hit in uportal-war/src/main.

In addition to that, the following diligence steps were applied to everything in the PR (sometimes in batches):
* $ant clean initportal // NB:  tests are run
* rm -rf export/
* $ant data-export -Ddir=export // Export every object in the portal
* $ls export // Verify that every entity type seems to be exporting as usual
* // Start the portal, log in as admin, faculty, staff, and student
* // Verify the portal log generates no additional WARNings

I used other tactics beyond these.  Anything with Spring annotations was treated with the utmost suspicion (I'm not sure there was any of that removed, in the end).  Classes that implement interfaces (or abstract classes) were treated with additional suspicion.

So in a way, yes:  if these things had not been done, you should have been deeply worried.  This PR would not have passed the CI checks if these steps had not been taken.

timlevett: Good detail @drewwills ! Looks really good. An additional step could be:

```
rm -rf data
ant clean initportal
```

I did see the initportal in there, but doing the rm -rf of data forces it to recreate the schema, which has an additional effect to what you listed.

drewwills:
@timlevett -- I followed your suggestion

```
rm -rf data
ant clean initportal
```

No issues.  I'm going to press the green button now.

jgribonvald:
I think you removed some utils class that are possible bean injection.

As example you removed uportal-war/src/main/java/org/jasig/portal/spring/beans/factory/MediatingFactoryBean.java which is a usefull database access selector between JNDI and JDBC, watch on the documentation : https://wiki.jasig.org/display/UPM43/Using+JNDI+managed+DataSources

If you have a better example of database configuration, or if jndi access is discouraged tell me ;)

Thanks

PR #641, opened 2016-03-23
UP-4643:  Provide a fix for bundled webapps suitable for the 4-2 line that is different, in necessary ways, from master

drewwills:
Fixed in Calendar, News Reader, Bookmarks, CAS, Widgets, FunctionalTests, Email, Simple CMS, and Weather


jameswennmacher:
What needed to be different from master?

drewwills:
@jameswennmacher -- NRP is on version 4.x on master, and it's not a drop-in replacement for 3.x.  (You have to rework your data.)

PR #642, opened 2016-04-05
UP-4654:  Update 'hover' chrome in the tab/column area to work better on mobile devices

drewwills:
https://issues.jasig.org/browse/UP-4654

## Before

before

## After

after


jameswennmacher:
+1.  Looks great.

PR #643, opened 2016-04-06
UP-4655:  Add some TRACE and DEBUG log entries at key milestones for understanding performance of some key requests

drewwills:
https://issues.jasig.org/browse/UP-4655


uportal-war/src/main/java/org/jasig/portal/spring/security/preauth/PortalPreAuthenticatedProcessingFilter.java
```diff
@@ -142,14 +142,22 @@ public void setLoginPath(String loginPath) {
     public void setLogoutPath(String logoutPath) {
         this.logoutPath = logoutPath;
     }
-    
+
     @Override
     public void doFilter(ServletRequest request, ServletResponse response,
             FilterChain chain) throws IOException, ServletException {
 
+        // Set up some TRACE logging for performance troubleshooting
```
drewwills: This approach, as-is, is not great.  We should focus on logging only for /Login, maybe one or two other key requests.  (There's too much noise.)

PR #645, opened 2016-04-09
UP-4483:  Looks like a cache for PersonAttributesGroupDefinitionImpl.…

drewwills:
…members may have been overlooked

https://issues.jasig.org/browse/UP-4483

Do we need a cache here as well?

The table that backs this collection is one of the high-usage tables identified in the graphs.  I have a newer graph from Manchester -- including the previous updates for this ticket -- that identifies this table (and only this table).


bjagg:
We likely do. It was skipped due to some time constraints.

drewwills:
Awesome -- I've now conferred with both @bjagg and @jameswennmacher on this topic, and I've seen the results of applying this change on an IlliniCloud dev server as well as a Manchester dev server.  I've removed 'WIP' from the title and I'm going to merge.

PR #646, opened 2016-04-09
UP-4657:  Portlet Manager -- deleting a portlet does not remove it fr…

drewwills:
…om categories or update permissions

https://issues.jasig.org/browse/UP-4657

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/io/xml/portlet/PortletDefinitionImporterExporter.java
+++ b/uportal-war/src/main/java/org/jasig/portal/io/xml/portlet/PortletDefinitionImporterExporter.java
@@ -59,6 +59,7 @@ import org.jasig.portal.portlet.om.IPortletType;
 import org.jasig.portal.portlet.om.PortletCategory;
 import org.jasig.portal.portlet.om.PortletLifecycleState;
 import org.jasig.portal.portlet.registry.IPortletCategoryRegistry;
+import org.jasig.portal.portlet.registry.IPortletDefinitionRegistry;
 import org.jasig.portal.portlet.registry.IPortletTypeRegistry;
 import org.jasig.portal.security.*;
 import org.jasig.portal.services.AuthorizationService;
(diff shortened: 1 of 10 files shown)


jameswennmacher:
Two thoughts:
1. The IPortletPublishingService implementation is called PortletDefinitionImporterExporter. I suggest making it more generic to fit your use case.
2. Could PortletDefinitionImporterExporter.removePortletDefinition also be enhanced to remove the portlet from layouts? That would be VERY nice.

drewwills:
### Copied from JIRA ticket comments...

@jameswennmacher,

Can you provide an example of what you mean for #1? (There are a series of *ImporterExporter classes that cover the various portal objects supported by ImportExport. It's not a part of this PR, or anything I've set up. It just happens that the concrete IPortletPublishingService class is also the PortletDefinitionImporterExporter... because the logic just overlaps, so far.)

Regarding #2 – I would like that very much, but I'm slightly afraid of what that would look like... it wouldn't be remotely safe to execute a DELETE FROM up_layout_struct WHERE... (I believe you'd have to DOM-up each layout first in order to remove a portlet).

jameswennmacher:
RE 1st item I guess it's OK that PortletAdministrationHelper's call to portletPublishingService.removePortletDefinition(def, person); invokes PortletDefinitionImporterExporter since PortletAdministrationHelper accesses an interface.  It just struck me as a bit weird on first consideration.

I can see deferring 2nd item to a separate PR.   It will be a bit more complicated but I don't think tremendously more.  You'd have to find the nodes you'll want to delete and back-track to its siblings or parents to adjust the NEXT_STRUCT_ID or CHLD_STRUCT_ID values so I don't think it would be that bad but perhaps I need to think about it more.

drewwills:
@jameswennmacher,

We would also need to take a quick review of rows that figure in XML structures like dlm:change-set, dlm:move-set, etc.  Off the top of my head, I can't say what the impact on them would be.

drewwills:
@jameswennmacher pointed out we should probably look into this issue:  https://issues.jasig.org/browse/UP-4247

PR #648, opened 2016-04-20
UP-4626 Add <portlet> type to permission targets, fix export

jameswennmacher:
Fix issue in import/export.  Currently an export creates permission-set XML files that list the ID number of the portlet so these files will not import properly.

Old export output:
```
<?xml version="1.0" encoding="UTF-8"?><permission-set script="classpath://org/jasig/portal/io/import-permission_set_v3-1.crn">
  <owner>UP_PORTLET_SUBSCRIBE</owner>
  <principal-type>org.jasig.portal.groups.IEntityGroup</principal-type>
  <principal>
    <group>Tenant Administrators</group>
  </principal>
  <activity>BROWSE</activity>
  <target permission-type="GRANT">
    <literal>PORTLET_ID.1</literal>
  </target>
  <target permission-type="GRANT">
    <literal>PORTLET_ID.10</literal>
  </target>
  <target permission-type="GRANT">
    <literal>PORTLET_ID.15</literal>
  </target>
</permission-set>
```

With fix:
```
<?xml version="1.0" encoding="UTF-8"?><permission-set script="classpath://org/jasig/portal/io/import-permission_set_v3-1.crn">
  <owner>UP_PORTLET_SUBSCRIBE</owner>
  <principal-type>org.jasig.portal.groups.IEntityGroup</principal-type>
  <principal>
     <group>Tenant Administrators</group>
  </principal>
  <activity>BROWSE</activity>
  <target permission-type="GRANT">
    <portlet>fragment-admin</portlet>
  </target>
  <target permission-type="GRANT">
    <portlet>instructions-for-new-tenants</portlet>
  </target>
  <target permission-type="GRANT">
    <portlet>portlet-admin</portlet>
  </target>
</permission-set>
```

Also with new code during import if a literal element is detected that starts with PORTLET_ID., an error is thrown with the message:

	... 79 common frames omitted
Caused by: java.lang.RuntimeException: Invalid literal string PORTLET_ID.39 in file:/home/jwennmacher/projects/uPortal/uPortal/perms2/permission-set/Authenticated_Users__BROWSE__UP_PORTLET_SUBSCRIBE.permission-set.xml. This is probably due to export of permission using an older version of uPortal doing an export that didn't support the new <portlet> element.  You will need to lookup the portlet with the ID value '39' in the UP_PORTLET_DEF table and replace the <literal> element in the XML file with <portlet>portletFname</portlet> then re-import.
	at org.jasig.portal.io.GroupKeyLiteralOrPortletPhrase.evaluate(GroupKeyLiteralOrPortletPhrase.java:101)

With these changes I didn't see a need to bump the permission-set version number as it catches the invalid data and provides instructions on remedy.  There is no way to 'convert' an incorrect xml file to correct data.

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/io/GroupKeyLiteralOrPortletPhrase.java
+++ b/uportal-war/src/main/java/org/jasig/portal/io/GroupKeyLiteralOrPortletPhrase.java
@@ -1,0 +1,112 @@
+/**
+ * Licensed to Apereo under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work
+ * for additional information regarding copyright ownership.
+ * Apereo licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License.  You may obtain a
+ * copy of the License at the following location:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.jasig.portal.io;
+
+import org.danann.cernunnos.Formula;
+import org.danann.cernunnos.Reagent;
+import org.danann.cernunnos.SimpleFormula;
+import org.danann.cernunnos.TaskRequest;
+import org.danann.cernunnos.TaskResponse;
+import org.dom4j.Element;
+import org.jasig.portal.portlet.dao.jpa.PortletDefinitionImpl;
+import org.jasig.portal.portlet.om.IPortletDefinition;
+import org.jasig.portal.portlet.registry.IPortletDefinitionRegistry;
+import org.jasig.portal.security.IPermission;
+import org.jasig.portal.spring.locator.ApplicationContextLocator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Class that parses an XML element to determine if it is
+ * <group>groupName</group>
+ * <portlet>fname</portlet>
+ * <literal>usernameSpecialPermissionStringEtc</literal>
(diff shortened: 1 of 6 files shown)


drewwills:
@jameswennmacher -- I have a feeling we discussed this point before, but it's been a while...

Why do we need to export (or import) a permission-set.xml file specifying the BROWSE permission for individual portlets?  (I could see it for categories... but they wouldn't exhibit this issue.)

A portlet will export (and import) in an XML format that includes...

```
    <group>Everyone</group>
    <permissions>
        <permission system="UP_PORTLET_SUBSCRIBE" activity="BROWSE">
            <group>Everyone</group>
        </permission>
    </permissions>
```

Can't we just treat BROWSE the way we always have SUBSCRIBE -- a matter that's handled in the portlet-definition.xml file?

jameswennmacher:
Ahhh.  I didn't realize that the export of the portlet definition put the permissions for BROWSE in it.  That's good to hear.  I didn't think to check for that since I saw it in the export permission set.

So we need to change export to ignore the permission set if the permission is BROWSE and the target is a portlet since it is handled in the portlet definition (we could allow BROWSE to a category though its not recommended at this point since it works best on a portlet -- the quickstart data set had that at one point though I know we are pulling it out. Some institutions may still have it in their data set).  If the target is a portlet the export should use <portlet>fname</portlet> for the target for consistency.

The <portlet>fname</portlet> aspect of the import should still be supported so you could specify a user or group has CONFIG permission on a portlet by fname (TAMU needed that).

Agreed?

drewwills:
AFAIK granting BROWSE on a category is great if that's what you intend;  I believe the issues with the quickstart_data essentially stemmed from the fact that the permission-set.xml files w used to ship were not really what people would probably intend, and were not a good match generally.

There is still an issue with how the marketplace only allows you to see portlets "By Category" for categories you have BROWSE over.

jameswennmacher:
Didn't realize that's how Marketplace worked.  Marketplace should change IMHO to just offer the categories that are available for the portlets you have access to, just like Customize does (I think https://issues.jasig.org/browse/UP-4086 might be the issue to address that).

drewwills:
Fixed per discussion.

PR #652, opened 2016-05-01
UP-4622:  Fix a display issue with the offcanvas menu on iOS

drewwills:
This PR corrects (I hope) an issue with the offcanvas menu not appearing on iOS.  The trouble was with the use of 'position:fixed' -- apparently this value works somewhat differently on iOS.

Thanks @cousquer for figuring it out.

### DIFF
--- a/uportal-war/src/main/webapp/media/skins/respondr/common/less/navigation.less
+++ b/uportal-war/src/main/webapp/media/skins/respondr/common/less/navigation.less
@@ -19,8 +19,6 @@
 
 .portal-nav {
     background: @navbar-background-color;
-    position : relative;
-    z-index  : 5px;
 
     .sidebar-offcanvas {
 
@@ -210,7 +208,8 @@
 
 }
 
-/* TODO:  this one is defined backwards;  we normally define "mobile first" and use media queries to alter styles for larger displays */
+/* TODO:  this one is defined backwards;  we normally define "mobile first" and
+   use media queries to alter styles for larger displays */
 @media only screen and (max-width: 769px) {
 
     /*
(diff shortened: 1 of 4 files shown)


cousquer:
@drewwills ,  So this resolves the problem on old IOS (iphone 4) in safari ? 

Oh my, when I suggested you the fix, I looked at the code and I had tell me in my mind : "if I was a safari browser on a old version of IOS, how would I react". no test on local, nor any test on iphone 4 (I don't have a iphone 4)...

drewwills:
I don't have any iOS devices at all.  Unicon has some for just this purpose (testing), and several folks at Unicon have their own.

I used the Simulator tool that comes with [Apple XCode](https://itunes.apple.com/us/app/xcode/id497799835?mt=12).  The issue was there (in Simulator) when I started this process, so I hope the fact that it _wasn't_ there with this patch applied means it won't on those devices.

drewwills:
The commit from @cparaiso makes some of the cosmetic stuff better.  The carrot for expanding the flyout menu is the wrong color on the tab you're currently using.  (It's blue-on-blue.)

uportal-war/src/main/webapp/media/skins/respondr/common/less/navigation.less
```diff
@@ -17,347 +17,299 @@
...
                     }
-                }
-                .active {
-                    .list-group-horizontal {
-                        .list-group-item {
-                            margin-bottom: 0;
+                    &.portal-navigation-add-item {
+                        padding: 8px 12px;
+                    }
+                    &:last-child {
```
drewwills: - [ ] Delete this selector

uportal-war/src/main/webapp/media/skins/respondr/common/less/navigation.less
```diff
@@ -17,347 +17,299 @@
...
+                            a.portal-subnav-link {
+                                width: 100%;
+                                min-width: 100%;
+                                color: @black;
+                                background-color: @white;
+                            }
+                            a {
+                                &:hover, &:focus {
+                                    background: @navbar-item-active-hover-background-color;
+                                    color: white;
```
drewwills: Shouldn't this value be a variable?

PR #658, opened 2016-05-10
UP-4473: Fixed component-scan exclude-filter regex expressions, addin…

groybal:
…g some new required 'portlets' includes.

I ran into an issue that stemmed from UP-4473, so spent some time addressing this.  The issue I had was that I was updating the Dynamic Skin portlet and had defined some beans in the portlet context file that needed to be autowired into a service class.  However, with the component scan exclude-filter misconfigured, the DynamicSkin @Service beans were being created outside of the portlet context, my new beans were not found since they are only defined at the portlet level, and I was getting a dependency resolution exception.

### DIFF
--- a/uportal-war/src/main/resources/properties/contexts/applicationContext.xml
+++ b/uportal-war/src/main/resources/properties/contexts/applicationContext.xml
@@ -33,15 +33,19 @@
     <aop:aspectj-autoproxy/>
     
     <context:annotation-config/>
+    <context:component-scan base-package="org.jasig.portal.portlets">
+        <context:include-filter type="regex" expression=".*\.PortletPreferencesJsonDao" />
+        <context:include-filter type="regex" expression=".*\.account\.UserAccountHelper" />
+    </context:component-scan>
     <context:component-scan base-package="org.jasig.portal">
-        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.portlets\."/>
-        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.security\.mvc\."/>
-        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.json\.rendering\."/>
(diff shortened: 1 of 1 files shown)


uportal-war/src/main/resources/properties/contexts/applicationContext.xml
```diff
@@ -33,15 +33,19 @@
...
-        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.redirect\."/>
-        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.rendering\."/>
+        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.portlets\..+"/>
+        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.security\.mvc\..+"/>
+        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.json\.rendering\..+"/>
+        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.layout\.dlm\.remoting\..+"/>
+        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.security\.remoting\..+"/>
+        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.rest\..+"/>
+        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.redirect\..+"/>
+        <context:exclude-filter type="regex" expression="org\.jasig\.portal\.rendering\..+"/>
```
jhelmer-unicon: Nitpick:  Generally, I think it's better to use a filter based on annotation type than by regex on the package name.   That way, you easily add Services anywhere and it will get picked up in the parent context.   Controllers can be placed anywhere and will get picked up in the portlet contex (not that you should!)

Having said that, I don't think there is anything here bad enough to hold up merge if the community is fine with this approach.

jhelmer-unicon:
Pulled down this PR and tested it locally.   From simple tests I did, I'm not seeing any issues with this.

Merging.

PR #661, opened 2016-05-13
sort flyout sub-menu alphabetical order

moonjeongkang:


timlevett:
I would suggest for this to be an option as a lot of organizations have very specific specifications when it comes to the order of the applications.

sragusa:
What if we change the comment on the line of code to indicate keeping the line of code or commenting out the line of code is a toggle/option?  Perhaps, the comment could say "Commenting out this line of code will order the flyout menu by the fragment layout.  Keeping this line of code will order the flyout menu alphabeticaly."  What are your thoughts?  It seems very simplistic, to comment out or leave the code - the same as changing a parameter.

cousquer:
Agree with @timlevett , it is perhaps a need for a language, perhaps a habit in some regions of the world-I do not know. But it goes a bit against the behavior of the portal. 
I would suggest for this sorting to make it optional.
I give you an example. you deploy a "Contact" tab with about 20 portlets. The portlet you want to put "forward" is called "Zimbra" (the mail access), So you need to put it at the first position in the first column. With alphabetizing the fly-out options  in the navigation, Zimbra portlet will be in twentieth position in the dropdown.
@moonjeongkang Why do you want to sort the options ? I'm really interested to learn cases which require this sorting.

bjagg:
Commenting this out would be an option. Having this driven by a property in portal.properties would be better.

Closing this PR w/o merging. Please re-submit when ready for another review.

PR #663, opened 2016-05-19
UP-4683: Data Import fix for Postgres with string to integer conversion.

groybal:
I tested fix against Postgres and HSQLDB, but not other databases.  I don't expect there would be issues.


gcarroll:
Hello everyone,

A quick note.  After having to rebuild my server, I attempted to build 4.3.0 source and have a couple issues.  

The first issue I have opened a jira ticket on which relates to a failure with dependencies with the CalendarPortlet.  I am pleased to say that I may have this fixed soon.  

I will post my workaround once I sufficiently test and ensure that it works. 

The second issue is one I discovered when I started a build with the CalendarPortlet removed, this broke the SimpleContentPortlet with my log4j being  higher version installed outside the version called within. I will look into this issue and will report further once I see the CalendarPortlet issue fixed.

Geordie Carroll
Interior Technology Inc

jhelmer-unicon:
:+1:   portlet_type is an int in the DB, so this conversion seems necessary.   I don't see anything particularly risky here.  If no objections, will probably merge soon.

PR #664, opened 2016-05-26
Changed jasig link to apereo on error page.

MichaelFrieze:


drewwills:
@MichaelFrieze -- Thanks for this contribution.  Seems like a no-brainer.  Are you willing to submit an ICLA?  https://www.apereo.org/licensing/agreements

asgrant:
@drewwills I believe that Mike submitted an ICLA via FAX a while ago. So he should be set, unless it wasn't received.

drewwills:
@asgrant and @MichaelFrieze -- I don't see Michael's name in the ICLA column here:  http://licensing.apereo.org/

I do, however, see Oakland's institutional agreement.

@apetro Do you know if that covers it?  Does anyone know of a reason we're not seeing it?

drewwills:
This omission was covered in the fix for UP-4718.

PR #666, opened 2016-06-08
Update session.json on backend to return timeout data.

andrewstuart:
This patch will add timeout information to the `session.json` API endpoint, so that smarter front-ends, which may not be interacting with the server whilst the user is still active on the page, can decide what the deadline is, and if/how to inform the user.

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/rest/SessionRESTController.java
+++ b/uportal-war/src/main/java/org/jasig/portal/rest/SessionRESTController.java
@@ -76,7 +76,7 @@ public class SessionRESTController {
         final ModelAndView mv = new ModelAndView();
         
         HttpSession session = request.getSession(false);
-        
+
         if (session == null) {
             response.setStatus(HttpServletResponse.SC_NOT_FOUND);
         }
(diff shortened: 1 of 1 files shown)


timlevett:
:+1:

bjagg:
FYI, I have a patch from a client coming soon that addresses some timeout shortcomings. Will keep an eye on this to make sure changes are compatible with this PR.

uportal-war/src/main/java/org/jasig/portal/rest/SessionRESTController.java
```diff
@@ -20,6 +20,7 @@
 
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.util.Date;
```
drewwills: Doesn't appear to be needed.

drewwills:
Seems sensible.

Are we at all concerned with adding more elements to this existing JSON structure?  Is this the kind of change that would make us want to rev the API version?  (I'm cool with "no" if that's acceptable... bumping versions is work.)

jameswennmacher:
IMHO I think keeping the same version is acceptable.  A 'smart' front end could/should have a default value if it doesn't get one in the data.

andrewstuart:
@drewwills My general opinion about JSON objects is that new properties should not be considered breaking, but changing property names or removing properties should.

It's my experience that most JSON libraries just ignore unknown properties when deserializing, though I think Jackson actually would break by default.

That said, I'm sure we'll want to consider the name, since if we decide later `timeoutMS` is ugly, changing it *would* be a breaking change.

vertein:
:+1: agree with @andrewstuart assesment on versioning.

jhelmer-unicon:
:+1:  I agree with @andrewstuart's take on versioning too.    This feels like such an innocuous change that I don't really see the harm in getting this merged.   Barring downvotes, will probably merge in the next 24 hours.

PR #668, opened 2016-06-09
UP-4685 Adds ServerProfileMapper

vertein:
Allows the profile of a user to be dictated by what server the user is coming from, so that the user can test different profiles based on server.

Came up for us when one of our system schools (UW-River Falls) wanted to test our new profile via a specific url (my.uwrf.edu) but still retain the current profile at the old host (my.wisconsin.edu).

Adds the ability to select servers based on regular expression.

This is turned off by default, but does include commented examples of how to use.

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/layout/profile/ServerProfileMapperImpl.java
+++ b/uportal-war/src/main/java/org/jasig/portal/layout/profile/ServerProfileMapperImpl.java
@@ -1,0 +1,90 @@
+package org.jasig.portal.layout.profile;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import javax.servlet.http.HttpServletRequest;
+
+import org.apache.commons.lang3.Validate;
+import org.jasig.portal.security.IPerson;
+import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


uportal-war/src/main/java/org/jasig/portal/layout/profile/ServerProfileMapperImpl.java
```diff
@@ -0,0 +1,90 @@
...
+          "Cannot get profile fname for a null request.");
+
+        String userName = person.getUserName();
+
+        final String userServerName = request.getServerName();
+
+        Matcher matcher = this.pattern.matcher(userServerName);
+
+        if(matcher.matches()){
+            logger.debug("User {} will be given {} profile because server {} matches regular expression {}",
```
drewwills: This message is clear & awesome.

drewwills:
👍 Looks very good.

timlevett:
👍

uportal-war/src/main/java/org/jasig/portal/layout/profile/ServerProfileMapperImpl.java
vertein: Thanks!

PR #669, opened 2016-06-10
UP-4686: Add new option to set Maven thread value for build. 

bjagg:
Updated antHelp.txt in docs/

https://issues.jasig.org/browse/UP-4686

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -1683,6 +1683,11 @@ IMPORTANT: The '-Dpattern=' property is no longer a regular expression and is a 
         <property name="goal8" value="-Djasig.ignore" />
         <property name="goal9" value="-Djasig.ignore" />
 
+        <condition property="threads" value="-T ${maven.threads}C">
+            <isset property="maven.threads" />
+        </condition>
+        <property name="threads" value="-Djasig.ignore" />
+
         <condition property="test.skip" value="-Dmaven.test.skip=true">
             <isset property="maven.test.skip" />
         </condition>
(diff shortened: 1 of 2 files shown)


jameswennmacher:
Based on your comments in about this setting (see https://groups.google.com/a/apereo.org/forum/#!topic/uportal-dev/GtcMaZn4D9c), could we instead default it to 2 rather than being not defaulted?

jameswennmacher:
Also please update the UP4.3 manual to mention the option of setting this value using MVN_OPTS.

drewwills:
I support James' suggestion, unless someone can point out a downside.

bjagg:
Waiting for some feedback on a good default. Someone at Oakland experimented and found 1.5 worked well for a default value.

I looked at MAVEN_OPTS. I could not get it to work with this parameter. Some searching exposed that only JVM parameters are allowed, unfortunately.

build.xml
```diff
@@ -1683,6 +1683,11 @@ IMPORTANT: The '-Dpattern=' property is no longer a regular expression and is a
         <property name="goal8" value="-Djasig.ignore" />
         <property name="goal9" value="-Djasig.ignore" />
 
+        <condition property="threads" value="-T ${maven.threads.per.cpu}C">
```
jameswennmacher: Propose changing this to
```
<condition property="threads" value="-T ${maven.threads.per.cpu}C" else="-T 2C">
```

bjagg: I received feedback that 1.5 was an optimal value from Oakland. It really depends on he server details.

Great idea! Using `else=` will make that section shorter and clear.

jameswennmacher:
+1 on merge

PR #670, opened 2016-06-16
UP-4689 Portlet worker threads get same SpringSecurityContext as HTTP thread

jameswennmacher:
https://issues.jasig.org/browse/UP-4689

Fixes issue noticed by CCC which uses Spring Security.  Portlet worker threads can inadvertently delete the Spring SecurityContext.  See issue for details.

This PR furthers uPortal being able to use Spring Security more extensively in the future without impacting the current portal behavior.


drewwills:
This change sounds like a good idea.

jhelmer-unicon:
:+1:

xdhmoore:
For the sake of posterity, without this fix you can have a race condition where one thread deletes the context (by calling close() on the outputstream, for example) and causes another concurrent thread to interpret the empty context as a new authentication, triggering the creating of a new CSRF token, which will cause links using the old one to break.

PR #671, opened 2016-06-16
UP-4690: Add CalDav4J artifacts to CalendarPortlet overlay. 

bjagg:
This should be removed once the artifact is available in Maven Central.

https://issues.jasig.org/browse/UP-4690


bjagg:
Merging this now to fix broken build

PR #672, opened 2016-06-16
UP-4691 Adds method to reset user's layouts on all profiles

vertein:
Currently, we have two profiles at Wisconsin.  We have "default" (universality/muniversality) and our new profile "bucky".

When a user clicks reset-my-layout, their "default" resets, but not their "bucky" layout.  So, this adds a method so that if you click that button, all your layouts get reset.  This is off by default, but has a comment on how to turn it on.

I'm guessing this was also a case when we had both respondr and univerality/muniversality.

I'll also mention, that yes, we will be moving away from universality/muniversality soon, so no judging on that.


timlevett:
👍

vertein:
Merging with 2 +1's

PR #673, opened 2016-06-20
NOJIRA: Adding pooled ldap property as configurable

jgribonvald:
It's more an easy to use and to avoid disagreement after rebase, in some install context we need to define this property depending on context of deploiement (test or production)

### DIFF
--- a/filters/data-test.properties
+++ b/filters/data-test.properties
@@ -92,3 +92,4 @@ environment.build.ldap.url=
 environment.build.ldap.baseDn=
 environment.build.ldap.userName=
 environment.build.ldap.password=
+environment.build.ldap.pooled=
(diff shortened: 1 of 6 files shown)


jgribonvald:
Just for feedback : Is there someone who keep pooled option always to false ?
I didn't watch if there is really a gain, but normaly there is...

jameswennmacher:
IMHO I think you should have the 'default' value in the various filters/*.properties files set to true for pooled connections.  I can't think of any reason why LDAP connections should not be pooled.  Anyone else know of a historic reason why the ldapContext.xml had pool connection defaulted to false?

jgribonvald:
So I've found why it's setted to false.

If you define your Ldap connection with the class org.springframework.ldap.pool.factory.PoolingContextSource the property on the bean defaultLdapContext_target should be set to false !

jgribonvald:
Setting the option pooled to true should be only when you use a custom configuration and not when using pooling bean definition purposed.

So I think my PR is useless !

Sorry I tried to report my last config from older uPortal version, but it's useless when using org.springframework.ldap.pool.factory.PoolingContextSource

PR #675, opened 2016-06-20
NOJIRA: Possibility setting env from build.properties

jgribonvald:
Adding the possibility to set the var env from build.properties

### DIFF
--- a/build.properties.sample
+++ b/build.properties.sample
@@ -71,3 +71,4 @@ cleanShared=true
 
 # Prompt the user before running ant tasks that could modify the database
 prodPrompt=false
+#env=local
--- a/build.xml
+++ b/build.xml
@@ -30,6 +30,9 @@
      | Determine if a named environment has been specified in the command.  
      | Examples:  dev, test, prod, local, etc.
      +-->
+    <!-- Load build properties here to be able to load the env property if defined -->
+    <property file="${basedir}/build.properties" />
+
     <condition property="environment.token" value="${env}">
         <isset property="env" />
     </condition>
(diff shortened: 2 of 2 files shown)


drewwills:
I'd love to support this option.  When I proposed this change a few years ago, it was rejected because (apparently) there isn't a way to get Maven to read it from there, and running $mvn commands (without Ant) is also supported.

jgribonvald:
Thanks Drew - 

To argument so running ant target and maven are for me totaly different. If this property is read from ant it can be applied like this, but if you run only from maven you need to specify it, in my mind it's two differents way. Ant is more an "easy to use mvn commands" so passing directly the env is in my mind a good thing and these properties will be passed to mvn called. After if the user want to use mvn directly i don't think that we have to try to use properties to set the env, direct mvn commands are for customs/user use (more like "doing what I want"), and so not/less for production use and authorized to everyone... 

Else if someone doesn't like this way, going on gradle will solve this problem...

Also this commit is a change done for a long time in the ESUP packaging. Packaging that I try to reduce in reporting here some PR, we would prefer to make it in an other way. We would prefer to provide commit/branch examples on how to customize uPortal than having a ESUP packaging, and so if some PR could be integrated here it would be great for us... It's complicated to maintain a translation and to adapt to a French context some parts, more if you follow each version of uPortal, etc...

jameswennmacher:
I support including this change to make ant easier to use, and avoid an error on a dev or test system (I don't set this in prod just to make it more difficult to accidentally do an ant initdb :smile: ).  

When running MVN, you can always specify the -Denv=xxx or have it in MAVEN_OPTS (or ANT_OPTS) which is what I've done on customer systems.  I set it in the ~/.bashrc for the account you sudo to when working on the portal in a dev/test(/prod?) system.

bjagg:
Merging since this does not change any existing build approach.

PR #678, opened 2016-06-21
UP-4694 added Ldap Query escape function to prevent LDAP Injection at…

mgillian:
…tack.  Route to James Wennmacher for review.

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/utils/LdapQueryUtils.java
+++ b/uportal-war/src/main/java/org/jasig/portal/utils/LdapQueryUtils.java
@@ -3,11 +3,24 @@ package org.jasig.portal.utils;
 public class LdapQueryUtils {
 
 	/**
-	 * 
-	 * @param query
-	 * @return
+	 * EscapeLdapSearchFilterTerms takes a string that represents a term in an
+	 * LDAP query and escapes any LDAP special characters that could be used to
+	 * perform an LDAP injection attack.  
+	 * <p>For example, a valid term might be "baseball", and a valid LDAP query
+	 * might be "(cn=baseball)".  We must prevent a malicious or accidental 
+	 * attempt to send special characters that would subvert the intended query.  
+	 * If, for example, the user sent a query term of "baseball)(cn=*)", we 
+	 * don't want to process the LDAP query of (cn=baseball)(cn=*);
+	 * this would potentially return additional information beyond what was 
+	 * intended.
+	 * <p>By escaping LDAP special characters, the query above would actually 
+	 * be (cn=baseball\\)\\(cn=\\*)".
+	 * <p>See <a href=https://docs.ldap.com/specs/rfc4515.txt>https://docs.ldap.com/specs/rfc4515.txt</a> for
+	 * additional information about LDAP special characters
+	 * @param query Search term that needs to have special characters replaced
+	 * @return Search term with special characters replaced with escaped versions.
 	 */
-	public static String escapeSpecialCharacters(String query) {
+	public static String escapeLdapSearchFilterTerms(String query) {
         // We need to escape regex special characters that appear in the query string...
         final String[][] specials = new String[][] {
                             /* backslash must come first! */
--- a/uportal-war/src/main/java/org/jasig/portal/utils/LdapQueryUtils.java
+++ b/uportal-war/src/main/java/org/jasig/portal/utils/LdapQueryUtils.java
@@ -14,34 +14,36 @@ public class LdapQueryUtils {
 	 * this would potentially return additional information beyond what was 
 	 * intended.
 	 * <p>By escaping LDAP special characters, the query above would actually 
-	 * be (cn=baseball\\)\\(cn=\\*)".
+	 * be (cn=baseball\\29\\28cn=\\2a\\29".
 	 * <p>See <a href=https://docs.ldap.com/specs/rfc4515.txt>https://docs.ldap.com/specs/rfc4515.txt</a> for
 	 * additional information about LDAP special characters
 	 * @param query Search term that needs to have special characters replaced
 	 * @return Search term with special characters replaced with escaped versions.
 	 */
-	public static String escapeLdapSearchFilterTerms(String query) {
-        // We need to escape regex special characters that appear in the query string...
-        final String[][] specials = new String[][] {
-                            /* backslash must come first! */
-                            new String[] { "\\", "\\\\"}, 
-                            new String[] { "[", "\\[" }, 
-                            /* closing ']' isn't needed b/c it's a normal character w/o a preceding '[' */
-                            new String[] { "{", "\\{" }, 
-                            /* closing '}' isn't needed b/c it's a normal character w/o a preceding '{' */
-                            new String[] { "^", "\\^" },
-                            new String[] { "$", "\\$" },
-                            new String[] { ".", "\\." },
-                            new String[] { "|", "\\|" },
-                            new String[] { "?", "\\?" },
-                            new String[] { "*", "\\*" },
-                            new String[] { "+", "\\+" },
-                            new String[] { "(", "\\(" },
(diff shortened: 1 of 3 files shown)


uportal-war/src/main/java/org/jasig/portal/utils/LdapQueryUtils.java
```diff
@@ -0,0 +1,34 @@
...
+                            new String[] { "^", "\\^" },
+                            new String[] { "$", "\\$" },
+                            new String[] { ".", "\\." },
+                            new String[] { "|", "\\|" },
+                            new String[] { "?", "\\?" },
+                            new String[] { "*", "\\*" },
+                            new String[] { "+", "\\+" },
+                            new String[] { "(", "\\(" },
+                            new String[] { ")", "\\)" }
+                        };
```
jameswennmacher: I think the above is a bit overkill.  Per section 4 of https://docs.ldap.com/specs/rfc4515.txt:
>   The <valueencoding> rule ensures that the entire filter string is a
>   valid UTF-8 string and provides that the octets that represent the
>   ASCII characters "*" (ASCII 0x2a), "(" (ASCII 0x28), ")" (ASCII
>   0x29), "\" (ASCII 0x5c), and NUL (ASCII 0x00) are represented as a
>   backslash "\" (ASCII 0x5c) followed by the two hexadecimal digits
>   representing the value of the encoded octet.

I believe only those characters need to be escaped, and the preceeding backslash AFAIK is an acceptable escaping mechanism.  The other characters are for RegEx-related comparisons which does not apply for an LDAP search filter term.  I don't think they would hurt, but it does add confusion IMHO so I'd remove the extraneous substitutions.

The above looks like a copy from https://github.com/Jasig/uPortal/blob/master/uportal-war/src/main/java/org/jasig/portal/groups/smartldap/SmartLdapGroupStore.java#L361.  The use cases are different.  In that case the keys (LDAP attribute names) are being compared using a regex expresssion so a regex substitution is reasonable.  However for LDAPGroupStore you are escaping the actual LDAP Search Filter.

uportal-war/src/main/java/org/jasig/portal/utils/LdapQueryUtils.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class LdapQueryUtils {
+
+	/**
+	 * 
+	 * @param query
+	 * @return
+	 */
+	public static String escapeSpecialCharacters(String query) {
+        // We need to escape regex special characters that appear in the query string...
```
jameswennmacher: Comment not appropriate.

uportal-war/src/main/java/org/jasig/portal/utils/LdapQueryUtils.java
jameswennmacher: Correction to above.  You should escape the characters specified above using \ (in Java Strings \\) plus 2 digit code which is what the spec says and is also what I've found on various other Google searches. Do not escape with just a backslash.  That is appropriate for the regex escaping but not an LDAP Search Filter escaping.

uportal-war/src/main/java/org/jasig/portal/utils/LdapQueryUtils.java
```diff
@@ -0,0 +1,34 @@
+package org.jasig.portal.utils;
+
+public class LdapQueryUtils {
+
+	/**
+	 * 
+	 * @param query
+	 * @return
+	 */
+	public static String escapeSpecialCharacters(String query) {
```
jameswennmacher: Please fill out the method JavaDoc with a good description.  Please reference the spec; i.e. section 4 of https://docs.ldap.com/specs/rfc4515.txt.

Also I suggest renaming the method to something more clear, such as escapeLdapSearchFilterTerms to make it clear it is for the search filter as other things, such as LDAP DN strings, require different escaping.

uportal-war/src/main/java/org/jasig/portal/utils/LdapQueryUtils.java
jameswennmacher: In looking at the spec more closely and looking at some other examples, such as section 'Specifying Non 7-Bit ASCII Characters in Search Filters' at https://docs.oracle.com/cd/E19424-01/820-4811/gdxpo/index.html where it states:
> Non 7-bit ASCII characters in search filters must be replaced with a representation of the character, where each byte > of the UTF-8 encoding is preceded by a backslash. In UTF-8, characters are represented by a hexadecimal code for > each byte.

> For example, the character é has UTF-8 representation c3a9. Thus, in a search filter, you represent é as \\c3\\a9. So, > to search for cn=Véronique Martin:

> ldapsearch -h myServer -b "dc=example,dc=com" "(cn=V\\c3\\a9ronique Martin)"

You will also need to look at each character and if it is above 0x7F represent it as \\XX (for 0x00XX) or \\XX\\YY for 0xXXYY.  This is required for internationalization.

jameswennmacher:
I would appreciate it if someone else, especially @cousquer or someone in France who has more experience with international character sets would comment on whether the LDAP search filter escaping is sufficient.  I'm pretty sure if we add escaping of characters above 0x7F that it accommodates international needs but a 2nd opinion is welcome.

See inline comments to the PR, along with section 4 of https://docs.ldap.com/specs/rfc4515.txt where it states
>   For AssertionValues that contain UTF-8 character data, each octet of
>   the character to be escaped is replaced by a backslash and two hex
>   digits, which form a single octet in the code of the character.  For
>   example, the filter checking whether the "cn" attribute contained a
>   value with the character "*" anywhere in it would be represented as
>   "(cn=*\2a*)".

>    As indicated by the <valueencoding> rule, implementations MUST escape
>    all octets greater than 0x7F that are not part of a valid UTF-8
>    encoding sequence when they generate a string representation of a
>    search filter.  Implementations SHOULD accept as input strings that
>    are not valid UTF-8 strings.  This is necessary because RFC 2254 did
>    not clearly define the term "string representation" (and in
>    particular did not mention that the string representation of an LDAP
>   search filter is a string of UTF-8-encoded Unicode characters).

cousquer:
@jameswennmacher I will have a look at the informations whether the LDAP search filter escaping is enough sufficient. I will contact the relevant people about it.
In french, we've got this kind of UTF-8 characters (look at the first one : "œ" : some people in France will enter indifferently  whether "œ" whether "oe" ... )
œuf, Véronique, à, bâtir,

cousquer:
@jameswennmacher, @mgillian, 
The documentation that you mention, said:

> implementations MUST escape all octets greater than 0x7F that are not part of a valid UTF-8 encoding sequence

So I do not think we should make escaping on accents, because the search string is supposed to be in utf8 ...

PHP now has its LDAP escaping function (http://php.net/manual/en/function.ldap-escape.php), there should have one in spring-ldap (http://projects.spring.io/spring-ldap/)

The only thing that we have found for java is this:

1. LDAP Injection Prevention Cheat Sheet (https://www.owasp.org/index.php/LDAP_Injection_Prevention_Cheat_Sheet )

2. Preventing LDAP Injection in Java (https://www.owasp.org/index.php/Preventing_LDAP_Injection_in_Java ) (outdated but looks like mgillian 's code)

Hope, it helps you.

cousquer:
http://docs.spring.io/spring-ldap/docs/2.1.0.RELEASE/reference/#use-spring-security

cousquer:
> In addition to simplifying building of complex search parameters, the LdapQueryBuilder and its associated classes also provide proper escaping of any unsafe characters in search filters. This prevents "ldap injection", where a user might use such characters to inject unwanted operations into your LDAP operations.

http://docs.spring.io/spring-ldap/docs/current/reference/
see for more details : 
http://docs.spring.io/spring-ldap/docs/current/reference/#query-builder-advanced

jhelmer-unicon:
Closing.   See:  https://github.com/Jasig/uPortal/pull/699

PR #679, opened 2016-06-23
NOJIRA: Solve I18N problems on xsl translated messages for javascript properties

jgribonvald:
When injecting translated messages into javascript (ECM) code from xsl you need to escape special ECM characters as xsl inject unescaped text.

As problem example set messages with simple quote (double simple quote in messages.properties to escape it) and watch on js property (options.messages.movePortletError as example, value generated in respondr.xsl from move.this.portlet.error message property). For this property the gallery isn't working anymore

To solve that, use a java specific function for xsl processors : upMsg:getMessageForEmacsScript, this function use commons.lang3 escaping function from EmacsScript (the library can provide also escaping for Json, HTML, etc...).

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/security/xslt/IXalanMessageHelper.java
+++ b/uportal-war/src/main/java/org/jasig/portal/security/xslt/IXalanMessageHelper.java
@@ -38,4 +38,12 @@ public interface IXalanMessageHelper {
     public String getMessage(String code, String language, String arg1, String arg2);
     
     public String getMessage(String code, String language, String arg1, String arg2, String arg3);
+
+    public String getMessageForEmacsScript(String code, String language);
+
+    public String getMessageForEmacsScript(String code, String language, String arg1);
+
+    public String getMessageForEmacsScript(String code, String language, String arg1, String arg2);
+
+    public String getMessageForEmacsScript(String code, String language, String arg1, String arg2, String arg3);
 }
--- a/uportal-war/src/main/java/org/jasig/portal/security/xslt/XalanMessageHelperBean.java
+++ b/uportal-war/src/main/java/org/jasig/portal/security/xslt/XalanMessageHelperBean.java
@@ -20,6 +20,7 @@ package org.jasig.portal.security.xslt;
 
 import java.util.Locale;
 
+import org.apache.commons.lang3.StringEscapeUtils;
 import org.jasig.portal.i18n.LocaleManager;
 import org.springframework.context.MessageSource;
 import org.springframework.context.MessageSourceAware;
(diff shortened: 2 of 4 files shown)


jgribonvald:
Sorry I was afraid of a possible problem, but not !
It was a wrong branch deployed locally, so i reopend the PR

jgribonvald:
Integrate UP- 4697 before this commit, it can be rebased after

cousquer:
@jgribonvald thanks :+1:

jgribonvald:
rebased to master ;)

jgribonvald:
This PR can be merged i won't rewrite it as it's working for us and will be a production in few time !

drewwills:
I just looked through this PR -- it looks like it addresses an important issue.

We should have a JIRA ticket for it;  we should be able to read about it and track when it was resolved.

uportal-war/src/main/java/org/jasig/portal/security/xslt/IXalanMessageHelper.java
```diff
@@ -38,4 +38,12 @@
...
     
     public String getMessage(String code, String language, String arg1, String arg2, String arg3);
+
+    public String getMessageForEmacsScript(String code, String language);
+
+    public String getMessageForEmacsScript(String code, String language, String arg1);
+
+    public String getMessageForEmacsScript(String code, String language, String arg1, String arg2);
+
+    public String getMessageForEmacsScript(String code, String language, String arg1, String arg2, String arg3);
```
drewwills: I wish there weren't 4 of these.  From the looks of things, only the first one is actually used.

uportal-war/src/main/java/org/jasig/portal/security/xslt/XalanMessageHelperBean.java
```diff
@@ -18,13 +18,14 @@
...
 
-import java.util.Locale;
-
+import org.apache.commons.lang3.StringEscapeUtils;
 import org.jasig.portal.i18n.LocaleManager;
 import org.springframework.context.MessageSource;
 import org.springframework.context.MessageSourceAware;
 import org.springframework.stereotype.Service;
 
+import java.util.Locale;
```
drewwills: IntelliJ Idea?  (Looks like imports were re-ordered.)

jgribonvald: yes, sorry sometimes I use IntelliJ Idea... (mainly when developping javascript...)

If you want I rewrite the PR removing this re-order ?

uportal-war/src/main/java/org/jasig/portal/security/xslt/IXalanMessageHelper.java
jgribonvald: Yes only first is used, but it was to match the signature to others methods as i seen others used (not all)...

uportal-war/src/main/java/org/jasig/portal/security/xslt/XalanMessageHelperBean.java
jgribonvald: yes there is on re-import that i can remove, but the import org.apache.commons.lang3.StringEscapeUtils; is usefull

jgribonvald: I rewrote my code for the order don't change

drewwills:
Looks like the small details (imports) have been cleaned up;  I'm going to build this change locally, then merge.

drewwills:
Tested locally -- looks good.

drewwills:
https://issues.jasig.org/browse/UP-4721

PR #682, opened 2016-06-25
WIP: Add weakKeys() to TagTrackingCache

bjagg:


uportal-war/src/main/java/org/jasig/portal/jgroups/protocols/DAO_PING.java
```diff
@@ -70,20 +73,24 @@ public boolean isDynamic() {
...
         final Future<?> wf = writer_future;
-        if(wf != null) {
+        if (wf != null) {
+            // Shutdown timer
             wf.cancel(false);
-            writer_future=null;
+            writer_future = null;
         }
-        
+        // TODO? Remove local_addr from table
```
jameswennmacher: We should test but I believe that will be taken care of by the coordinator responding to a View change.

uportal-war/src/main/java/org/jasig/portal/jgroups/protocols/DAO_PING.java
```diff
@@ -101,17 +108,18 @@ public void destroy() {
...
-        //Get the current cluster members
-        final Map<Address, PhysicalAddress> existing_mbrs = pingDao.getAddresses(clusterName);
 
         //Add our address to the store
         getAndSavePhysicalAddress(clusterName);
-        
+
+        //Get the current cluster members
+        final Map<Address, PhysicalAddress> existing_mbrs = pingDao.getAddresses(clusterName);
+
```
jameswennmacher: I don't agree with the above change.  I think you want your list of existing members to be everyone BUT yourself.  You don't want to send messages to yourself.

uportal-war/src/main/java/org/jasig/portal/jgroups/protocols/DAO_PING.java
```diff
@@ -134,6 +142,7 @@ protected void getAndSavePhysicalAddress(String clusterName) {
     }
 
     protected void handleView(View view) {
+        log.warn("Change in JGroup membership: " + view.toString());
```
jameswennmacher: Should be INFO, or DEBUG.  You don't want to clutter the logs with these messages, though you hope they are fairly rare and only occur during software updates.

bjagg: Yes, they should be rare and only occur when an instance is joining or leaving. Thought WARN would be appropriate, especially if unexpected

uportal-war/src/main/java/org/jasig/portal/jgroups/protocols/DAO_PING.java
bjagg: The DAO implementation retrieves all addresses from the table, including the local address if it is present. Also, the method getAndSavePhysicalAddress() is called on a timer. The original order may play a part on the very first call to fetchClustermembers(), which should then be documented.

bjagg: I will revert this change and squash changes into a single commit to streamline cherrypicking

uportal-war/src/main/java/org/jasig/portal/jgroups/protocols/DAO_PING.java
```diff
@@ -142,8 +151,12 @@ protected void handleView(View view) {
         
         final Collection<Address> mbrs = view.getMembers();
         final boolean is_coordinator = !mbrs.isEmpty() && mbrs.iterator().next().equals(local_addr);
-        if (is_coordinator) {
+        if (is_coordinator != is_coord) {
+            log.warn("Inconsistent state: is_cood != calculated is_coordinator");
+        }
+        if (is_coordinator || is_coord) {
```
jameswennmacher: Per jGroups standards the coordinator is the first node in the view.  See http://www.jgroups.org/manual/html/user-channel.html#JoiningACluster.

I don't think is_coordinator will ever be != is_coord.  is_coord is computed by parent class Discovery.java as
{code}
            case Event.VIEW_CHANGE:
                List<Address> tmp;
                view=(View)evt.getArg();
                if((tmp=view.getMembers()) != null) {
                    synchronized(members) {
                        members.clear();
                        members.addAll(tmp);
                    }
                }
                current_coord=!members.isEmpty()? members.get(0) : null;
                is_coord=current_coord != null && local_addr != null && current_coord.equals(local_addr);

                return down_prot.down(evt);
{code}
You see is_coord is calculated to be the first member in the view in the event being equal to the local_addr.  I believe calling down_prot.down(evt) eventually calls DAO_PING.down(event) which calls this method which does the identical computation.

I think the previous code was fine as far as it went.  I believe the issue we are running into at Texas A&M University (the UP_JGROUPS_PING table not getting cleared by the coordinator) is due to a combination of
* the coordinator crashing with an OutOfMemory error or other catastrophic event
* uPortal's implementation uses a database table to coordinate the members of the cluster rather than using Multicast messaging or a GossipRouter (see class documentation of http://jgroups.org/javadoc/org/jgroups/protocols/Discovery.html), the first of which and possibly the second also  would allow the remaining members of the cluster to realize the coordinator was dead.

I think we need to attack the core issue ... giving one or more nodes in the cluster the opportunity to realize the coordinator is dead and isn't updating the UP_JGROUPS_PING table.  

One approach to solve that issue is to add another column called 'last_updated' to UP_JGROUPS_PING table.  Each node in the cluster every DAO_PING.interval milliseconds posts its address to the table.  This should also update the timestamp stored in the last_updated column.  Then when a view change occurs if a node is not the coordinator have some mechanism where it checks the coodinator's entry in the UP_JGROUPS_PING table and if it is older than interval*2, continue through and act like the coordinator to update the table, possibly insuring the coordinator's entry is not in the view.  I'm not sure if this strategy will cause a loop or race condition where multiple nodes try to sync the table or if whomever is the first remaining node in the view doesn't know they are now the coordinator (would we need to trigger a view change event to notify everyone the coordinator is down and would it get broadcast to the nodes in the cluster?).  We should look at any recovery strategy carefully to insure it doesn't cause cycles or still have an issue.  In development setting a breakpoint in the JVM should freeze it enough to allow us to test the recovery mechanism reasonably easily.

jameswennmacher:
I suggest separating DAO_PING into a separate PR. It is a separate issue/behavior.

PR #685, opened 2016-06-28
UP-4700: Configure PAG caches with TTL of 180 and added JGroups liste…

bjagg:
…ner. Also Added TTL of 8 hours to resource caches.

https://issues.jasig.org/browse/UP-4700

### DIFF
--- a/uportal-war/src/main/resources/properties/ehcache.xml
+++ b/uportal-war/src/main/resources/properties/ehcache.xml
@@ -145,7 +145,7 @@
      +-->
     <cache name="org.jasig.portal.web.skin.Resources"
         eternal="false" maxElementsInMemory="50" overflowToDisk="false" diskPersistent="false"
-        timeToIdleSeconds="0" timeToLiveSeconds="0" memoryStoreEvictionPolicy="LRU" statistics="true" />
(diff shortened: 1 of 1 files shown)


uportal-war/src/main/resources/properties/ehcache.xml
```diff
@@ -1922,7 +1922,14 @@
      +-->
     <cache name="org.jasig.portal.groups.pags.dao.EntityPersonAttributesGroupStore.entityGroup"
```
jameswennmacher: On all of these caches that you add the cacheEventListenerFactory too, please remove from the comments 'Not replicated.' and replace with a separate line that says
'- replicated by invalidation'
like https://github.com/bjagg/uPortal/blob/1a0e8b78d17db9ee5268415cad5e29f7b7d05f1c/uportal-war/src/main/resources/properties/ehcache.xml#L1242

Also for those PAGS caches that don't have a comment, please add a comment like https://github.com/bjagg/uPortal/blob/1a0e8b78d17db9ee5268415cad5e29f7b7d05f1c/uportal-war/src/main/resources/properties/ehcache.xml#L408-L416

jameswennmacher:
+1

PR #686, opened 2016-06-29
UP-4722: Adding some pags

jgribonvald:
https://issues.jasig.org/browse/UP-4722

These pags are used in our context. For details see each commits.

Unit test are provided.

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/InjectAttributeRegexTester.java
+++ b/uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/InjectAttributeRegexTester.java
@@ -1,0 +1,100 @@
+/**
+ * Licensed to Apereo under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work
+ * for additional information regarding copyright ownership.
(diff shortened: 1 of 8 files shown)


jgribonvald:
For the last commit I can't provide a test case as I don't know you we can mock the applicationContext needed in class (their is a code blocker to avoid to load applicationContext in test mode) If you tell me how to do i can make it ;)

bjagg:
These additions look great! The only quibble is the import order. See https://wiki.jasig.org/display/UPM42/JetBrains+IntelliJ on correcting it.

jgribonvald:
Ok, so 
updated for import order 
adapted from IntegerTester to AbstractIntegerTester rebuild
and tested with unit test

tell me if something remaining

drewwills:
@jgribonvald,

Thanks for staying on top of this item.

> tell me if something remains

I'm going to take care of it (in just a minute), but please open a JIRA ticket for Pull Requests (in most cases) and provide a link in the summary.

jgribonvald:
Ho sorry ! I will do it so for others tomorrow.

drewwills:
Merged out-of-band.

PR #688, opened 2016-06-30
UP-4726 :  Configuration for proxyCAS url and for load-balancing portals

jgribonvald:
In this way all CAS basic configurations works !

Adding a real (local) domain name and real contextPath to be able to configure directly with an external CAS for a possible load-balanced portal context..
These properties are needed to work correctly in the proxy cas mode with load-balanced portal. Also it configure proxy-cas-test-portlet to be able to work in proxy mod.
(But see PR: https://github.com/Jasig/cas-proxy-test-portlet/pull/1 to remove class provided in the overlay when the new version will be released of the portlet and revert the pom.xml in the overlay)

Also added the cas context path property to be able to set it to empty as in some context the cas is on a specific domain without path.

For documentation :

```
# the public domain, on balanced portal url
environment.build.uportal.server= my.univ.edu
# A public or internal domaine name, used on server's sides only for proxy request (CAS and portal)
environment.build.real.uportal.server=my.portal1.edu
# the public portail context
environment.build.uportal.context=/uPortal
# the internal portail context used in proxy cas
environment.build.real.uportal.context=/uPortal

# no specific protocol is needed as it should be always the same.
```

### DIFF
--- a/filters/data-test.properties
+++ b/filters/data-test.properties
@@ -48,8 +48,10 @@ environment.build.hibernate.connection.validationQuery=select 1 from INFORMATION
 
 # uPortal server configuration properties
 environment.build.uportal.server=localhost:8080
+environment.build.real.uportal.server=localhost:8080
 environment.build.uportal.protocol=http
 environment.build.uportal.context=/uPortal
+environment.build.real.uportal.context=/uPortal
 environment.build.uportal.secureSessionCookie=false
 environment.build.uportal.email.fromAddress=uportal@localhost
 environment.build.uportal.email.host=localhost
--- a/uportal-portlets-overlay/cas-proxy-test-portlet/pom.xml
+++ b/uportal-portlets-overlay/cas-proxy-test-portlet/pom.xml
@@ -42,6 +42,20 @@
             <version>${cas-proxy-test-portlet.version}</version>
             <type>war</type>
         </dependency>
+        <dependency>
+            <groupId>org.jasig.portlet</groupId>
+            <artifactId>cas-proxy-test-portlet</artifactId>
+            <version>${cas-proxy-test-portlet.version}</version>
+            <type>pom</type>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>javax.portlet</groupId>
+            <artifactId>portlet-api</artifactId>
+            <version>2.0</version>
+            <scope>provided</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/uportal-portlets-overlay/cas-proxy-test-portlet/src/main/java/org/jasig/portlet/cas/test/mvc/ProxyCasController.java
+++ b/uportal-portlets-overlay/cas-proxy-test-portlet/src/main/java/org/jasig/portlet/cas/test/mvc/ProxyCasController.java
@@ -1,0 +1,110 @@
+/**
+ * Licensed to Apereo under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work
+ * for additional information regarding copyright ownership.
+ * Apereo licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License.  You may obtain a
+ * copy of the License at the following location:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.jasig.portlet.cas.test.mvc;
+
+import javax.portlet.PortletRequest;
+import javax.portlet.RenderRequest;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.jasig.cas.client.validation.Assertion;
+import org.jasig.cas.client.validation.TicketValidationException;
+import org.jasig.cas.client.validation.TicketValidator;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.portlet.ModelAndView;
+
+/**
+ *
+ *
+ * @author Jen Bourey, jbourey@unicon.net
+ * @version $Revision$
+ */
+@Controller
+@RequestMapping("VIEW")
+public final class ProxyCasController {
+
+	private Log log = LogFactory.getLog(ProxyCasController.class);
+
+	private TicketValidator validator;
+
+	/**
+	 * Set the ticket validator to use for proxy ticket validation.
+	 *
+	 * @param validator
+	 */
+	@Autowired(required = true)
+	public void setTicketValidator(TicketValidator validator) {
+		this.validator = validator;
+	}
+
+	private String serviceUrl = "http://localhost:8080/cas-proxy-test-portlet";
+
+	public void setServiceUrl(String serviceUrl) {
+		this.serviceUrl = serviceUrl;
+	}
+
+	private String proxyTicketKey = "casProxyTicket";
+
+
+	/**
+	 * Attempt to validate the proxy ticket supplied to the UserInfo map and
+	 * display the result in the main view of the portlet.  If the ticket is
+	 * not found or cannot be validated, a short debugging message will be
+	 * displayed in the portlet.
+	 *
+	 * @param request
+	 * @return
+	 */
+	@RequestMapping
+	public ModelAndView validateProxyCas(RenderRequest request) {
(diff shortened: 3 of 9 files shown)


cousquer:
:+1:  at UPMC, we use this configuration

jgribonvald:
Maybe some more conf should be done en embeded apps using proxy cas like cas, calendar, I didn't use them in proxy mode so I can't test them ...

jonathanmtran:
[UP-4703](https://issues.jasig.org/browse/UP-4703) was created for '/cas' being hardcoded in security.properties. Please close the issue when this gets merged.

jgribonvald:
If you prefer I can rewrite my PR to provide 2 commits, one for the CAS context Path and CAS real Path to works correctly in proxy-cas and a second one for working with cas-proxy-test-portlet.

jgribonvald:
So I rewrote the PR and provided several commits to "explain" the configuration and the use. The third commit is useless if you release a new version from master of the proxy-cas-portlet and that you upgrade the portlet version in uportal parent pom.

jgribonvald:
I know that cas-proxy was disabled in master, but it would be great to be able to add these properties that will serve a lot of deploiement in french countries.

drewwills:
@jgribonvald -- I just re-read this contribution.  I think I understand it quite a bit better now than I did earlier.

This PR appears to be a solution to the very annoying problems that occur when you set up Proxy CAS and/or CAS Clearpass with a load-balanced uPortal.  James, Benito, and I have also spent countless hours fighting these battles in the past.

For our part, we _have_ been able to get proxy CAS to work with the distributed ehcache setup.  It's not fun to do, and it's easy to mess it up, but we have been successful with it.  I believe the instructions are [here](https://wiki.jasig.org/display/UPM43/Caching+and+Re-playing+Credentials), in the section under _Step 3: If uPortal servers are clustered in any environment (certainly Prod), enable PGP Ticket cluster replication_.

Your approach works too -- and it could well be easier than distributed ehcache -- but I wonder if it's something we document rather than something we bake right into the standard build & deploy setup for everyone.  I'm a little concerned about adding `environment.build.real.uportal.server` to each of the filters files.  I'm concerned that it makes the configuration more complex & confusing (and benefits a slim minority of deployers).

I'm going to look through it and perhaps add some in-line comments as well.

jgribonvald:
@drewwills : thanks for your feadback, if this can argument in favor of my PR, in ESUP community it's used a lot, more in old versions we integrated a such feature in an ESUP distribution and all ESUP members requested this configuration as a need, it's more an easy to use when it's well documented. In my mind it's not a lot to do and more a in my mind, more as it was tested nobody have to check if it works, it's always a problem in checking if proxy CAS is working.
For us the distributed ehcache setup is not efficient, in my case i've seen problems due to network possibilites on an external web host, the multicast can't be used sometimes ;)

On an other way maybe the CAS client configured with memcache can be applied (memcache implementation is provided in the cas client lib), I didn't checked if it can be used on proxy-mod... but could be usefull for global logout from CAS...

filters/data-test.properties
```diff
@@ -48,8 +48,10 @@ environment.build.hibernate.connection.validationQuery=select 1 from INFORMATION
 
 # uPortal server configuration properties
 environment.build.uportal.server=localhost:8080
+environment.build.real.uportal.server=localhost:8080
 environment.build.uportal.protocol=http
 environment.build.uportal.context=/uPortal
+environment.build.real.uportal.context=/uPortal
 environment.build.uportal.secureSessionCookie=false
```
bjagg: Also, setting "real" to the original value would be a good default, as in
`environment.build.real.uportal.server=${environment.build.uportal.server}`

Same for context.

uportal-portlets-overlay/cas-proxy-test-portlet/pom.xml
```diff
@@ -42,6 +42,20 @@
...
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>javax.portlet</groupId>
+            <artifactId>portlet-api</artifactId>
+            <version>2.0</version>
+            <scope>provided</scope>
+        </dependency>
     </dependencies>
```
bjagg: Can this now be removed per the summary?

uportal-portlets-overlay/cas-proxy-test-portlet/src/main/java/org/jasig/portlet/cas/test/mvc/ProxyCasController.java
```diff
@@ -0,0 +1,110 @@
...
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.portlet.ModelAndView;
+
+/**
+ *
+ *
+ * @author Jen Bourey, jbourey@unicon.net
+ * @version $Revision$
+ */
```
bjagg: Add a useful description of this class and remove @author and @version

uportal-portlets-overlay/cas-proxy-test-portlet/src/main/java/org/jasig/portlet/cas/test/mvc/ProxyCasController.java
```diff
@@ -0,0 +1,110 @@
...
+@Controller
+@RequestMapping("VIEW")
+public final class ProxyCasController {
+
+	private Log log = LogFactory.getLog(ProxyCasController.class);
+
```
bjagg: This could be made `final`

uportal-portlets-overlay/cas-proxy-test-portlet/src/main/java/org/jasig/portlet/cas/test/mvc/ProxyCasController.java
```diff
@@ -0,0 +1,110 @@
...
+	 *
+	 * @param validator
+	 */
+	@Autowired(required = true)
+	public void setTicketValidator(TicketValidator validator) {
+		this.validator = validator;
+	}
+
+	private String serviceUrl = "http://localhost:8080/cas-proxy-test-portlet";
+
```
bjagg: This should be moved up to where the other members are declared.

uportal-portlets-overlay/cas-proxy-test-portlet/src/main/java/org/jasig/portlet/cas/test/mvc/ProxyCasController.java
```diff
@@ -0,0 +1,110 @@
...
+	public void setServiceUrl(String serviceUrl) {
+		this.serviceUrl = serviceUrl;
+	}
+
+	private String proxyTicketKey = "casProxyTicket";
+
```
bjagg: Since this does not have a setter, it should be `private static final`.
This should also be moved up.

bjagg:
I like this approach. It has a lot of potential in environments with load balancers. A question often comes up about testing individual hosts in a cluster. While this does not address this issue, it touches on the idea of accessing hosts directly

bjagg:
N.B. I do not think this would be hard to merge in post uPortal 5 modular breakup.

ChristianMurphy:
@jgribonvald I also like the approach! :+1: :1st_place_medal: 
Will able able move this PR forward so it can be merged?

filters/data-test.properties
drewwills: Can we reach the goals of this change without changing the filters files?

I'm concerned that the difference between these properties is not readily apparent by looking at them.  The filters files are something every adopter has to look at and understand... is it possible to support this feature without introducing additional, complexity to the configuration?

jgribonvald: @bjagg : yes setting "real" to the original value from property would be better I will take in note to change it.
And for the name in that case I would prefer to rename it as "cluster" or "instance" or "local", in my mind the "casproxy" isn't appropriated as i'm using the host also for other properties. 
@drewwills : I can't do on an other way, I need to specify all cluster host and it's the only place at now, expect if all cas configuration go on a spring bean, I tried to do in that way but had problems with uPortal auth services implementation. 
The best would be to make a big clean on uportal auth schema/chain and use maybe something like spring-security, like that users will be able to adapt it on a way that would be a "standard" and generaly it's reusable/already used on an other app.

jgribonvald:
In my mind before to merge this PR you need to make a new release of the portlet cas-proxy-test-portlet, change the portlet version in uPortal and after I will review my PR ;)

ChristianMurphy:
@jgribonvald What needs to change in the `cas-proxy-test-portlet`?

jgribonvald:
@ChristianMurphy you need to make a new version of the portlet that's all !
After in uPortal source upgrade the version of the portlet and you will be able to merge my 2 first commits (the third commit was only needed before my PR was merged on the cas-proxy-test-portlet and a new version is done and used in uportal portlet overlay), or if you prefer I can review this PR to remove the last commit. Let me know what you prefer.

ChristianMurphy:
@Jasig/uportal-developers is there anyone available to cut a release of `cas-proxy-test-portlet` to unblock this PR?

ChristianMurphy:
Hey @jgribonvald! :wave: 
Similar to https://github.com/Jasig/uPortal/pull/698#issuecomment-316594194
The use of `filters` doesn't mesh with uPortal 5.
I appreciate the time you've put into these changes. :+1: 
It would really great to see a new PR with a Gradle/Spring approach! :100:

PR #690, opened 2016-07-06
NOJIRA: Purpose a French translation

jgribonvald:
It's a purpose, and I can change some text if there is something not well translated.

A part of translation comes from ESUP translations, but not all, somes texts like "Principals" wasn't adapted and I prefer to doesn't translate "Principals" as it's diffciult to find something related to permissions in french. More in Unix world we are familiar with Principals.


cousquer:
:+1: @jgribonvald tu es sûr de devoir échapper les "!" et les ":" ?  /! et /:
si c'est lié à ta pull [https://github.com/Jasig/uPortal/pull/679](url), indique le. ;-)

jgribonvald:
@cousquer this is due to my tool as eclipse RB doesn't work anymore, ResourceBundle in Idea doesn't translate to unicode, ... so I used a tool in geany (ide) and it escaped these char, it's usefull in some langages ;)

So I rewrote my PR to remove these char, but there were no incidence. But escaping some chars could be a good practice even if it's useless for the main's cases.

cousquer:
:+1:

jhelmer-unicon:
:+1:   This looks like a good and reasonable change.  Since I don't speak French, I need to rely on French speakers here to confirm the translations are good, but I don't see any technical reasons this update should not be merged.   Baring downvotes or objection will try to merge in the next 24 hours.

PR #691, opened 2016-07-06
NOJIRA: fix hard coded uPortal contextPath

jgribonvald:
Make some change on jsp and respondr.xsl to remove hard coded uPortal ContextPath using normalized way.


jgribonvald:
It would be great to have a way to check (maybe with Travis or an other thing), that you don't have hard coded path in uPortal project. It should disabled for datas, comments and config files ;)

cousquer:
+1

jhelmer-unicon:
:+1:    This seems like a good and reasonably safe change.  Baring downvotes will merge in the next day or so.

PR #692, opened 2016-07-07
UP-4702:  Add support for encrypted values to the property files with…

drewwills:
…in primaryPropertyPlaceholderConfigurer

https://issues.jasig.org/browse/UP-4702

This solution leverages [Jasypt](http://www.jasypt.org/).

The feature is optional and non-invasive.  Use it if you want to;  if you don't, carry on the way you were.

### Setting the Encryption Key

This solution expects the key as an environment variable named `UP_JASYPT_KEY`.  Use a command like...

`export UP_JASYPT_KEY=CHANGEME`

To set the key in the shell before starting the portal.  (Or set the variable another way.)

### Encrypting a Property Value

Use the [Jasypt Command Line Interface (CLI) Tools](http://www.jasypt.org/cli.html) to encrypt a string...

`./encrypt.sh input=sa password=CHANGEME`

Then copy & paste the output into one of the supported properties files (any of them within `primaryPropertyPlaceholderConfigurer`) _and place ENC() around it_.

#### Example

`environment.build.hibernate.connection.username=ENC(6EX9JW+LZDR2ErQgRh0pEA==)`

### ToDo:

- [ ] Update the Manual

### DIFF
--- a/uportal-war/src/main/java/org/springframework/context/support/PortalPropertySourcesPlaceholderConfigurer.java
+++ b/uportal-war/src/main/java/org/springframework/context/support/PortalPropertySourcesPlaceholderConfigurer.java
@@ -48,7 +48,7 @@ import org.springframework.util.CollectionUtils;
  *   <li>Force configuration setting ignoreResourceNotFound=true and (safely)
  *   ignore noisy WARNings in the log concerning missing properties files that
  *   are optional</li>
- *   <li>...</li>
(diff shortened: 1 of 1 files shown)


timlevett:
:+1: So does this decrypt at run time or at build time?

drewwills:
@timlevett -- At runtime.  The encrypted values are still encrypted sitting in tomcat/webapps/uPortal/.

timlevett:
Very neat

PR #695, opened 2016-07-12
UP-4706:  Possible memory leak in DLM causing hundreds of millions of…

drewwills:
… NodiInfo objects

https://issues.jasig.org/browse/UP-4706

There is some evidence in the community of a bug within the <code>mergePLFintoILF</code> process that can occur with certain data.  When triggered, it pulls the process into an infinite(?) loop, causing it to instantiate {@link NodeInfo} objects until the JVM runs out of memory.

I am currently unable to see how this issue could occur.

I can, however, offer a patch that tracks the number these objects created and fails noisily -- with detailed info -- when the specific threshold is crossed.  This patch should...

* Prevent the bug from bringing down servers;  and
* Provide better info (from the logs) that will allow us to fix the issue properly

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/layout/dlm/PLFIntegrator.java
+++ b/uportal-war/src/main/java/org/jasig/portal/layout/dlm/PLFIntegrator.java
@@ -16,14 +16,12 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.jasig.portal.layout.dlm;
 
-import java.util.List;
+package org.jasig.portal.layout.dlm;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.jasig.portal.PortalException;
-import org.jasig.portal.xml.XmlUtilitiesImpl;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.w3c.dom.NodeList;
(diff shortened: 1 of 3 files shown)


uportal-war/src/main/java/org/jasig/portal/layout/dlm/PLFIntegrator.java
```diff
@@ -277,4 +273,41 @@ static Element appendChild( Element plfChild,
...
+      * Helps {@link PLFIntegrator} track the number of {@link NodeInfo} objects
+      * created for a single layout.  There is some evidence in the community
+      * that there is a bug within the <code>mergePLFintoILF</code> process that
+      * can occur with certain data.  When triggered, it pulls the process into
+      * an infinite(?) loop, causing it to instantiate {@link NodeInfo} objects
+      * until the JVM runs out of memory.  This class tracks the number these
+      * objects created and fails noisily when the specific threshold is
+      * crossed.
+      */
+     /* package-private */ static final class NodeInfoTracker {
```
apetro: I'd much rather see a new `NodeInfoTracker.java` file than this inner class.

jhelmer-unicon:
:+1:

uportal-war/src/main/java/org/jasig/portal/layout/dlm/PLFIntegrator.java
drewwills: In this case I think your view is prudent.  NodeInfoTracker objects are referenced by and manipulated by 2 classes in the package.  I often like static inner classes for small details that are known only to the class in which they reside.

Done.

PR #697, opened 2016-07-20
UP-4710: Bug on GuestUserTester

jgribonvald:
see on JIRA for more details : https://issues.jasig.org/browse/UP-4710


jgribonvald:
use the test without the class update and you will see the problem

jgribonvald:
Now there are problems on guest evaluation. There is something but it would be great to know what, do you know why the pags was done on this way ?

drewwills:
@jgribonvald -- For starters... thank you so much!

This is excellent work.  Accurate diagnosis of the issue, correct solution, JIRA, test cases, and a pull request... this contribution is essentially perfect.  You're exactly the sort of person we want involved.  Please, more of these.

jgribonvald:
@drewwills : :blush:
With an easy correction it was the least that I could do

PR #699, opened 2016-08-02
Up 4694 ldap encoding

jhelmer-unicon:
Update to @mgillian's PR (https://github.com/Jasig/uPortal/pull/678).   I'll close his PR.   The only real difference is using the spring LdapEncoder vs. rolling our own.   

There were some suggestions to try and move to LdapQueryBuilder (which ultimately, also just uses LdapEncoder).  That change is much more significant because it requires updating the spring-ldap lib and rewriting large chunks of code.  For now, I want to focus on the core issue.

https://issues.jasig.org/browse/UP-4694


jhelmer-unicon:
Last call for reviews on this.  Barring objections, will probably merge on Thurs.

PR #700, opened 2016-08-03
allow security.properties to use the environment.build.cas.context variable

markamace1:
Update security.properties to use the environment.build.cas.context property that is set in the filters file.  This allows for adjusting for implementations of CAS that might not be deployed in the standard context configuration without having to further modify the security.properties file.

### DIFF
--- a/uportal-war/src/main/resources/properties/security.properties
+++ b/uportal-war/src/main/resources/properties/security.properties
@@ -48,17 +48,17 @@ credentialToken.root.cas=ticket
 ## (See comments in the LogoutController class)
 ## It would be better to escape the value of the url parameter, but since there are no parameters on the
 ## unescaped URL and since there are no further parameters on the logout URL, this does work.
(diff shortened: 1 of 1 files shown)


jonathanmtran:
FYI @jgribonvald addressed this pull request #688

markamace1:
Point taken.  I did not see that pull request.  It appears as if that pull request addresses that issue along with some other items.

drewwills:
You guys are very helpful... I ran into this item myself the other day.

markamace1:
It only became an issue when we implemented CAS in the ROOT context.  When I looked at it I couldn't figure out why we had the value/placeholder in the filter files, but did not actually use it.  

This was several months ago and I just now decided to fork the repo and make the pull request.  Since this *is* addressed in another pull request I can close this one if it helps keep things clean.

jhelmer-unicon:
I may be in the minority here, but I'd personally rather see this get merged independently and then get the other PR updated.  I'll wait for community feedback before doing anything though.

drewwills:
I think @jhelmer-unicon makes a good point.  This change is succinct and distinct;  plus, it allow us to apply a commit from the excellent Mark Mace :).

drewwills:
Merged out-of-band due to conflicts.

PR #701, opened 2016-08-03
[NOJIRA] fix duplicate portlet name

jhelmer-unicon:
…inistration.   I think the second one was a copy-paste error.  This avoids having 2 drop down items in the Add Portlet flow that have the exact same name, but do different things.


drewwills:
Awesome @jhelmer-unicon -- thanks!

PR #702, opened 2016-08-03
Add in various portlet preferences for widget configuration to layoutDoc

timlevett:
This will mostly benefit [angularjs-portal's widgets mode](http://uw-madison-doit.github.io/angularjs-portal/latest/#/md/expanded), but others could use this as well.

+ Adds some attributes to the `/uPortal/api/layoutDoc` API for widget configuration
+ refactors the efficiency flag check to a generic function.


vertein:
:+1:

PR #704, opened 2016-08-09
UP-4713:  Subclasses of o.j.p.g.pags.testers.IntegerTester don't hand…

drewwills:
…le non-String values

https://issues.jasig.org/browse/UP-4713


uportal-war/src/main/java/org/jasig/portal/groups/pags/testers/IntegerTester.java
```diff
@@ -52,31 +56,51 @@ public IntegerTester(String attribute, String test) {
...
+
+        if (atts != null) {
+            for (int i=0; i<atts.length && result == false; i++) {
+                Object objValue = atts[i];
+                try {
+                    Integer intValue = null;
+
+                    if (objValue == null) {
+                        continue;
+                    }
```
bjagg: This null check could be moved before the try block

PR #706, opened 2016-08-10
UP-4714: parametize db driver

bjagg:
https://issues.jasig.org/browse/UP-4714


bjagg:
Closing PR since this is specific to Maven which is going to be replaced with Gradle very soon.

PR #709, opened 2016-08-26
UP-4716: Fixed delegate portlet config save/cancel parsing issue with…

groybal:
… RespondrJS.

https://issues.jasig.org/browse/UP-4716

This work was done for CCC Student Services Portal, which uses RespondrJS.

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/portlet/registry/PortletEntityIdStringUtils.java
+++ b/uportal-war/src/main/java/org/jasig/portal/portlet/registry/PortletEntityIdStringUtils.java
@@ -1,0 +1,105 @@
+/**
+ * Licensed to Apereo under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work
+ * for additional information regarding copyright ownership.
+ * Apereo licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License.  You may obtain a
+ * copy of the License at the following location:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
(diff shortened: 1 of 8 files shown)


uportal-war/src/main/java/org/jasig/portal/portlet/registry/PortletEntityIdStringUtils.java
```diff
@@ -0,0 +1,105 @@
...
+ * UI using 'exclusive' window state URLs.
+ * 
+ * @see flows/edit-portlet/configMode.jsp
+ * @see org.jasig.portal.portlet.delegation.*
+ * @see https://issues.jasig.org/browse/UP-2563
+ * @see PortletWindowIdStringUtils
+ * @see PortletWindowRegistryImpl#STATELESS_PORTLET_WINDOW_ID
+ * 
+ * @author Gary Roybal
+ * @version $Revision$
```
drewwills: I have been removing `@version $Revision$` from files that I touch because I have read that this metadata makes no sense in a git repo (where commit ids are not sequential).  Instead, I believe `@since` adds good value.

uportal-war/src/main/java/org/jasig/portal/portlet/registry/PortletEntityIdStringUtils.java
```diff
@@ -0,0 +1,105 @@
...
+ */
+public class PortletEntityIdStringUtils {
+
+    static final char ID_PART_SEPARATOR = '_';
+    static final Pattern ID_PART_SEPARATOR_PATTERN = Pattern.compile(Pattern.quote(String.valueOf(ID_PART_SEPARATOR)));
+
+    static final char DELEGATE_LAYOUT_NODE_ID_SEPARATOR = '-';
+    static final String DELEGATE_LAYOUT_NODE_ID_PREFIX = "dlg" + DELEGATE_LAYOUT_NODE_ID_SEPARATOR;
```
drewwills: I'm thinking these 4 (and the 2 like them below) should probably be `private` or (maybe) `public`, if we actually intend other classes to use them.

drewwills:
This PR looks like excellent work -- the level of diagnosis, solution, and thoroughness are all very impressive.

drewwills:
I made a few contributions...

* Implemented some suggestions on the PR
* Removed several '@version $' from the package
* Used UCDetector to identify and remove a small amount of unused code

Then merged out-of-band.

PR #714, opened 2016-09-06
UP-4613: Fixed issue where exiting Fragment Admin logs out user when …

groybal:
…using a two step login process, where the first step sets up the SecurityContext Authentication and the second step is the normal /uPortal/Login process.

There are really three sets of changes here:

1.   refactor identity swap code in "PortalPreAuthenticatedProcessingFilter" to use an inner class ("IdentitySwapHelper"), in order to make the code easier to understand and to clean up the main authentication logic

2.  addition of "clearSecurityContextPriorToPortalAuthentication" flag in order to disable the clearing of the security context prior to authentication

3.  updates so that during an identity swap the current Authentication object gets save and later restored during the unswap

Numbers 2. and 3. are changes that may or may not need to be included, based on whether the problem they solve is thought to be general enough.  What they do is enable the use of a two step log in process where the first step establishes the SecurityContext Authentication, and the second step is the normal uPortal login logic. In the scenario for which these were added, we are using Shibboleth.  For login, the IDP redirects to uPortal at "/uPortal/saml/SSO" which is maps to  "org.springframework.security.saml.SAMLProcessingFilter".  On success, the redirect handler redirects to "/uPortal/Login" which uses the updated code in this pull request.  Because it is a two step login process, for identity swaps we need to save the Authentication object and restore it during the unswap.  The two step login also requires that we do not clear the security context in step 2 since it has been set up in step 1.

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/security/IdentitySwapperManager.java
+++ b/uportal-war/src/main/java/org/jasig/portal/security/IdentitySwapperManager.java
@@ -22,6 +22,8 @@ import javax.portlet.PortletRequest;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpSession;
 
+import org.springframework.security.core.Authentication;
+
 /**
  * Manages workflow around use of the identity swapper features.
  * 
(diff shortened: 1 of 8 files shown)


drewwills:
### General Note on New Pull Requests

Please hold off on merging pull requests to `master` for a brief period.  A significant refactor is already underway for uPortal 5, and the potential for difficult conflicts is large.  I'll do my best to make this change happen quickly.  Thank you for your patience.

jhelmer-unicon:
An updated version of the PR can be found at:  https://github.com/Jasig/uPortal/pull/736

Closing this PR.

PR #715, opened 2016-09-06
UP-4179:  uPortal 5 project overhault with a Gradle-based build

drewwills:
https://issues.jasig.org/browse/UP-4179

This is only the beginning of this work.

Please see the wiki page for this effort:  https://wiki.jasig.org/display/UPC/uPortal+5+Project+Overhaul

The complete plan is outlined there.

### Tasks

- [x] Move all the 'export-*' Import/Export scripts from `/org/jasig/portal/` (classapth) to `/org/apereo/portal/`
- [x] Review changes to import/export to verify that _appropriate_ pre-uP5 entity files are still importable
- [x] Leave the `script` attributes of existing files in `src/main/data` unchanged
- [x] Verify that no _unwanted_ files remain under `org.jasig.portal` within...
* ~~uportal-war/src/main/java~~
* uportal-war/src/main/resources
* uportal-war/src/test/java
* uportal-war/src/test/groovy
* uportal-war/src/test/resources
- [x] Test CONFIG mode
- [x] Troubleshoot the `org.apereo.portal.portlets.image.ImagePortletController`
- [x] grep for `org.jasig.portal` and verify that remaining occurrences make sense (some will, particularly a few within Import/Export)
- [x] Look into @holdorph's observations


drewwills:
Travis CI seems to be invoking `$gradle assemble` where it should be invoking `$gradle install`... anyone know why that is?

holdorph:
I think you'll need to undo the .gradle change to .gitignore, and remove the .gradle files then once the .gradle files are gone, put the .gradle change to .gitignore back.  Otherwise everytime someone builds they will see those files as "out of date" and in need of being committed.

drewwills:
@holdorph is that happening to you?  (It's not happening to me.)  I don't know a lot about how the Gradle Wrapper works, but I researched the topic of what entries are typically added to `.gitignore` for a Gradle-based project.  I didn't think to research whether to handle `.gitignore` or the wrapper in a specific order.

holdorph:
My comment is that right now the commits to not match the gitgnore.  you are ignoring files you have *IN* the repository, that's inconsistent.  One of the two things is wrong.  Either the files shouldn't be in the PR, or the line shouldn't be in the .gitinore. But putting files into the PR *THEN* updating the .gitignore to ignore those files, looks like an accident and mistake.  And it can have negative side effects in addition to just being plain confusing.

drewwills:
@holdorph -- I ran some tests.  Here's what I now believe...

I _thought_ that the command `$gradle wrapper` would create only files that are a necessary part of the build system, but I was wrong about that.  It also generated 6 files in `.gradle`.  Those files don't need to be included -- they're work/temp files used by Gradle during the execution of `$gradle wrapper` itself (and later used by other commands).  Nothing in `.gradle/` needs to be in the repo, and `.gradle/` can & should be listed in the `.gitignore`.

I have made these changes.

uportal-war/src/main/data/required_entities/entity-type/org.apereo.portal.security.IPermissionSet.entity-type.xml
```diff
@@ -20,6 +20,6 @@
 
 -->
 <entity-type script="classpath://org/jasig/portal/io/import-entity-type_v3-2.crn">
-  <name>org.apereo.portal.security.IPermissionSet</name>
+  <name>org.jasig.portal.security.IPermissionSet</name>
```
ChristianMurphy: wait, shouldn't these be changing from `jasig` to `apereo`, not vice verse?

uportal-war/src/main/data/required_entities/entity-type/org.apereo.portal.security.IPerson.entity-type.xml
```diff
@@ -20,6 +20,6 @@
 
 -->
 <entity-type script="classpath://org/jasig/portal/io/import-entity-type_v3-2.crn">
-  <name>org.apereo.portal.security.IPerson</name>
+  <name>org.jasig.portal.security.IPerson</name>
```
ChristianMurphy: same here, shouldn't these be changing from `jasig` to `apereo`, not vice verse?

build.xml
```diff
@@ -728,6 +728,20 @@
     <!--                                                                                    -->
 
     <!-- Internal use -->
+    <target name="gradle-install" description="Build and install (in maven local) the uPortal modules that have been ported to a Gradle-based build">
+        <!--
+         | TEMPORARY MEASURE!
```
ChristianMurphy: This was a temporary measure for before move to gradle, can this be removed now?

uPortal-soffit-renderer/src/main/java/org/apereo/portal/soffit/renderer/SoffitRendererController.java
```diff
@@ -0,0 +1,294 @@
+package org.apereo.portal.soffit.renderer;
```
ChristianMurphy: File is missing ASF License header

build.xml
drewwills: Not yet.  Probably we will need that target as long as we need the `build.xml` file.  We'll be able to remove that file entirely after Phase II (though maybe not immediately).

uPortal-soffit-renderer/src/main/java/org/apereo/portal/soffit/renderer/SoffitRendererController.java
drewwills: I've just run `$mvn license:format` on the whole project.  It picked up that, plus a hand full of other items.  Soon to be committed.

uportal-war/src/main/data/required_entities/entity-type/org.apereo.portal.security.IPermissionSet.entity-type.xml
drewwills: Yep -- that was a mistake that crept in when I was trying to erase _unwanted_ changes the IDE had made auto-magically to `uportal-war/src/main/data`.  Should be fixed already.

PR #719, opened 2016-09-26
UP-4733:  For non-production deployments, configure setting logAbando…

drewwills:
…ned=true in the Tomcat dbcp

https://issues.jasig.org/browse/UP-4733

If abandoned connections are happening in uPortal, I think we want visibility into that.


drewwills:
I like it @timlevett!

PR #720, opened 2016-10-03
UP-4743: Made cookie HTTP Only

bpowell:


uportal-war/src/main/java/org/apereo/portal/portlet/container/services/PortletCookieServiceImpl.java
```diff
@@ -332,6 +332,7 @@ protected Cookie convertToCookie(IPortalCookie portalCookie, boolean secure) {
     	cookie.setComment(this.comment);
     	cookie.setMaxAge(this.maxAge);
     	cookie.setSecure(secure);
+        cookie.setHttpOnly(true);
```
ChristianMurphy: could this be adjusted to match the indentation of the surrounding code?

drewwills: It should match the indentation.  Our current standard (iirc) is 4 spaces, no tabs.

drewwills:
Okay -- I took a closer look at this PR.  I have some findings.

1. This PR needs a JIRA ticket.  Please find or create one, edit the title, and provide a link in the description field
1. Please someone also fix the whitespace
1. Excellent find @ChristianMurphy;  I'm very happy you thought to look;  great work;  but there are a couple issues...
  * The Java change is a change to the PortletCookie, which is not the same kind of cookie that the JS is reading
  * The JS is trying to manipulate a kind of sidebar we no longer have;  nowhere does responder put out a `#portalSidebar` selector
1. We need to remove that JS file, and remove the references to that file (probably in the skin manifest)

blamonet:
@drewwills issue created: https://issues.jasig.org/browse/UP-4743

bpowell:
@drewwills I think I got everything you asked for. Let me know if I missed something.

PR #721, opened 2016-10-04
UP-4738:  Re-tune caches settings for Groups+Permissions in ehcache.xml

drewwills:
https://issues.jasig.org/browse/UP-4738

These are proposed changes to default cache sizes.  Please be liberal with your comments and help us choose good defaults.


vertein:
:+1:

drewwills:
Folks -- I'm going to back the 2nd commit out (and then merge).  The 2nd is more controversial and worth it's own discussion.  Honestly I don't know how I feel about it... a better choice might be simply to reduce the use of jGroups (I think we're using it places we probably shouldn't).

PR #717, opened 2016-09-07
UP-4729 : Activating flyout menu option from portal.properties

jgribonvald:
Usefull to avoid to modify the option in the xsl.

PR #681, opened 2016-06-24
UP- 4697 Enhancement (epic) of the Off canvas

cousquer:
[https://issues.jasig.org/browse/UP-4697](url)
to test don't forget to switch : \<xsl:param name="USE_FLYOUT_MENUS" select="'true'" /> in navigation.xsl

UP-4697 is trying to fix several existing Jiras to give me a branch name...
UP-4622 Add offcanvas stickynav flyout : there is a bug with the optional fly-out menu on iPhone
UP-4670 An horizontal scrollbar appears on narrow display (new mobile nav)
UP-4671 the new mobile nav doesn't have the right background-color (improvement: new LESS variable to handle colors and position in variable.less)
UP-4707 Desktop Nav is missing in IE 11 and previous version
UP-4708 IE 11 and previous browsers: The space created for the "responsive left nav" does not disappear when desktop users change window size if the menu is open
No-Jira Implement a new animated pure CSS Hambuger menu on mobile
No-Jira Use GPU accelerated CSS properties to animate the offcanvas
No-Jira Add support for the offcanvas to ie9, ie10, ie 11
No-Jira Improve uPortal Accessibility, a blind people can navigate, open/close main Navigation, open/close optional subNav (aka optional fly-out menus) and reach directly one portlet in maximize view located into Xth tab of the main Navigation with Voiceover (see https://t.co/1UdCIlfLKY). Improvement on keyboard navigation (tab/enter/space)
No-jira Enhancement of the Touch Area of the sub nav toggle, tabs gripper and tabs delete buttons for Touch devices
No-jira The bootstrap's Font (glyphicons) wasn't update with the last uPortal Bootstrap update : some Bootstrap css classes don't work correctly because glyphicons font wasn't bump to the right version + add font in Wolff2 + security Fix on the fonts.
partly (99%) resolve UP-4695 for breackpoints :765px, 766px 767px, breackpoint 769px, 770px, .... Still a display problem at 768px tue to two differents breackpoints in Bootstrap respondr. but it's usable in 768px now.
Tested on Windows 1O, OSX, Linux on Chrome / FF / Edge /Safari / IE11 / IE10 / IE 9
IOS latest (Iphone 6S / 5 /4) [chrome /safari] / IOS Old (old IPAD 1 not updated) (old chrome / old Safari) / Android
with and without the fly-out enable.

PR #722, opened 2016-10-06
UP-4729 : Activating flyout menu option from portal.properties - rel-4-3-patches

ChristianMurphy:
Useful to avoid to modify the option in the XSL.

Adapted from #717 

@jgribonvald 's patch rebased on top of current rel-4-3-patches.


jgribonvald:
Just a warning : don't merge this commit on branch rel-4-3-patches until PR #681 isn't merged on !

see my JIRA : https://issues.jasig.org/browse/UP-4741 that correspond to a such request

ChristianMurphy:
@jgribonvald thanks for bringing my attention to that.
That commit contains changes to the XSL transforms, while I'd don't think that should cause an issue.
It would be safer for the change to be merged into release 4.4

PR #723, opened 2016-10-06
UP-4729 : Activating flyout menu option from portal.properties - master

ChristianMurphy:
Useful to avoid to modify the option in the XSL.

Adapted from #717 

@jgribonvald 's patch rebased on top of current master.


ChristianMurphy:
resolved out of band in #717

PR #724, opened 2016-10-10
UP-4740:  Report on caches in poor condition

drewwills:
https://issues.jasig.org/browse/UP-4740

This PR writes warnings to the logs when a cache that should be big enough (for all objects of the corresponding type) starts evicting cache elements due to the configured maxElementsInMemory.

Examples of such caches are those that hold:
- Groups
- Portlets
- Fragments

NOTE:  I'm not sure if a solution like this is a good idea, or if we're better off simply lifting the maxElementsInMemory... allowing the caches to grow to whatever size is needed.

### DIFF
--- a/uportal-war/src/main/java/org/apereo/portal/utils/cache/CacheHealthReporterService.java
+++ b/uportal-war/src/main/java/org/apereo/portal/utils/cache/CacheHealthReporterService.java
@@ -1,0 +1,128 @@
+/**
+ * Licensed to Apereo under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work
+ * for additional information regarding copyright ownership.
+ * Apereo licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License.  You may obtain a
+ * copy of the License at the following location:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apereo.portal.utils.cache;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import net.sf.ehcache.Ehcache;
+import net.sf.ehcache.Element;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.scheduling.annotation.Async;
+import org.springframework.stereotype.Service;
+
+/**
+ * This service supports custom {@link net.sf.ehcache.event.CacheEventListener}
+ * objects that recognize when "something bad" is happening with a uPortal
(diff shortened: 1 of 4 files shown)


drewwills:
Thanks for your thoughts @ChristianMurphy -- I hope I can encourage others to chime in as well.  This is one where I'd like to hear a handful of perspectives.

uportal-war/src/main/java/org/apereo/portal/utils/cache/CacheHealthReporterService.java
```diff
@@ -0,0 +1,128 @@
...
+                    }
+                    report.append("The following cache(s) have insufficient maxElementsInMemory;  " +
+                            "there must be room in the cache to hold every object of " +
+                            "the corresponding type in the portal:");
+                    for (Map.Entry<String,Integer> y : counts.entrySet()) {
+                        report.append("\n\t- ").append(y.getKey()).append(" (")
+                                .append(y.getValue())
+                                .append(" elements evicted due to insufficient size since last report)");
+                    }
+                    logger.warn(report.toString());
```
apetro: `.toString()` probably isn't necessary here.

uportal-war/src/main/java/org/apereo/portal/utils/cache/CacheHealthReporterService.java
```diff
@@ -0,0 +1,128 @@
...
+    public void reportCacheSizeNotLargeEnough(Ehcache cache, Element element) {
+        final ReportTuple tuple = new ReportTuple(cache, element);
+        Reports.CACHE_SIZE_NOT_LARGE_ENOUGH.add(tuple);
+    }
+
+    /*
+     * Nested Types
+     */
+
+    private static final class ReportTuple {
```
apetro: Style: This nested type would be better modeled as a full-fledged class. Non-`public` even, if you like.

andrewstuart:
Sounds like a good strategy to me. I'd love to see some metrics tracking as well at some point, perhaps using a system like [Prometheus](https://prometheus.io) that lets you track numbers without having to explicitly query logs.

Makes for a good dashboard and monitoring system, and you can set up alerts if, say, you have too many cache evictions in a given time span.

uportal-war/src/main/java/org/apereo/portal/utils/cache/CacheHealthReporterService.java
drewwills: I don't belong to the school of thought that holds there is no place for inner classes in Java.  I understand there are others -- smart folks -- who feel precisely that way... but I feel (strongly) otherwise.

I **do**, on the other hand, recognize & agree that inner classes can be (a) overused and/or (b) confusing if abused.

This inner class (`ReportTuple`) is 100% implementation detail of the surrounding class (`CacheHealthReporterService`).  It is not visible to other classes -- even in the same package -- nor is it referenced anywhere outside the class... even in configuration.  This is the kind of use case where inner classes can shine.  I would say an inner class here is an acceptable -- though not mandatory -- approach.

uportal-war/src/main/java/org/apereo/portal/utils/cache/CacheHealthReporterService.java
drewwills: I tried, but apparently there isn't a method signature on `org.slf4j.Logger` for that.

I'd have to do...

```
logger.warn("{}", report);
```

or...

```
logger.warn("" + report);
```

PR #725, opened 2016-10-10
UP-4506: Fix alt max link to be spelled correctly - rel-4-3_patches

jgribonvald:
Change alternativeMaximixedLink to be alternativeMaximizedLink.

This if for 4.3 branch.

### DIFF
--- a/uportal-war/src/main/resources/layout/structure/columns/columns.xsl
+++ b/uportal-war/src/main/resources/layout/structure/columns/columns.xsl
@@ -271,14 +271,14 @@
                             <xsl:sort select="number(@dlm:precedence)" order="descending"/>
                             <sidebarChannel name="{@name}" title="{@title}" ID="{@ID}" fname="{@fname}" description="{@description}">
                                 <xsl:choose>
-                                    <xsl:when test="parameter[@name='PORTLET.alternativeMaximixedLink']">
-                                        <xsl:attribute name="alternativeMaximixedLink">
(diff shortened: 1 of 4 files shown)


jgribonvald:
@timlevett : what do you mean by default entity files ?
The property AlternativeMaximizedLink is a parameter of portlets configuration, and there isn't any data that contains this param.

After I tested this change and the marketplace is working pretty well now with this property for portlets where it's defined. There is only the sitemap that need an update (see other PR for that)

drewwills:
@timlevett makes an interesting point.  I don't see any portlet publications in required/default/quickstart data (any data in Apereo master) that refer to "alternativeMaximixedLink" but wonder what would happen when this change encounters data with the misspelling.  Broken feature?  Stack trace, and an unusable portal?

jgribonvald:
@drewwills : with misspelling there isn't any error or broken feature ! It doesn't use it only, when I Introduced the parameter all test was done to doesn't break anything. The only thing that I didn't test is on muniversality theme as we don't use it !
And with this PR all is working well even with marketPlace (default activated in portal.properties)

If you want I can give you an access on our "under development" instance to watch how this is working.

jgribonvald:
do you need something to test the feature before merging it ?

drewwills:
Looks great -- sorry for the delay.

jgribonvald:
Don't worry  ; )

Thanks for your job Drew  !

PR #727, opened 2016-10-10
UP-4742: Use AlternativeMaximizedLink into sitemap - rel-4-3_patches

jgribonvald:
Set anchor url with Alternative Maximized Link on sitemap

Commit for branch rel-4-3_patches (see other PR for branch master)

### DIFF
--- a/uportal-war/src/main/resources/layout/theme/respondr/respondr.xsl
+++ b/uportal-war/src/main/resources/layout/theme/respondr/respondr.xsl
@@ -401,6 +401,11 @@
                                 <ul>
                                     <xsl:for-each select="tabChannel">
                                         <xsl:variable name="portletLinkUrl">
+                                            <xsl:choose>
+                                                <xsl:when test="@alternativeMaximizedLink and string-length(@alternativeMaximizedLink) > 0">
+                                                  <xsl:value-of select="@alternativeMaximizedLink" />
+                                                </xsl:when>
+                                                <xsl:otherwise>
                                             <xsl:call-template name="portalUrl">
                                                 <xsl:with-param name="url">
                                                     <url:portal-url>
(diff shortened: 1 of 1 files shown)


drewwills:
I'm 100% for the change, but is it a change we want to make in a patch release?

jgribonvald:
This can follow up-4506, and it's advised !

After the sitemap should be rewrited as portlet (a TODO in xsl tell that)

jgribonvald:
Could this be merged into 4.3 ? or you need something ?
Also there is the PR for the master ;)

bjagg:
PR for master branch was merged.

The issue is semantic versioning where a patch release should only contain "fixes" while a minor release would contain enhancements. See http://semver.org/

drewwills:
If the intent was already to use the Alternative Maximized Link in the sitemap -- and especially if there was a comment to that effect -- I'm cool with considering this a "bug fix" and merging.

jgribonvald:
Sorry Drew I'm back to my work after long deserved hollidays and a good break.

Else yes it's needed if you use this parameter on navigation links, more if you don't have it in the sitemap the user won't understand if he uses the sitemap as navigation, ie when the channel is opened in an other way than in navigation.

Thanks

drewwills:
A few additional observations...

- [] There appear to be some whitespace/indentation adjustments needed (element nesting levels changes, but existing elements weren't updated to reflect the changes)
- [] I think `<xsl:if>` is probably preferable over `<xsl:choose>` with an empty `<xsl:otherwise>`
- [] Anywhere we use `target="_blank"` we are also obligated to add `rel="noopener noreferrer"`

I'm going to update these items on this PR and merge.  I will also review the same things in `master`.

drewwills:
Suggestions applied to both `rel-4-3-patches` and `master`;  merged out-of-band.

PR #733, opened 2016-10-16
UP-4747: Setup Checkstyle

ChristianMurphy:
https://issues.jasig.org/browse/UP-4747

Configures checkstyle.
Code style is currently set to the Google's style.
This will likely require some [bikeshedding](https://en.wiktionary.org/wiki/bikeshedding) before it can be merged 🚲 

Reference for choice of Google code Style:
http://apetro.ghost.io/citizens-for-a-better-uportal-tomorrow-tomorrow/

cc @apetro


ChristianMurphy:
Checkstyle includes the [Google styleguide](https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml) as part of the plugin.
However it is not official documentation on how to access that configuration directly from the plugin.
There is an [unofficial solution](https://discuss.gradle.org/t/checkstyle-how-to-use-an-official-style-configuration/6952) on the gradle forum, but did not work in my tests.
Anyone have any wisdom on how to get that to work?
Or is it fine as is with the configuration included in the uPortal source?

cousquer:
@ChristianMurphy 
Setting 'transitive = false' on the dependency didn't resolved the issue ?

build.gradle
```diff
@@ -23,12 +23,20 @@ buildScan {
...
     repositories {
         mavenLocal()
         mavenCentral()
     }
 
+    checkstyle {
+        // Stay on checkstyle version 6 until Java version 7 support can be dropped
+        toolVersion = '6.19'
+        showViolations = true
+        configFile = file("${project.rootDir}/checkstyle.xml")
```
jhelmer-unicon: May wanna add suppression file support until all existing code can be converted.   See: https://gist.github.com/soucekv/a962f788ceb70e418e2f

In my current project, I basically used a script to add *every* file to the suppression script and then just have a standing rule that team members to try and remove files from the suppression list whenever they need to significantly modify a file.  That at least keeps the problem from getting worse as new files are added, but also encourages devs to help clean up the old stuff when they have the chance.

jhelmer-unicon:
I've always preferred to have a local copy anyhow of the rules anyhow.   It just makes it easier if you start to find some rule too onerous.   Also, ensures that updating the plugin doesn't all the sudden introduce a bunch of new rules that breaks your build.

build.gradle
ChristianMurphy: :+1: Great idea,
Originally I was thinking of moving rule by rule going from warning to error.
Going file by file could work as well, and it sound like using this to encourage community contributors to participate in the process could work out better.
As long as we don't try to do everything all at once it should work out. :smile:

bjagg:
Is this PR waiting for the suppression file?

ChristianMurphy:
@bjagg Added suppression file

PR #743, opened 2016-10-25
UP-4756: Ensure links have a valid href destination - rel-4-3-patches

ChristianMurphy:
#### Issue

https://issues.jasig.org/browse/UP-4756

Anchor element found with a valid href attribute, but no link content has been supplied.

``` html
<a href="javascript:;"> ...</a>
```
#### Resolution

Use empty anchor instead.

``` html
<a href="#"> ...</a>
```

### DIFF
--- a/uportal-war/src/main/resources/layout/theme/respondr/content.xsl
+++ b/uportal-war/src/main/resources/layout/theme/respondr/content.xsl
@@ -128,7 +128,7 @@
         <div id="column_{@ID}" class="portal-page-column {$POSITION_CSS_CLASS} {$WIDTH_CSS_CLASS} {$MOVABLE} {$DELETABLE} {$EDITABLE} {$CAN_ADD_CHILDREN} {$FRAGMENT_OWNER_CSS}"> <!-- Unique column_ID needed for drag and drop. -->
           <div id="inner-column_{@ID}" class="portal-page-column-inner"> <!-- Column inner div for additional presentation/formatting options.  -->
             <xsl:if test="$IS_FRAGMENT_ADMIN_MODE='true'">
-                <div class="column-permissions"><a class="button portal-column-permissions-link" href="javascript:;"><span class="icon permissions"></span><xsl:value-of select="upMsg:getMessage('edit.column.x.permissions', $USER_LANG, $NUMBER)"/></a></div>
+                <div class="column-permissions"><a class="button portal-column-permissions-link" href="#"><span class="icon permissions"></span><xsl:value-of select="upMsg:getMessage('edit.column.x.permissions', $USER_LANG, $NUMBER)"/></a></div>
             </xsl:if>
             <xsl:apply-templates select="channel|blocked-channel"/> <!-- Render the column's portlets.  -->
           </div>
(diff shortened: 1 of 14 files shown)


ChristianMurphy:
> the "javasript:void(0);" is a valid protocol

While the `javascript:` scheme is provisionally supported in the [whatwg url standard](https://url.spec.whatwg.org/#dom-url-hash).
Using the `javascript:` scheme violates the [semantics of the `<a>` element](https://html.spec.whatwg.org/multipage/semantics.html#the-a-element).

> If the a element has an href attribute, then it represents a hyperlink (a hypertext anchor) labeled by its contents.

Since scheme `javascript:` represents a command not a hyperlink.

The [more semantically correct](https://html.spec.whatwg.org/multipage/forms.html#the-button-element) and [more accessible](https://formidable.com/blog/2014/05/08/anchors-buttons-and-accessibility/) way would be to use the `<button>` element with [`class="btn btn-link"`](https://getbootstrap.com/css/#buttons-options).
However using the `<button>` element has the potential to break skins/themes that style the `<a>` element rather than modifying the less `@link-color` and `@link-hover-color` variables.

For uPortal 5 switching javascript commands to the `<button>` element would be an option, but uPortal 4.3 should stick with the `<a>` element.

> From the a11y POV, you've got downsides with href="#" : <http://stackoverflow.com/a/1292546>

Since the code is [already calling `preventDefault()`](https://github.com/Jasig/uPortal/search?utf8=%E2%9C%93&q=preventDefault), `href="#"` would have no scroll side effect.

------

> as for me: +1 for href="javascript:void(0);"

Okay
Thoughts on moving uPortal 5 to using the `<button>` element?

cousquer:
Didn't notice that it calls on every <a href="#"> preventDefault(); , so that's ok. 
That was my little fear.
+1 for your POV. 😀
@christianMurphy rocks! 👌

cousquer:
´´´ * <a href="#"> ´´´

cousquer:
* links

ChristianMurphy:
> Didn't notice that it calls on every `preventDefault()`

It is on *most*, I'll need to go though and confirm that it is called on *every* `<a href="#">` element event listener.

cousquer:
@ChristianMurphy 
Sorry for the late response : 

> Thoughts on moving uPortal 5 to using the `<button>` element?

You need to wonder two things (WCAG2 POV)
Does the hyperlink load a new page after clicking on it ? Or does all the content is still on the same page?

In the first case, links with href="#" is better, it is the primary function of link. It will be more robust, especially in case of JS loading delay. Furthermore it allows to navigate through the browser history, which could be very significant (thanks to Backspace).

In the second case , the buttons may agree, taking care to make them very explicit (with `title` or `aria-label` or `aria-labelledby`) because it's necessary to inform clearly the user about the action he will do. If there is in a long series of steps with buttons, it can quickly become difficult to know at which step the user stands.

You should know that `links` and `button` are different in a11y, because there are specific behaviors to 2 elements. For example: Space triggers a button but not a link. And if you touch / click on a button, keeping the pressure on it and then leave the touchable / clickable area, the button will not be triggered, while the link will be triggered (eg, on mobile browser, triggering the opening of a contextual menu).

hope this help. (If you think this can be usefull, do not hesitate to copy this into the jira for WCAG2 to keep track) 😸 
@cousquer

ChristianMurphy:
Thanks much @cousquer !

ChristianMurphy:
After a conversation with @drewwills `href="javascript:;"` allows support for older IE and should not be removed in **rel-4-3-patches** this would still be an option from uPortal 5.

PR #744, opened 2016-10-25
UP-4757: Label for inputs in permissions administration portlet - rel-4-3-patches

ChristianMurphy:
https://issues.jasig.org/browse/UP-4757
#### Issue

``` html
<input class="up-autocomplete-searchterm form-control" type="text" name="principalDisplayName" value="John" autocomplete="off">

<input class="up-autocomplete-searchterm form-control" type="text" name="activityDisplayName" value="Permission" autocomplete="off">
```
#### Resolution

Add title attribute with description.


jhelmer-unicon:
Generally, I think best practice is to *not* use the title attribute.  The title attribute is not well supported and  during the accessibility audit on my current project, we were told that it does not improve accessibility.

See:  https://www.paciellogroup.com/blog/2012/01/html5-accessibility-chops-title-attribute-use-and-abuse/ 

(paciellogroup does accessibility reviews)

The way we approached theses issues:
- If possible, for every input element, include a label tag with a valid "for" attribute.
- If you can't add a label, but you have some other descriptive text, use aria-labelledby="" to point to that text
- If no descriptive element exists on the page, then add a new span with descriptive text and hide it from view.  Then, add an aria-labelledby that points at the new span.

Even for the stuff we used to use title for (eg. more descriptive tool-tip info), we were told to switch to putting that content on the page and then adding a aria-describedby instead on the input.  That seems to be much better supported by the 3 main screenreaders we have tested with (JAWS, NVDA and Voiceover)

ChristianMurphy:
Thanks @jhelmer-unicon in this case `<label>` will not work well. I'll switch the `title` attributes over to `aria-labelledby` attributes.

PR #756, opened 2016-11-08
UP-4763: Added labels on the input elements in the directory portlet

alasak:


uportal-war/src/main/webapp/WEB-INF/jsp/Directory/directory.jsp
```diff
@@ -40,7 +40,7 @@
       <div class="portlet-section-body">
 
         <form action="${ formUrl }" method="POST">
-            <input name="query" value="${ fn:escapeXml(query )}"/> <input type="submit" value="<spring:message code='search.directory.submit'/>"/>
+            <label for="query"><input name="query" value="${ fn:escapeXml(query )}"/> <input type="submit" value="<spring:message code='search.directory.submit'/>"/></label>
```
ChristianMurphy: :art: Could this be broken out across several lines?

Something like:

``` jsp
<label for="query">
    Query
</label>
<input type="search" name="query" value="${ fn:escapeXml(query )} "/>
<input type="submit" value="<spring:message code='search.directory.submit'/>" />
```

Thanks!

bjagg:
Awaiting edits

uportal-war/src/main/webapp/WEB-INF/jsp/Directory/directory.jsp
alasak: Thanks for looking at the changes and providing your suggestions.

uportal-war/src/main/webapp/WEB-INF/jsp/Directory/directory.jsp
```diff
@@ -40,7 +40,11 @@
       <div class="portlet-section-body">
 
         <form action="${ formUrl }" method="POST">
-            <input name="query" value="${ fn:escapeXml(query )}"/> <input type="submit" value="<spring:message code='search.directory.submit'/>"/>
+            <label for="query">
+                 <spring:message code='search.directory'/>
+            </label>
+            <input name="query" value="${ fn:escapeXml(query )}"/>
```
ChristianMurphy: `<input>` tag needs and `id` attribute with value `query` for `<label>` tag to link correctly on screen readers.

``` html
<input id="query" name="query" value="${ fn:escapeXml(query )}"/>
```

ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#attr-for

drewwills:
@alasak et al. -- was this one already applied to `master`?

ChristianMurphy:
@drewwills this has not been applied to `master`

bjagg:
This was the last of the accessibility changes from last week.

drewwills:
@ChristianMurphy -- Roger.

IMHO PRs should pretty much always be opened against master... exclusively.  Right now we have an oddball situation where 4.x has different packaging than 5.x, so we're commonly opening a master version plus a 4.x version, and that's understandable and cool.

We don't want to get in a situation where we've applied a fix/improvement to a maintenance branch but not to master.  Incremental dev is on master;  some of that work (in accordance w/ Semantic Versioning) qualifies to be applied to one or more maintenance branches.

drewwills:
I took care of it.

PR #761, opened 2016-11-10
UP-4765: register portlet flow labels - master

ChristianMurphy:
https://issues.jasig.org/browse/UP-4765

#### Issue

This select element does not have a name available to an accessibility API. Valid names are: label element, title attribute.

#### Resolution

Add labels, update input types to more specific html 5 types where possible.

#### Note

This code depends on the `preferenceInput` changes made in #753


ChristianMurphy:
There looks to be some disconnect with the Travis build, the status is reported as fail, but the log reports

``` log
BUILD SUCCESSFUL

Done. Your build exited with 0.
```

PR #762, opened 2016-11-15
UP-4766: Fix 'SecurityException: Not Authorized' bug when adding a po…

groybal:
…rtlet as a Tenant Admin user.


groybal:
I forgot to create a branch specific to UP-4766.  I'll create a new updated pull request.

PR #768, opened 2016-11-17
UP-4769: Add labels to inputs in Translator portlet - rel-4-3-patches

ChristianMurphy:
https://issues.jasig.org/browse/UP-4769

#### Issue

*Code Snippet*
``` html
<select id="Pluto_39_ctf8_14_entityType"> <option value="portlet...</select>
<select id="Pluto_39_ctf8_14_locale"> <optio...</select>
<input type="text" name="title" class="uptrans-pfrm-loc-title">
<input type="text" name="name" class="uptrans-pfrm-loc-name">
<textarea name="description" class="uptrans-pfrm-loc-descr"></textarea>
```
... all inputs, selects, textareas.

This textarea element does not have a name available to an accessibility API. Valid names are: label element, title attribute.

*Code Snippet*
``` html
<textarea name="description" class="uptrans-pfrm-loc-descr"></textarea>
```

This text input element does not have a name available to an accessibility API. Valid names are: label element, title attribute.

*Code Snippet*
``` html
<input type="text" name="title" class="uptrans-pfrm-loc-title">
<input type="text" name="name" class="uptrans-pfrm-loc-name">
```
...all text inputs instances.

This fieldset element does not have a name available to an accessibility API. Valid names are: legend element.

*Code Snippet*
``` html
<fieldset> <div><label>Title:...</fieldset>
<fieldset> <div><label>Name:<...</fieldset>
```
... all fieldset instances.

Fieldset does not contain a legend element. All fieldsets should contain a legend element that describes a description of the field group.

*Code Snippet*
``` html
<fieldset> <div><label>Title:...</fieldset>
<fieldset> <div><label>Name:<...</fieldset>
```
... all fieldset instances.

Added boostrap form-control to inputs to smooth appearance

#### Resolution

Added labels to inputs, and legends to fieldsets.
Additionally added `class="form-control"` to inputs to smooth appearance of input fields.

screencapture-localhost-8080-uportal-p-translator-ctf3-max-render-up-1479406124447


cousquer:
@ChristianMurphy 
btw I've never see this view in uPortal admin . Where can I reach it in admin ?
I cc @auxepaul (my colleague at UPMC)

ChristianMurphy:
@cousquer @auxepaul 
The translator portlet is located at the relative url `/uPortal/p/translator.ctf4/max/render.uP`

cousquer:
after that maybe [aria-flowto](https://www.w3.org/TR/wai-aria/states_and_properties#aria-flowto) on the buttons to scroll up and [aria-live="polite"](https://www.w3.org/TR/wai-aria/states_and_properties#aria-live) on the updated form can be usefull.
do not pay attention in full force mode... 😸

ChristianMurphy:
> beware in french, there is an unsecable space before ":"

If that is unsecable should that be a `&nbsp;` instead of ` `?

cousquer:
> If that is unsecable should that be a &nbsp; instead of ?

\u00A0 I think [Unicode Character 'NO-BREAK SPACE' (U+00A0)](http://www.fileformat.info/info/unicode/char/00a0/index.htm)

I'm quoting myself : 

> do not use aria-flowto... it's a big mistake...

In fact, after a good night : you can use **[`aria-flowto`](https://www.w3.org/TR/wai-aria/states_and_properties#aria-flowto)** to change the reading order after clicking on a button in the portlet list to go up to reach the `id `of the updated form with a `tabindex="0"` on it.
[[aria-flowto with a link inside to the **keyboard layout source** just a reminder of a past discussion on the list :wink: ](http://3needs.org/en/testing/code/aria-flowto.html)]

ChristianMurphy:
I updated edit portlet actions to be buttons, and added listener to the enter key so editing can be triggered via keyboard events. :keyboard: 

-----

> In fact, after a good night : you can use aria-flowto to change the reading order after clicking on a button in the portlet list to go up to reach the idof the updated form with a tabindex="0" on it.

I added the `aria-flowto` on the `click` and enter `keydown` event.
When I tested on ChromeVox (my default screen reader), the form was not correctly added to the tab flow.

I also tried marking up the edit actions as links.
This resulted in the screen reader describing each portlet action as an internal link, and when clicked/entered jumps to the edit form.
However there was a flash of old content when the screen reader started reading before the AJAX request came back with the updated `<form>` content.

PR #772, opened 2016-11-18
UP-4770: Label user administration inputs - master

ChristianMurphy:
https://issues.jasig.org/browse/UP-4770

#### Issue

Code Snippet

``` html
<input id="username" name="username" type="text" value="">
...all text input/password instances.
This text input element does not have a name available to an accessibility API. Valid names are: label element, title attribute.
Code Snippet
<input id="Pluto_40_ctf9_14_queryValue" type="text" name="queryValue" data-com.agilebits.onepassword.user-edited="yes">
```

This select element does not have a name available to an accessibility API. Valid names are: label element, title attribute.

Code Snippet

``` html
<select id="Pluto_40_ctf9_14_queryAttribute" name="queryAttribute"> <optio...</select>
<select><option value="">Name</option><...</select>
<select><option value="">Username</opti...</select>
```

Code Snippet

``` html
<select id="Pluto_40_ctf9_14_queryAttribute" name="queryAttribute"> <optio...</select>
<input id="Pluto_40_ctf9_14_queryValue" type="text" name="queryValue" data-com.agilebits.onepassword.user-edited="yes">
<select><option value="">Name</option><...</select>
<select><option value="">Username</opti...</select>
```

#### Resolution

Added `<label>` for inputs.

### DIFF
--- a/uportal-war/src/main/webapp/WEB-INF/flows/edit-account/editLocalAccount.jsp
+++ b/uportal-war/src/main/webapp/WEB-INF/flows/edit-account/editLocalAccount.jsp
@@ -62,17 +62,7 @@
             <!-- Portlet Section -->
             <div class="portlet-section" role="region">
                 <div class="content">
-                    <table class="portlet-table table table-hover">
-                        <thead>
-                            <tr>
-                                <th>
-                                    <spring:message code="attribute.name"/>
-                                </th>
-                                <th>
-                                    <spring:message code="attribute.value"/>
-                                </th>
-                            </tr>
-                        </thead>
+                    <table class="portlet-table table table-hover" role="presentation">
                         <tbody>
                             <c:if test="${ accountForm.id < 0 }">
                                 <tr>
@@ -120,17 +110,7 @@
                 </div>
                 <div id="${n}standardAttributes" class="content">
 
-                    <table class="portlet-table table table-hover">
-                        <thead>
-                            <tr>
-                                <th>
-                                    <spring:message code="attribute.name"/>
-                                </th>
-                                <th>
-                                    <spring:message code="attribute.value"/>
-                                </th>
-                            </tr>
-                        </thead>
+                    <table class="portlet-table table table-hover" role="presentation">
                         <tbody>
 
                             <!-- Print out each attribute -->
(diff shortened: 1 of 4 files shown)


uportal-war/src/main/webapp/WEB-INF/flows/edit-account/editLocalAccount.jsp
```diff
@@ -62,62 +62,78 @@
...
                 </div>
                 <div id="${n}standardAttributes" class="content">
-                
-                    <table class="portlet-table table table-hover">
-                        <thead>
-                            <tr>
-                                <th><spring:message code="attribute.name"/></th>
-                                <th><spring:message code="attribute.value"/></th>
-                            </tr>
-                        </thead>
```
bjagg: Was this omitted deliberately?

uportal-war/src/main/webapp/WEB-INF/flows/edit-account/editLocalAccount.jsp
```diff
@@ -127,50 +143,54 @@
...
                 <div id="${n}customAttributes" class="content">
-                
-                    <table class="portlet-table table table-hover">
-                        <thead>
-                            <tr>
-                                <th><spring:message code="attribute.name"/></th>
-                                <th><spring:message code="attribute.value"/></th>
-                                <th></th>
-                            </tr>
-                        </thead>
```
bjagg: (as above)

uportal-war/src/main/webapp/WEB-INF/flows/edit-account/editLocalAccount.jsp
cousquer: yes because this table is a layout table
If the table is used only to visually layout elements within the content
Check that the table does not contain any th elements.
Check that the table element does not contain a non-empty summary attribute.
Check that the table element does not contain a caption element.
[F46: Failure of Success Criterion 1.3.1 due to using th elements, caption elements, or non-empty summary attributes in layout tables](https://www.w3.org/TR/WCAG-TECHS/F46.html)

uportal-war/src/main/webapp/WEB-INF/flows/edit-account/editLocalAccount.jsp
cousquer: as above 😉

PR #776, opened 2016-11-28
UP-4775 : Use alternativeMaximizedLink into portlet-titlebar - master

jgribonvald:
Replace the portlet maximized url by alternativeMaximizedLink when setted to open directly in a new browser tab. For master branch


bjagg:
Tests positive

PR #781, opened 2016-12-08
UP-4777: Guard for missing dynamic variable in skin generator - master

ChristianMurphy:
https://issues.jasig.org/browse/UP-4777

#### Issue

Skin generator crashes when there is a missing dynamic variable.

#### Resolution

Check if value is empty or missing, skip that variable and warn.


cousquer:
This is bloody interesting. 👍

PR #785, opened 2016-12-13
UP-4783: Import Export Portlet - Add missing id input - rel-4-3-patches

ChristianMurphy:
https://issues.jasig.org/browse/UP-4783

#### Issue

Import Export portlet, delete view has an `<label>` that does not have an associated `<input>`

#### Resolution

Remove unused `<label>`

***edit***

id `<input>` and `<label>` are actually both needed. The `<input>` was accidentally removed before.


ChristianMurphy:
Thanks @drewwills and @cousquer `id` input and label has been added to the delete view in the portlet.

screencapture-localhost-8080-uportal-p-importexportportlet-ctf3-max-render-up-1481817206244

PR #787, opened 2016-12-20
NOJIRA: Gradle fail will stop ant build

bpowell:
If a build of a gradle project fails, it will stop the ant build. Also added a new ant target to switch to Spring 4 and build the Soffit projects and then switch back the version of Spring


build.xml
```diff
@@ -712,7 +722,7 @@
          | uPortal is moving from Ant/Maven to Gradle.  We are porting sources bit by bit.
          | Once we can remove Maven, we can remove this target too.
          +-->
-        <exec executable="./gradlew">
+        <exec executable="./gradlew" failonerror="true">
```
ChristianMurphy: :+1:

build.xml
```diff
@@ -159,6 +159,16 @@
     </target>
 
     <!-- Public API -->
+    <target name="spring4-soffit" description="Rebuild uPortal-soffit packages with Spring 4">
```
ChristianMurphy: Using one build system to modify another build system seems risky to me.
Could this be managed completely within `gradle`?

build.xml
```diff
@@ -159,6 +159,16 @@
...
+        <property name="spring3version" value="3.2.9.RELEASE" />
+        <property name="spring4version" value="4.2.6.RELEASE" />
+
+        <replace file="gradle.properties" token="${spring3version}" value="${spring4version}"/>
```
ChristianMurphy: Using a hardcoded replace could cause issues if the `springVersion` is manually changed.
It could be safer to use the ant [propertyTask](https://ant.apache.org/manual/Tasks/propertyfile.html) or call the [propertyTask from gradle](http://stackoverflow.com/a/20549712).

bpowell: I think this comment and the one above are better solved by upgrading the version of Spring in uPortal to version 4. We only have a couple of days left of Spring 3.2.x support: https://spring.io/blog/2015/06/15/spring-framework-3-2-x-eol-on-dec-31-2016 I think leaving this how it is for right now is a good enough band-aid until Spring 4 is a thing for us.

bjagg:
There is a very challenging issue with upgrading uPortal with a more modern version of Spring. While playing around with the versions with the tools adds some otherwise unnecessary complexity, it may be something we will need to adopt for now.

drewwills:
Folks,

If things go according to plan, Ant & Maven will not be a part of this repo by mid-year.

build.xml
drewwills: This part makes a lot of sense.

drewwills:
@bpowell and @asgrant -- I agree that uPortal needs to update the version of the Spring Framework to at least 4.x, possibly 5.x.  This is a major priority.

FWIW, I'm not having any difficulty when I build my Soffit jars against Spring 3.x (the version in uPortal) but use them in a Spring Boot application based on Spring 4.x.  So far, tt just works.

Here is some of the output from `$ gradle dependencies` in my Spring Boot project (`soffit-samples`)...

```
 Runtime classpath for source set 'main'.

[...]

+--- org.jasig.portal:uPortal-soffit-renderer:5.0.0-SNAPSHOT
|    +--- org.jasig.portal:uPortal-soffit:5.0.0-SNAPSHOT
|    |    +--- io.jsonwebtoken:jjwt:0.6.0
|    |    |    \--- com.fasterxml.jackson.core:jackson-databind:2.4.2 -> 2.6.6
|    |    |         +--- com.fasterxml.jackson.core:jackson-annotations:2.6.0 -> 2.6.6
|    |    |         \--- com.fasterxml.jackson.core:jackson-core:2.6.6
|    |    +--- org.apache.commons:commons-lang3:3.4
|    |    +--- org.jasypt:jasypt:1.9.2
|    |    +--- org.slf4j:slf4j-api:1.7.21
|    |    \--- org.springframework:spring-web:3.2.9.RELEASE -> 4.2.6.RELEASE
|    |         +--- org.springframework:spring-aop:4.2.6.RELEASE
|    |         |    +--- aopalliance:aopalliance:1.0
|    |         |    +--- org.springframework:spring-beans:4.2.6.RELEASE
|    |         |    |    \--- org.springframework:spring-core:4.2.6.RELEASE
|    |         |    \--- org.springframework:spring-core:4.2.6.RELEASE
|    |         +--- org.springframework:spring-beans:4.2.6.RELEASE (*)
|    |         +--- org.springframework:spring-context:4.2.6.RELEASE
|    |         |    +--- org.springframework:spring-aop:4.2.6.RELEASE (*)
|    |         |    +--- org.springframework:spring-beans:4.2.6.RELEASE (*)
|    |         |    +--- org.springframework:spring-core:4.2.6.RELEASE
|    |         |    \--- org.springframework:spring-expression:4.2.6.RELEASE
|    |         |         \--- org.springframework:spring-core:4.2.6.RELEASE
|    |         \--- org.springframework:spring-core:4.2.6.RELEASE
|    \--- org.springframework:spring-webmvc:3.2.9.RELEASE -> 4.2.6.RELEASE
|         +--- org.springframework:spring-beans:4.2.6.RELEASE (*)
|         +--- org.springframework:spring-context:4.2.6.RELEASE (*)
|         +--- org.springframework:spring-core:4.2.6.RELEASE
|         +--- org.springframework:spring-expression:4.2.6.RELEASE (*)
|         \--- org.springframework:spring-web:4.2.6.RELEASE (*)
```

You can see that -- although `org.jasig.portal:uPortal-soffit-renderer:5.0.0-SNAPSHOT` declares Spring dependencies for version 3.2.9.RELEASE, they are being "marshaled up" to version 4.2.6.RELEASE (to work with Spring Boot).

I'm going to add `failonerror="true"` to the `build.xml`, but I'd like to pass on the rest of this PR.  Please get back to be in the next 48 hours with any objections, otherwise I'd like to close this PR.

bpowell:
@drewwills  sorry for the late reply. We have been a little busy! I still think it makes sense to merge the rest in. The reason being, I'd rather be explicit with versions rather than implicit. We have had issues in the past were we think something is using version x.y.z but is really using a.b.c because of some dependency pulling in something else.

ChristianMurphy:
Using one build system to modify another seems too risky to me, even as a band-aid. :-1:
If this was purely in Gradle it could be okay, but with automatic marshaling between Spring version 3 and version 4, using implicit versions could be a low risk band-aid that requires no build changes. :face_with_head_bandage:

drewwills:
@bpowell, @ChristianMurphy, @apetro, @bjagg,

Please take a look at...

  - https://github.com/Jasig/uPortal/commit/f0805a00465795102796d3c7ffef93628887bf83

I believe this change addresses the remaining anxieties WRT this PR.  It means that the `uPortal-soffit-renderer` module will...

  - Build against Spring version 4.3.1.RELEASE
  - Cary Spring-related transitive dependencies that are version 4.3.1.RELEASE

This change should save you from having to build twice.  We are also working on the general Spring 4 update.

PR #788, opened 2016-12-28
UP-4784:  Make Favorites permissions-governed

drewwills:
https://issues.jasig.org/browse/UP-4784

Currently any authenticated user may favorite any portlet that he or she can render.

It would be better to control this behavior through permissions.  Some uPortal adopters will want to restrict what can appear in the Favorites portlet to certain portlets or groups of portlets.


uportal-war/src/main/webapp/WEB-INF/jsp/Marketplace/portlet/entry.jsp
```diff
@@ -273,6 +273,7 @@
         </a>
     </li>
     <li class="divider"></li>
+    {% if (canFavorite) { %}
```
ChristianMurphy: maybe use JSTL instead?
``` jstl
<c:if test="canFavorite">
   ...
</c:if>
```

uportal-war/src/main/java/org/apereo/portal/security/provider/AnyUnblockedGrantPermissionPolicy.java
```diff
@@ -155,7 +155,7 @@ public boolean doesPrincipalHavePermission(
         boolean rslt;
         try {
             // Track groups we've already explored to avoid infinite loop
-            final Set<IGroupMember> seenGroups = new HashSet<IGroupMember>(100);
+            final Set<IGroupMember> seenGroups = new HashSet<>(100);
```
ChristianMurphy: Question purely from curiosity, why start the initial capacity of the set to 100?

uportal-war/src/main/webapp/WEB-INF/jsp/Marketplace/portlet/entry.jsp
```diff
@@ -273,6 +273,7 @@
...
     <li>
         <spring:message code="add.this.portlet.to.my.favorite" text="Add this Portlet to My Favorites" var="atptmfTitle"/>
```
cousquer: This message already exists in [Messages.properties#L47](https://github.com/Jasig/uPortal/blob/master/uportal-war/src/main/resources/properties/i18n/Messages.properties#L47)

uportal-war/src/main/webapp/WEB-INF/jsp/Marketplace/portlet/entry.jsp
drewwills: It's hard to say in this case.  This UI is build with Underscore, which was briefly popular for portlet development some time ago.  I don't know it well, and I'd say most of the folks who might need to maintain this page don't know it well.

But I don't have time to rewrite it.  I went with the Underscore-style conditional because this template has loads more Underscore-style conditional and looping right beside it.  I din't want to make things more confusing by introducing yet another syntax within the same section.

drewwills:
@cousquer -- c'est fait!

uportal-war/src/main/java/org/apereo/portal/security/provider/AnyUnblockedGrantPermissionPolicy.java
drewwills: The core Java collection classes scale automatically, but specifying an `initialCapacity` is a way to provide a "hint" about what size you think the thing will probably need to support.  If you do a good job estimating -- the theory goes -- the code will perform better.

I didn't change the `initialCapacity`, FWIW;  I didn't apply fresh consideration as to whether this number is a good one or whether we should specify one at all.

ChristianMurphy: > The core Java collection classes scale automatically, but specifying an initialCapacity is a way to provide a "hint" about what size you think the thing will probably need to support. If you do a good job estimating -- the theory goes -- the code will perform better.

`HashSet` grows in amortized constant time `O(1)`, which is pretty good.
Unless the magic number `100` has some meaning here (which should be documented), the initial allocation is probably slower than using the defaults letting `HashSet` grow on its own.

> I didn't change the initialCapacity, FWIW; I didn't apply fresh consideration as to whether this number is a good one or whether we should specify one at all.

Fair enough

uportal-war/src/main/java/org/apereo/portal/security/IPermission.java
```diff
@@ -45,18 +59,47 @@
    */
   public static final String PORTLET_BROWSE_ACTIVITY = "BROWSE";
 
+  /**
+   * Portlet subscribe permission to view ("browse") marketplace entry.
```
apetro: :-1: JavaDoc documentation does not match the static field being documented.

drewwills:
@apetro -- FWIW, SCCCD (a unit within CCC) want to have a portal with maybe 70% app-launchers, 30% other content.  They want to use Favorites to allow users to select their own most-used app-launcher style portlets only.

uportal-war/src/main/java/org/apereo/portal/security/IPermission.java
drewwills: I'll clean that up.

cousquer:
@apetro  @drewwills 
I was thinking at these use cases :
- restrict favorites to a type of portlets like CCC,
- "one shot portlet" : (UPMC) we can lanch "a special event portlet" : it's life is just 6 hours and then it will never exist anymore. what happens if a user favorites this portlet during its existance ?

PR #790, opened 2016-12-29
Modernize supported browser documentation.

apetro:
Riffing on [uportal-dev@ discussion](https://groups.google.com/a/apereo.org/d/topic/uportal-dev/ipZKN9THoBM/discussion) of what browsers uPortal 5 ought to support.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,6 +60,10 @@ Consider pushing your changes to a topic branch in your (public) fork of the rep
 
 _Prior to committing, if you want to pull in the latest upstream changes  (highly appreciated btw), please use rebasing rather than merging.  Merging creates "merge commits" that really muck up the project timeline._
 
+## Test
+
+If your change is to a user-facing experience in MyUW, it should not regress support for [supported browser versions](http://jasig.github.io/uPortal/SUPPORTED_BROWSERS.html).
+
 ## Submit
 
 * You submitted a [Contributor License Agreement][], right?  Seriously, Apereo cannot accept your contribution other than under a Contributor License Agreement.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -62,7 +62,7 @@ _Prior to committing, if you want to pull in the latest upstream changes  (highl
 
 ## Test
 
-If your change is to a user-facing experience in MyUW, it should not regress support for [supported browser versions](http://jasig.github.io/uPortal/SUPPORTED_BROWSERS.html).
+If your change is to a user-facing experience, it should not regress support for [supported browser versions](http://jasig.github.io/uPortal/SUPPORTED_BROWSERS.html).
 
 ## Submit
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -60,6 +60,10 @@ Consider pushing your changes to a topic branch in your (public) fork of the rep
 
 _Prior to committing, if you want to pull in the latest upstream changes  (highly appreciated btw), please use rebasing rather than merging.  Merging creates "merge commits" that really muck up the project timeline._
 
+## Test
+
+If your change is to a user-facing experience, it should not regress support for [supported browser versions](http://jasig.github.io/uPortal/SUPPORTED_BROWSERS.html).
```
ChristianMurphy: :art: Relative links usually work better, it allows the parser to determine how the link should be expanded.

``` md
[supported browser versions](SUPPORTED_BROWSERS.md)
```

bjagg:
+1

docs/SUPPORTED_BROWSERS.md
```diff
@@ -0,0 +1,29 @@
+# Policy
```
ChristianMurphy: :art: Usually its best to have a single h1 header

``` md
# Supported Browsers

## Policy

...

## Practice

...

## Rational

...
```

docs/SUPPORTED_BROWSERS.md
```diff
@@ -0,0 +1,31 @@
+# Supported Browsers
+
+## Policy
+
+uPortal supports **the latest general audience release** and **one prior release** of these major browsers:
+
++ Chrome (on Android, MacOS, Windows) (including support for TalkBack)
++ Firefox (on MacOS and on Windows)
```
cousquer: Could you add Linux to Firefox ? something like:
`Firefox (on MacOS, IOS, Android, Windows, Linux) (including support for ChromeVox)`
Sometimes Linux users get upset very very quickly if they feel they are being forgotten.

docs/SUPPORTED_BROWSERS.md
```diff
@@ -0,0 +1,31 @@
+# Supported Browsers
+
+## Policy
+
+uPortal supports **the latest general audience release** and **one prior release** of these major browsers:
+
++ Chrome (on Android, MacOS, Windows) (including support for TalkBack)
```
cousquer: Chrome (on Android, MacOS, IOS, Windows) (including support for TalkBack)

docs/SUPPORTED_BROWSERS.md
apetro: Here's the thing: No browser support fairy comes along and makes it true that uPortal actually supports the browsers targeted in this policy.  Supporting browsers is work. I think the best strategy for uPortal is to commit to as little as we can get away with and then really fulfill those commitments. 

If it proves feasible, sustainable to fulfill a more modest browser support policy, then sure, uPortal could look to make bigger commitments. :)

PR #792, opened 2016-12-30
UP-4785:  Support for multiple guest users (who may have different content)

drewwills:
https://issues.jasig.org/browse/UP-4785

Since the dawn of time uPortal has assigned all unauthenticated users to the same user account:  _guest_.  Not all uPortal deployments support anonymous access, but the vast majority of those that do show *EXACTLY* the same content to every unauthenticated user.

But sometimes that's not what you want.  Sometimes you want to offer a customized guest experience based on the campus, tenancy, language, _etc._ that is used to access the portal (even anonymously).  This task has been difficult, traditionally, due to how aggressively content -- especially guest content -- is cached in the portal.

We should provide the ability to configure multiple guest accounts, and a pluggable, open-ended way to choose a guest account when a user accesses the portal anonymously.

This improvement should (I believe) supersede the need to implement several other tickets in JIRA that describe (risky?) changes to how caching works intending to support varied guest experiences.  I think this feature will cover those use cases, and I believe it will be less risky & disruptive to the codebase as a whole.  (I will try to link all those issues.)

### DIFF
--- a/uportal-war/src/main/java/org/apereo/portal/security/provider/PersonImpl.java
+++ b/uportal-war/src/main/java/org/apereo/portal/security/provider/PersonImpl.java
@@ -232,9 +232,7 @@ public class PersonImpl implements IPerson {
      * <p>
      * This person is a "guest" if both of the following are true:
      * <ol>
-     *   <li>This person's user name matches the value of the property
-     *       <code>org.apereo.portal.security.PersonImpl.guest_user_name</code>
-     *       in <code>portal.properties</code>.</li>
+     *   <li>This person's user name is listed as a guest user account.</li>
      *   <li>This person does not have a live instance ISecurityContext that 
      *       states he/she has been successfully authenticated.  (It can be 
      *       either null or unauthenticated.)</li>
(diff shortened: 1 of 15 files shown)


uportal-war/src/main/java/org/apereo/portal/security/provider/PersonImpl.java
```diff
@@ -244,13 +242,13 @@ public void setFullName(String sFullName) {
      */
     @Override
     public boolean isGuest() {
-        boolean isGuest = false;  // default
+        boolean rslt = false;  // default
         String userName = (String) getAttribute(IPerson.USERNAME);
-        if (PersonFactory.GUEST_USERNAME.equalsIgnoreCase(userName) && 
+        if (PersonFactory.GUEST_USERNAMES.contains(userName) &&
```
ChristianMurphy: :art: is `rslt` needed here?
could this directly return the result?

``` java
String userName = (String) getAttribute(IPerson.USERNAME);
return PersonFactory.GUEST_USERNAMES.contains(userName) && (m_securityContext == null || !m_securityContext.isAuthenticated());

Comment (drewwills) [inline uportal-war/src/main/java/org/apereo/portal/security/provider/PersonImpl.java]: The method could be implemented exactly like that -- and it would be correct, and it would be fine -- but here I'm modestly concerned about readability by humans.  There are a lot of things happening on that 2nd line in your example... it might be too easy to twist the logic trying to hold it in one's brain all at once.  Some times it's more readable when it's broken down into steps.

For my part, I mostly followed the number of steps that was already there.  I didn't bring much fresh energy to thinking about the preferable number of steps.

I think there are some who would have looked at those lines and broke it down more.

Comment (ChristianMurphy) [inline uportal-war/src/main/java/org/apereo/portal/security/provider/PersonImpl.java]: > There are a lot of things happening on that 2nd line in your example...

The same things are currently happening on line 247.

> Some times it's more readable when it's broken down into steps.

How about

``` java
String userName = (String) getAttribute(IPerson.USERNAME);
boolean isGuest = PersonFactory.GUEST_USERNAMES.contains(userName);
boolean isAuthenticated = m_securityContext != null && m_securityContext.isAuthenticated();
return isGuest && !isAuthenticated;
```

PR #802, opened 2017-01-04
Ability to alter soffit modelandview, only send encryptedToken

kdjohnson:
Trying to modify the PortalRequest is causing a bunch of issues with trying to access the Bearer encryptedToken. So I believe that just passing the encryptedToken around allows for a few things. One, it allows for the render modelandview for the soffit to be altered thus allowing for user specific data to be     added. Also, it has the added security benefit of only having the encryptedToken visible.

### DIFF
--- a/uPortal-soffit-renderer/src/main/java/org/apereo/portal/soffit/renderer/SoffitRendererController.java
+++ b/uPortal-soffit-renderer/src/main/java/org/apereo/portal/soffit/renderer/SoffitRendererController.java
@@ -82,8 +82,6 @@ public class SoffitRendererController {
      */
     public static final String CACHE_MAXAGE_PROPERTY_FORMAT = PROPERTY_PREFIX + "%s.cache.max-age";
 
-    private static final String PORTAL_REQUEST_MODEL_NAME = "portalRequest";
-
     private static final String DEFAULT_MODE = "view";
     private static final String DEFAULT_WINDOW_STATE = "normal";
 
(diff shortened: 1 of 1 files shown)


bpowell:
At OU we are trying to pass more information to the JSP, for example passing the IP address of the specific soffit to the page so it can grab js/css resources from it self. The only way to do this is per soffit. Also from a security stand point now the page only gets the encrypted token. It also cuts down on the payload being sent, allowing for faster response times.

bpowell:
@drewwills sure that works for me. I'm open from 8am-3pm tomorrow to talk about it.

uPortal-soffit-renderer/src/main/java/org/apereo/portal/soffit/renderer/SoffitRendererController.java
```diff
@@ -123,8 +121,10 @@ public ModelAndView render(final HttpServletRequest req, final HttpServletRespon
         // Set up cache headers appropriately
         configureCacheHeaders(res, module);
 
-        return new ModelAndView(viewName.toString(), PORTAL_REQUEST_MODEL_NAME, portalRequest);
+        ModelAndView mav = new ModelAndView(viewName.toString());
```
apetro: `viewName` is already a `String`, so is the call to `.toString()` necessary?

kdjohnson: After talking with @bpowell. It seems like him and @drewwills  have come to the conclusion of going a different route. Therefore, I'll be closing this request.

PR #809, opened 2017-01-13
UP-4790:  (rel-4-3-patches branch) Cache entries that establish a user's group affiliations ar…

drewwills:
…e not invalidated when that user authenticates

https://issues.jasig.org/browse/UP-4790

##### Checklist

-   [x] the [individual contributor license agreement] is signed
-   [x] commit message follows [commit guidelines]

##### Description of change

We have some use cases that involve passing parameters to the login process that drive user attributes and (in turn) group affiliations.  We are having trouble because group affiliations are very aggressively cached.

In other words, your attributes may be different from one authentication to the next, but the changes are ignored as far as calculating you group affiliations... because the group affiliations are not recalculated (they are taken from cache).

Just like the layout DOM and a few other items in the portal, we need to remove the cache elements that tie you to groups whenever you authenticate.


ChristianMurphy:
@drewwills could you cherrypick or backport 97ed6f7 into this PR?

PR #821, opened 2017-02-14
UP-4802: Add alt text attribute to Background Preference portlet thumbnails

CR96:
##### Checklist
-   [x] the [individual contributor license agreement](https://github.com/Jasig/uPortal/blob/master/CONTRIBUTING.md#individual-contributor-license-agreement) is signed
-   [x] commit message follows [commit guidelines](https://github.com/Jasig/uPortal/blob/master/CONTRIBUTING.md#commit)

##### Description of change

https://issues.jasig.org/browse/UP-4802

Images in the Background Preference portlet currently have no alt text, which isn't ADA compliant. This can be fixed by adding a new portlet-preference containing an alt-text string for each image.

### DIFF
--- a/uportal-war/src/main/java/org/apereo/portal/portlets/backgroundpreference/BackgroundSetSelectionStrategy.java
+++ b/uportal-war/src/main/java/org/apereo/portal/portlets/backgroundpreference/BackgroundSetSelectionStrategy.java
@@ -33,6 +33,9 @@ public interface BackgroundSetSelectionStrategy {
 
     String[] getImageThumbnailSet(PortletRequest req);
 
+    /**
+     * @since 4.3.2
+     */
     String[] getImageCaptions(PortletRequest req);
 
     String getSelectedImage(PortletRequest req);
--- a/uportal-war/src/main/java/org/apereo/portal/portlets/backgroundpreference/BackgroundSetSelectionStrategy.java
+++ b/uportal-war/src/main/java/org/apereo/portal/portlets/backgroundpreference/BackgroundSetSelectionStrategy.java
@@ -33,6 +33,8 @@ public interface BackgroundSetSelectionStrategy {
 
     String[] getImageThumbnailSet(PortletRequest req);
 
+    String[] getImageCaptions(PortletRequest req);
+
     String getSelectedImage(PortletRequest req);
 
     String getBackgroundContainerSelector(PortletRequest req);
--- a/uportal-war/src/main/java/org/apereo/portal/portlets/backgroundpreference/RoleBasedBackgroundSetSelectionStrategy.java
+++ b/uportal-war/src/main/java/org/apereo/portal/portlets/backgroundpreference/RoleBasedBackgroundSetSelectionStrategy.java
@@ -78,6 +78,10 @@ public class RoleBasedBackgroundSetSelectionStrategy implements BackgroundSetSel
         public String getBackgroundContainerSelectorPreferenceName() {
             return prefix + "BackgroundContainerSelector";
         }
+
+        public String getImageCaptionsName() {
+            return "backgroundImageCaptions";
+        }
     }
 
     private static final String[] EMPTY_STRING_ARRAY = new String[0];
@@ -116,6 +120,13 @@ public class RoleBasedBackgroundSetSelectionStrategy implements BackgroundSetSel
             images[i] = evaluateImagePath(images[i]);
         }
         return images;
+    }
+
+    @Override
+    public String[] getImageCaptions(PortletRequest req) {
+        PreferenceNames names = PreferenceNames.getInstance(req);
+        PortletPreferences prefs = req.getPreferences();
+        return prefs.getValues(names.getImageCaptionsName(), null);
     }
 
     @Override
--- a/uportal-war/src/main/java/org/apereo/portal/portlets/backgroundpreference/RoleBasedBackgroundSetSelectionStrategy.java
+++ b/uportal-war/src/main/java/org/apereo/portal/portlets/backgroundpreference/RoleBasedBackgroundSetSelectionStrategy.java
@@ -57,6 +57,7 @@ public class RoleBasedBackgroundSetSelectionStrategy implements BackgroundSetSel
         public static PreferenceNames getInstance(PortletRequest req) {
             return req.isUserInRole(MOBILE_DEVICE_ROLE_NAME) ? MOBILE : DEFAULT;
         }
+
         private final String prefix;
 
         private PreferenceNames(String prefix) {
@@ -79,9 +80,11 @@ public class RoleBasedBackgroundSetSelectionStrategy implements BackgroundSetSel
             return prefix + "BackgroundContainerSelector";
         }
 
-        public String getImageCaptionsName() {
-            return "backgroundImageCaptions";
-        }
+        /**
+         * The portlet-preference containing the image captions.
+         * This preference name will not change.
+         */
+        public static final String IMAGE_CAPTIONS_PREFERENCE_NAME = "backgroundImageCaptions";
     }
 
     private static final String[] EMPTY_STRING_ARRAY = new String[0];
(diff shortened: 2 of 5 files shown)


uportal-war/src/main/java/org/apereo/portal/portlets/backgroundpreference/RoleBasedBackgroundSetSelectionStrategy.java
```diff
@@ -78,6 +78,10 @@ public String getSelectedBackgroundImagePreferenceName() {
         public String getBackgroundContainerSelectorPreferenceName() {
             return prefix + "BackgroundContainerSelector";
         }
+
+        public String getImageCaptionsName() {
+            return "backgroundImageCaptions";
```
apetro: Realistically, is this value ever going to change, be overridden in a sub-class, etc? Would it be better to factor it as a public static field, maybe with some JavaDoc on that field documenting it?

uportal-war/src/main/webapp/WEB-INF/jsp/BackgroundPreference/viewBackgroundPreference.jsp
```diff
@@ -139,8 +139,8 @@
         </a>
         <c:forEach var="image" items="${thumbnailImages}" varStatus="status">
             <a href="#">
-                <img src="${image}" />
-                <span class="caption">Background ${status.index + 1}</span>
+                <img src="${image}" alt=${imageCaptions[status.index]}/>
+                <span class="caption">${imageCaptions[status.index]}</span>
```
apetro: Looks to me like `imageCaptions` defaults to `null` in `RoleBasedBackgroundSetSelectionStrategy` line 129. Will this JSP fail gracefully if the image captions portlet-preference hasn't been set?

apetro: Bigger picture: background images are decoration, right? What's the use case for screen-reader users to select among decorative background images? Would screen reader users be better served by this feature (both the selection of decorative background images, and the presentation of decorative background images) responsively degrading out of the product when used on a screen reader?

That is, screen reader users may not be best served by being given more information to inform selecting a background image. They might be best served by not being bothered about background images? Is this an admin-facing user, a use case of a screen-reader-using tenant administrator selecting a background image for use by non-screen-reader-using tenants?

ChristianMurphy: > Bigger picture: background images are decoration, right? What's the use case for screen-reader users to select among decorative background images?

Whether there is a use case or not, WCAG auditing tools will flag `` tags without the `alt` attribute.
There should be at least an `alt=""` and preferably a `role="presentation"` attached.

ref: https://www.w3.org/WAI/tutorials/images/decorative/

> Would screen reader users be better served by this feature (both the selection of decorative background images, and the presentation of decorative background images) responsively degrading out of the product when used on a screen reader?

This could be beneficial.
`aria-hidden="true"` would do the trick for most screen readers.
Even with graceful degradation, some `alt` attribute needs to be in place to satisfy WCAG audit tools.

cousquer: and because there is just after `<span class="caption">the description of the image</span>` everything is ok.
even a blind user can change his own background image...

cousquer: it is not the final result here in the main content. it is the menu to change the background image...

CR96:
Just updated the actual alt attribute to an empty string since these are decorative images. The caption attribute still shows the description string.

uportal-war/src/main/webapp/WEB-INF/jsp/BackgroundPreference/viewBackgroundPreference.jsp
CR96: @apetro If the portlet-preference hasn't been set the image just shows up in the list without a caption.
@ChristianMurphy @cousquer +1 on the empty alt string with role="presentation" - that fixed a small screen reading issue too.

uportal-war/src/main/java/org/apereo/portal/portlets/backgroundpreference/BackgroundSetSelectionStrategy.java
```diff
@@ -33,6 +33,9 @@
 
     String[] getImageThumbnailSet(PortletRequest req);
 
+    @since 4.3.1
```
ChristianMurphy: @CR96 This PR is targeting `master` (i.e. release `5.0.0`), if it is a candidate to be backported to `rel-4-3-patches` could you make a note of that in the PR's "Description of change" section? Thanks!

Also note that [the `4.3.1` release](https://github.com/Jasig/uPortal/releases/tag/uportal-4.3.1) has been cut, this change would become a part of release `4.3.2` or release `4.4.0`.

CR96: Have the other ADA fixes been targeting release 4.3.2? If so, I don't believe this would be a candidate for backporting.

ChristianMurphy: The many of the remaining ADA tasks are targeting release `4.3.2`, because they are considered fixes instead of new features.

uportal-war/src/main/java/org/apereo/portal/portlets/backgroundpreference/RoleBasedBackgroundSetSelectionStrategy.java
CR96: I have this change, but I'm curious - would it be better to match the style of the surrounding code in the PreferenceNames enum? Putting a static field in with four other methods feels inconsistent.

drewwills:
This PR looks pretty solid -- thanks @CR96!

uportal-war/src/main/webapp/WEB-INF/jsp/BackgroundPreference/viewBackgroundPreference.jsp
```diff
@@ -139,8 +139,8 @@
         </a>
         <c:forEach var="image" items="${thumbnailImages}" varStatus="status">
             <a href="#">
-                <img src="${image}" />
-                <span class="caption">Background ${status.index + 1}</span>
+                <img src="${image}" alt="" role="presentation"/>
```
drewwills: @CR96 -- The title of the PR says "Add alt text," but here it appears the `alt` attribute is blank.  Is there something I'm missing?

Also, I think it would be smart to be prepared for the case that the `backgroundImageCaptions` preference is undefined, which would make the `imageCaptions` array `null` in the JSP.

CR96: @drewwills I made it more specific - the alt text attribute itself was added (it has to be present to pass ADA checks).

I've tested the change without the preference - it just shows the picture without a caption and doesn't appear to cause any crashes.

drewwills: @CR96 

    the alt text attribute itself was added (it has to be present to pass ADA checks)

Okay, but shouldn't the `alt` attribute be non-blank?

ChristianMurphy: In this case it should be okay
https://www.w3.org/WAI/tutorials/images/decorative/

CR96: Also if I added alt text, screen readers would pick it up twice because of the caption.

drewwills: Gotcha -- it looks funny to me, but apparently this is the right thing to do in this case.  No problem.

drewwills:
This solution was merged into master and hand-applied to rel-4-3-patches.

PR #832, opened 2017-03-15
UP-4816: Auto handle line ending formats across operating systems

ChristianMurphy:
##### Checklist

##### Description of change

[UP-4816](https://issues.jasig.org/browse/UP-4816)

Setting `eol=lf` ensures git will checkout files with the line feed ending expected by Google code style.
https://git-scm.com/docs/gitattributes#_effects


drewwills:
I'm not entirely sure how these features work, but I'm a little concerned.

My main concern is that this approach appears to obligate us to maintain a list of file extensions that should be treated as binary and (therefore) left unaltered by the line endings engine.  If it falls to us to maintain that list, I can guarantee we will screw it up.  :(

Based on my reading, it appears that the Git best practice is to allow Windows users to have the files on their disk with `CRLF`, everyone else so have `LF`, and to use `LF` for the copy in the repo.  (I am unconcerned if Google Code Style believes Windows users should have `LF` on their disk;  as long as it's `LF` in the repo, I think we're hitting our goals.)

I see that this commit cleans up "bad" endings in several files.  That part is excellent... :+1:

ChristianMurphy:
> My main concern is that this approach appears to obligate us to maintain a list of file extensions that should be treated as binary and (therefore) left unaltered by the line endings engine. If it falls to us to maintain that list, I can guarantee we will screw it up. 

Git will also attempt to automatically determine which files are binary or not.
https://stackoverflow.com/questions/6119956/how-to-determine-if-git-handles-a-file-as-binary-or-as-text

> Based on my reading, it appears that the Git best practice is to allow Windows users to have the files on their disk with `CRLF` everyone else so have `LF`, and to use `LF` for the copy in the repo. 

That was my original thought as well. (see e3d93b6)
I changed course, instead targeting the linter being fully functional on all operating systems, without changing the styleguide.

> as long as it's `LF` in the repo, I think we're hitting our goals.

I agree, having the files consistent in the repository is of paramount importance.

.gitattributes
```diff
@@ -0,0 +1,15 @@
+# Lock batch file ending
+*.bat text eol=crlf
```
drewwills: Based on my reading, this line says...

    Unless I say otherwise, consider all .bat files anywhere in
    the project to be text, and force their line endings to be crlf

.gitattributes
```diff
@@ -0,0 +1,15 @@
...
+
+# Checkout as OS specific line endings, check in as LF
+* text eol=lf
```
drewwills: Based on my reading, this line says...

    Unless I say otherwise, consider all files (of any kind) anywhere in
    the project to be text, and force their line endings to be LF

In other words, this rule seems to undo the previous rule.

.gitattributes
```diff
@@ -0,0 +1,15 @@
...
+
+# Unset formating for binary files
+*.gif binary
+*.jar binary
+*.jpg binary
+*.png binary
+*.tar.gz binary
+*.tgz binary
+*.ttf binary
+*.zip binary
```
drewwills: Based on my reading, each of these lines tells Git to treat a specific file extension -- `.gif`, `.jar`, _etc._ -- as binary.

These lines override the previous rule (setting everything to text) for the extensions listed.

drewwills:
@ChristianMurphy -- I added in-line comments for each of the rules in the `.gitattributes` to show clearly what my reading says they mean.  I may be reading it wrong -- if I am, please point it out and please also provide a link to something that explains it.

But if my reading is accurate, I propose we tackle this task from another angle...

What are the files we currently lint or are looking to lint?  Which are the files that the Google Code Style has opinions about?  I propose we create a `.gitattributes` file that targets them specifically, leaving other file types unaffected.

I used this command to produce a list of all extensions currently in the repo...

```
find . -type f | perl -ne 'print $1 if m/\.([^.\/]+)$/' | sort -u
```

Here are the extensions to which I expect we want to apply this rule (it's actually longer than just the list of extensions I suspect we're linting)...

```
*.css text eol=lf
*.dtd text eol=lf
*.gradle text eol=lf
*.groovy text eol=lf
*.html text eol=lf
*.java text eol=lf
*.js text eol=lf
*.json text eol=lf
*.jsp text eol=lf
*.less text eol=lf
*.php text eol=lf
*.pl text eol=lf
*.properties text eol=lf
*.sh text eol=lf
*.tld text eol=lf
*.txt text eol=lf
*.wsdl text eol=lf
*.xml text eol=lf
*.xsd text eol=lf
*.xsl text eol=lf
*.yml text eol=lf
```

ChristianMurphy:
@drewwills that is a fair interpretation.
I personally prefer the blanket approach of `* text eol=lf` to be applied. :100: 
But if we need to start with a more limited approach, that is fine.
Progress is progress. :+1:

ChristianMurphy:
@drewwills I tracked down some more file types. Adding an additional filter for only checked in file types assisted with generating more accurate results.

``` sh
git ls-files | perl -ne 'print $1 if m/\.([^.\/]+)$/' | sort -u
```

.gitattributes
```diff
@@ -0,0 +1,30 @@
...
+*.sh         text eol=lf
+*.svg        text eol=lf
+*.tld        text eol=lf
+*.txt        text eol=lf
+*.wsdl       text eol=lf
+*.xjb        text eol=lf
+*.xml        text eol=lf
+*.xsd        text eol=lf
+*.xsl        text eol=lf
+*.yml        text eol=lf
```
drewwills: @ChristianMurphy -- Some of these don't exist.  The command we've been using to identify extensions returns some "red herrings" due to (I suspect) poor handling of compound file extensions.

So `import-user.crn.xml` adds both `crn` and `xml` to the list.

At least these extensions don't (or shouldn't) exist:
  - crn
  - pom

ChristianMurphy: Both exist:
[crn example](https://github.com/Jasig/uPortal/blob/master/uportal-war/src/main/resources/org/apereo/portal/groups/smartldap/init.crn)
[pom example](https://github.com/Jasig/uPortal/blob/master/uportal-portlets-overlay/CalendarPortlet/3rd-party/repository/org/osaf/caldav4j/0.7/caldav4j-0.7.pom)

ChristianMurphy: [`*.md5`](https://github.com/Jasig/uPortal/blob/master/uportal-portlets-overlay/CalendarPortlet/3rd-party/repository/org/osaf/caldav4j/0.7/caldav4j-0.7-javadoc.jar.md5) and [`*.sha1`](https://github.com/Jasig/uPortal/blob/master/uportal-portlets-overlay/CalendarPortlet/3rd-party/repository/org/osaf/caldav4j/0.7/caldav4j-0.7-javadoc.jar.sha1) are two other unusual filetypes that exist.

ChristianMurphy: @drewwills full list of checked in file types

```
bat
crn
css
dtd
ent
entity-type
eot
gif
gitattributes
gitignore
gradle
groovy
gz
html
ico
Integrator
jar
java
jpg
js
json
jsp
less
md
md5
MF
Module
pl
png
pom
properties
psd
rc
sample
sh
sha1
springWebflow
st
svg
tag
template
tgz
tld
ttf
txt
woff
woff2
xjb
xml
xsd
xsl
yml
zip
```

drewwills: Heh!  Ideally we would have neither of those files... but I suppose listing them is fine for mow.

PR #833, opened 2017-03-16
 UP-4817 - Remove console.log debug message

jgribonvald:
##### Checklist

##### Description of change

[UP-4817](https://issues.jasig.org/browse/UP-4817)
Console.log message should be disabled for production. 
Or in a better way a function to log should be called with a parameter setted to active or not console.log messages.


uportal-war/src/main/resources/layout/theme/respondr/respondr.xsl
```diff
@@ -305,7 +305,7 @@
             return false;
           });
           // Console for debugging.
-          console.log("menu", menu, "menuToggle", menuToggle);
+          //console.log("menu", menu, "menuToggle", menuToggle);
```
drewwills: I agree with @ChristianMurphy -- it's better to remove the line than to comment it out.

I took a quick look at the full file.  There are already some other lines nearby -- including more `console.log` statements -- that are commented out.  It would be great to clean those up as well. :)

jgribonvald: On an other side you could need some logs in development, in this case the best would be to provide a function that do logs and which is configured by a XSL property to activate logs.
But in all case I'm agree these lines should be cleaned !

ChristianMurphy:
Also applied to `rel-4-3-patches` as 3ddfbf8889108e1ffa5f67e17195a99d20c7ce40 and f923bdcb41162691c80406248704ecc8502a796d

PR #813, opened 2017-01-24
UP-4792:  Provide an annotation-based mechanism for soffits based on …

drewwills:
…the SoffitRendererController to inject model attributes

https://issues.jasig.org/browse/UP-4792

### Checklist

-   [x] the [individual contributor license agreement] is signed
-   [x] commit message follows [commit guidelines]
-   [x] tests are included
-   [x] documentation is changed or added

### Description of change

This feature is inspired by the `ModelAttribute` annotation in Spring MVC:  https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/bind/annotation/ModelAttribute.html

uPortal soffit support should provide a similar annotation (e.g. `SoffitModelAttribute`) that allows soffit developers to inject arbitrary beans and/or return values on bean methods into the EL context when a JSP is rendered via the SoffitRendererController.

#### Bean class that uses `@SoffitModelAttribute`

```
@Component
public class SnooperModelAttributes {

    private final ObjectMapper mapper = new ObjectMapper();

    @SoffitModelAttribute(value="bearerJson",viewRegex=".*/snooper/.*")
    public String getBearerJson(Bearer bearer) {
        String rslt = null;
        try {
            rslt = mapper.writerWithDefaultPrettyPrinter().writeValueAsString(bearer);
        } catch (JsonProcessingException e) {
            final String msg = "Unable to write the Bearer object to JSON";
            throw new RuntimeException(msg, e);
        }
        return rslt;
    }

    @SoffitModelAttribute(value="portalRequestJson",viewRegex=".*/snooper/.*")
    public String getPortalRequestJson(PortalRequest portalRequest) {
        String rslt = null;
        try {
            rslt = mapper.writerWithDefaultPrettyPrinter().writeValueAsString(portalRequest);
        } catch (JsonProcessingException e) {
            final String msg = "Unable to write the PortalRequest object to JSON";
            throw new RuntimeException(msg, e);
        }
        return rslt;
    }

    @SoffitModelAttribute(value="preferencesJson",viewRegex=".*/snooper/.*")
    public String getPreferencesJson(Preferences preferences) {
        String rslt = null;
        try {
            rslt = mapper.writerWithDefaultPrettyPrinter().writeValueAsString(preferences);
        } catch (JsonProcessingException e) {
            final String msg = "Unable to write the Preferences object to JSON";
            throw new RuntimeException(msg, e);
        }
        return rslt;
    }

[...]

PR #853, opened 2017-04-11
WIP UP-4592 Add uportal docker images, k8s deployments

andrewstuart:
This PR would add the configuration necessary to build docker images that can be deployed standalone or on a kubernetes cluster.

#### Still in progress
-   [ ] CAS login server/redirects
-   [ ] Non-dockerized database support (use pg_dumped files to init external DB)

##### Checklist

-   [ ] [documentation] is changed or added

[documentation]: https://github.com/andrewstuart/uPortal/blob/dockerize/docs/docker-and-kubernetes.md


k8s/chart/values.yaml
```diff
@@ -0,0 +1,11 @@
+# Default values for chart.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+replicas: 1
+image:
+  registry: andrewstuart
```
andrewstuart: This is my public docker hub prefix, and should have the latest build. We'll need someone else to push to any other repository as I only have access to my own.

Until such a time, this is a valid value for anybody to use.

uportal-db/includes/00-uportal-user-and-db.sql
```diff
@@ -0,0 +1,3 @@
+CREATE DATABASE uportal;
+CREATE USER uportal WITH SUPERUSER PASSWORD 'password';
+ALTER USER postgres WITH SUPERUSER PASSWORD 'password';
```
andrewstuart: I was looking back through this and I think at least here the best option is to go with the postgres container's options for using `POSTGRES_USER` and `POSTGRES_PASSWORD`. That lets us use some more native k8s concepts, and keep things flexible for end users. Good call. :grin:

andrewstuart:
Re: Instrumentation, I'd love to get some [prometheus](/prometheus/client_java) instrumentation added to core uPortal, and it would probably make a good separate Jira ticket.

Aside from specific white-box instrumentation, though, there are plenty of black box monitoring solutions (including a lot of prometheus/k8s integrations) that will give you all the info you want about resource usage, cluster status, HTTP response codes, times, cert status, etc. That's a choice that will be specific to the k8s cluster this is deployed on and in many cases should already exist on the cluster.

docs/docker-and-kubernetes.md
```diff
@@ -0,0 +1,56 @@
...
+1. Start the previously mentioned database in docker locally to be used for
+   the uPortal init process
+1. Query the docker daemon for the IP address of the locally-running database,
+   and use this information to create a temporary
+   filters/docker-build.properties file, inserting the IP address.
+1. Use ant to initialize the database into the running docker postgres
+   container.
+1. Commit the database container, and run the Docker build for the uPortal
+   container.
+1. Push both containers to the named REGISTRY
```
ChristianMurphy: See https://github.com/Jasig/uPortal/pull/813#discussion_r101292214 for the previous discussion on markdown order lists.

Outcome from previous discussion

> Write for readability. Use the style that includes the numbers in the documentation source.

k8s/chart/values.yaml
ChristianMurphy: Got it, thanks @andrewstuart! :+1: 
Fine for now, in the future it could be nice if there was an official `uportal` or `apereo` prefix.

drewwills: I'm still new to Docker... do we need a public registry at all?  Can't we simply build a container locally and run it locally?

ChristianMurphy: @drewwills The docker registry is to docker, as maven centeral is to Java.
yes, we can run without it, but using the registry would allow uPortal to distribute pre-built images.

andrewstuart: You can build and run it locally, but for distribution you really want to host it on the hub or a registry of some sort. For k8s, that's a baseline requirement as each node is transient and will need to pull the pre-built image from somewhere the first time it starts up. You definitely don't want to be building the docker image each time a node wants to run the portal for the first time, as it's a noticeably time-consuming process.

drewwills: Gotcha.  That's essentially what I imagined.

I would like uPortal to have an appropriate public registry to push to, whatever that means... some kind of "Docker Central" or an Apereo/uPortal registry.  I think we should work on that.

But for as long as we don't have an appropriate registry, I think...
  - We shouldn't list one in the default configs;  but
  - We should make it easy to specify one, and document (1) how you do that and (2) why you might want to

.gitignore
```diff
@@ -421,3 +421,22 @@ modules.xml
...
+logs
+node_modules
+npm-debug.log
+target/
+uportal-impl
+uportal-portlets-overlay/*/overlays
+docs/_site
+.ruby-version
+Gemfile.lock
+/filters/docker-build.properties
```
ChristianMurphy: It would be nice to combine the duplicated entries.

docs/docker-and-kubernetes.md
andrewstuart: I see that it was briefly touched upon before, but since markdown really isn't the target language here (this docs page will be rendered as html with `<ol><li>...` regardless of the input numbers), I think the better rule would be:

> Write for readability and maintainability.

I don't think using `1.` repeatedly detracts from the readability in the same way using `<li>` repeatedly doesn't retract from readability, since the target audience will almost certainly be reading the rendered HTML. If they're reading the markdown, it's not hard to infer that the steps are ordered, but by using `1.` as a sort of keyword, it makes the markdown maintenance less painful. I've edited enough markdown lists to for this to become a significant enough peeve that I'll advocate a little more strongly for its use.

I'm not trying to be difficult, but on the contrary I think the better paradigm is to err on the side of maintainability when it doesn't affect the final presentation.

Thoughts, @drewwills and @apetro ?

ChristianMurphy: > I don't think using `1.` repeatedly detracts from the readability in the same way using `<li>` repeatedly doesn't retract from readability

Markdown offers the ability for the source in a mono-spaced font to closely resemble the rendered output.
Yes, we could match HTML as close as possible.
However, I would view that as a missed opportunity to have documentation that is very readable both in source and in rendered form.

>  I've edited enough markdown lists to for this to become a significant enough peeve that I'll advocate a little more strongly for its use.

RemarkLint was added to easily identify and resolve markdown stylistic issues.
If you find it frustrating to update by hand the [autofix feature built into RemarkLint](https://github.com/wooorm/remark-lint#using-remark-to-fix-your-markdown) could be a good alternative.

.gitignore
andrewstuart: Done.

k8s/chart/values.yaml
andrewstuart: Unsurprisingly, I think @mmoayyed (presumably, since cas is the only image) already has [an apereo Docker hub group](https://hub.docker.com/r/apereo/cas/) set up. So it's probably just a matter of getting access For this build process, using Makefile for orchestration, I don't think we can quite piggyback on the automated Docker hub builds. AFAIK that essentially requires you to perform the build inside the Dockerfile, which means complicating the build to the point of either using a base image that has all the tools for the build, or installing all the tools and cleaning up in a single `RUN` command in order to avoid dragging all the build tools into the runtime environment.

It probably wouldn't be too time consuming to set up Travis CI with permissions to push to the Apereo group on Docker hub, and at that point we just need to alter the script to additionally call the appropriate make targets so that the image gets pushed automatically.

ChristianMurphy: > AFAIK that essentially requires you to perform the build inside the Dockerfile

Reference: <https://docs.docker.com/docker-hub/builds/>

> It probably wouldn't be too time consuming to set up Travis CI with permissions to push to the Apereo group on Docker hub

Sounds good.
I've also heard good things about [using Circle CI to generate docker images](https://circleci.com/docs/2.0/building-docker-images/).
That is also what [CAS is currently using for CI](https://circleci.com/gh/apereo/cas), maybe @mmoayyed has some insight?

docs/docker-and-kubernetes.md
drewwills: @andrewstuart & @ChristianMurphy -- I will say again here what I said earlier:  it's not important to me that we all do this exactly the same way.  I think both numbering formats are readable, all else being equal.

Makefile
```diff
@@ -0,0 +1,55 @@
+.PHONY: clean start-ctr db ear image deploy-db deploy
```
jhelmer-unicon: Nitpick:  I completely understand *why* this is being done this way, but for general uPortal acceptance, I'd love not to introduce yet another build system.  We already have ant, maven and gradle.  Adding make into the mix just feels gross.  I really hope the long term plan for uPortal is to move to 100% gradle.  This feels counter to that.

docs/docker-and-kubernetes.md
```diff
@@ -0,0 +1,56 @@
+# Docker and Kubernetes
```
ChristianMurphy: entry has been added

build/conf/catalina.properties
```diff
@@ -0,0 +1,125 @@
...
+# - Common non-Tomcat JARs
+# - Tes(JUnit, Cobertura and dependencies)
+tomcat.util.scan.DefaultJarScanner.jarsToSkip=\
+bootstrap.jar,commons-daemon.jar,tomcat-juli.jar,\
+annotations-api.jar,el-api.jar,jsp-api.jar,servlet-api.jar,websocket-api.jar,\
+catalina.jar,catalina-ant.jar,catalina-ha.jar,catalina-tribes.jar,\
+jasper.jar,jasper-el.jar,ecj-*.jar,\
+tomcat-api.jar,tomcat-util.jar,tomcat-coyote.jar,tomcat-dbcp.jar,\
+tomcat-jni.jar,tomcat-spdy.jar,\
+tomcat-i18n-en.jar,tomcat-i18n-es.jar,tomcat-i18n-fr.jar,tomcat-i18n-ja.jar,\
```
ChristianMurphy: Would it be possible to read in the `catalina.properties` and run property replacements for the necessary changes rather than hard coding a configuration?

k8s/chart/values.yaml
drewwills: :+1: for pushing our images to the apereo Docker hub group set up by @mmoayyed.

build/conf/catalina.properties
drewwills: What settings need to be injected into `catalina.properties`?  If everyone can use the same settings, I don't think injection is necessary.

andrewstuart: You know, that's a very good question. I don't think we need this file at all.

andrewstuart: So, I lied. That's what I get for replying from my phone.

This file doesn't get injected at runtime, but it is added to the image in the tomcat conf directory during the build process, since uPortal has a shared jar folder that Tomcat does not load by default. This is added during the docker build process and is not changed at runtime. This is necessary because the base tomcat image just uses the standard tomcat configuration file, with no shared

build/conf/catalina.properties
```diff
@@ -0,0 +1,125 @@
...
+# Path may be relative to the CATALINA_BASE path or absolute. If left as blank,
+# the "common" loader will be used as Catalina's "shared" loader.
+# Examples:
+#     "foo": Add this folder as a class repository
+#     "foo/*.jar": Add all the JARs of the specified folder as class
+#                  repositories
+#     "foo/bar.jar": Add bar.jar as a class repository
+# Please note that for single jars, e.g. bar.jar, you need the URL form
+# starting with file:.
+shared.loader=${catalina.base}/shared/lib/*.jar
```
andrewstuart: This is the critical line that is different from Tomcat's default config.

drewwills: If this comment is connected to the "injection" discussion above (it looks like it is), I would say there's no point in using token replacement here.

ChristianMurphy:
Hey @andrewstuart, :wave: 
Thanks for getting the ball rolling on this. :basketball: 
The discussion it has generated is invaluable. :100: 
Most of the deployment type code has moved to uPortal-start. :arrow_right: 
uPortal [core] is no longer a good fit for it.
uPortal-start may want to adopt this piece by piece, starting with a Docker image.👍🏼
If you have time, a PR to uPortal-start with be amazing! :1st_place_medal: 
Thanks again!

PR #867, opened 2017-04-24
UP-4833: Fix portlet preference name/value hibernate mapping for Post…

groybal:
…gresql, so that these are stored as TEXT and not as large objects (which results in orphaned data).

##### Checklist

##### Description of change

UP-4833: Fix portlet preference name/value hibernate mapping for Postgresql, so that these are stored as TEXT and not as large objects (which results in orphaned data).  See:  https://issues.jasig.org/browse/UP-4833


drewwills:
## Database Schema Change

This is a change to the portal database schema.  It's fine to implement for uP5, but shouldn't be applied to uP4.  Database migration (with Import/Export) will be required.

drewwills:
Nice work @groybal!  :+1: 

This item is ready, and won't produce a conflict with the "Decompose Sources" effort (#886).  Merging.

PR #870, opened 2017-04-28
UP-4835 - removed the running groovy in interactive mode code 

stalele:
Removed the running the groovy in interactive mode code from up-shell ant target and PortelShell.java
Issue  link - https://issues.jasig.org/browse/UP-4835

##### Checklist

##### Description of change


uportal-war/src/main/java/org/apereo/portal/shell/PortalShell.java
```diff
@@ -76,9 +74,6 @@ public static void main(String[] args) throws Exception {
                         new CompilerConfiguration(System.getProperties());
                 final GroovyShell shell = new GroovyShell(binding, conf);
                 shell.run(scriptFile, args);
-            } else {
```
ChristianMurphy: With the else removed, is the conditional on line 69 still needed?

stalele: Yes, I would like to keep it. Think about the scenario tomorrow somebody forget to/does not pass script parameter, we do not want to start Groovy shell.

PR #872, opened 2017-05-01
UP-4838:  Provide a new permission governing access to one's own user…

drewwills:
… attributes

https://issues.jasig.org/browse/UP-4838

##### Checklist

##### Description of change

This PR introduces a the VIEW_OWN_USER_ATTRIBUTE permission.  It allows you to see more of your own attributes than the attributes you can generally see for users visible (to you) in the portal.

Using this new permission is optional -- having the (original) "VIEW_USER_ATTRIBUTE" permission still gives you access to information about yourself.

### DIFF
--- a/uportal-war/src/main/data/default_entities/permission_owner/UP_USERS.permission-owner.xml
+++ b/uportal-war/src/main/data/default_entities/permission_owner/UP_USERS.permission-owner.xml
@@ -32,7 +32,7 @@
     <activity>
         <name>View own attribute</name>
         <fname>VIEW_OWN_USER_ATTRIBUTE</fname>
-        <desc>View the value(s) of one's own user attribute</desc>
+        <desc>View one's own value(s) for the specified user attribute.  This permission is cumulative with VIEW_USER_ATTRIBUTE</desc>
         <targetProvider>userAttributesTargetProvider</targetProvider>
     </activity>
     <activity>
--- a/uportal-war/src/main/data/default_entities/permission_owner/UP_USERS.permission-owner.xml
+++ b/uportal-war/src/main/data/default_entities/permission_owner/UP_USERS.permission-owner.xml
@@ -30,6 +30,12 @@
         <targetProvider>userAttributesTargetProvider</targetProvider>
     </activity>
     <activity>
+        <name>View own attribute</name>
+        <fname>VIEW_OWN_USER_ATTRIBUTE</fname>
+        <desc>View the value(s) of one's own user attribute</desc>
+        <targetProvider>userAttributesTargetProvider</targetProvider>
+    </activity>
+    <activity>
         <name>View user</name>
         <fname>VIEW_USER</fname>
         <desc>View a user.</desc>
--- a/uportal-war/src/main/java/org/apereo/portal/portlets/lookup/PersonLookupHelperImpl.java
+++ b/uportal-war/src/main/java/org/apereo/portal/portlets/lookup/PersonLookupHelperImpl.java
@@ -420,8 +420,9 @@ public class PersonLookupHelperImpl implements IPersonLookupHelper {
     }
 
     /**
-     * Filter the provided set of user attribute names to contain only those the specified principal
-     * has permissions to view.
+     * Filter the specified set of user attribute names to contain only those that the specified
+     * principal may perform <code>IPermission.VIEW_USER_ATTRIBUTE_ACTIVITY</code>.  These are
+     * user attributes the user has general permission to view, for all visible users.
      *
      * @param principal
      * @param attributeNames
@@ -440,12 +441,15 @@ public class PersonLookupHelperImpl implements IPersonLookupHelper {
     }
 
     /**
-     * Filter the provided set of user attribute names to contain only those the specified principal
-     * has permissions to view.
+     * Provide a complete set of user attribute names that the specified principal may view within
+     * his or her own collection.  These will be attributes for which the user has <em>either</em>
+     * <code>IPermission.VIEW_USER_ATTRIBUTE_ACTIVITY</code> or
+     * <code>IPermission.VIEW_OWN_USER_ATTRIBUTE_ACTIVITY</code>.
      *
-     * @param principal
-     * @param generallyPermittedAttributes
-     * @return
+     * @param principal Represents a portal user who wishes to view user attributes
+     * @param generallyPermittedAttributes The collection of user attribute name this user may view
+     *                                     for any visible user
+     * @since 5.0
      */
     protected Set<String> getPermittedOwnAttributes(
             final IAuthorizationPrincipal principal, final Set<String> generallyPermittedAttributes) {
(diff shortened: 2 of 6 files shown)


uportal-war/src/main/data/default_entities/permission_owner/UP_USERS.permission-owner.xml
```diff
@@ -30,6 +30,12 @@
         <targetProvider>userAttributesTargetProvider</targetProvider>
     </activity>
     <activity>
+        <name>View own attribute</name>
+        <fname>VIEW_OWN_USER_ATTRIBUTE</fname>
+        <desc>View the value(s) of one's own user attribute</desc>
```
apetro: Suggest being explicit here that `VIEW_USER_ATTRIBUTE` on an attribute effectively implies this permission. Suggest also updating `desc` of `VIEW_USER_ATTRIBUTE` to note that it effectively implies this permission.

uportal-war/src/main/java/org/apereo/portal/portlets/lookup/PersonLookupHelperImpl.java
```diff
@@ -440,19 +440,44 @@ protected IAuthorizationPrincipal getPrincipalForUser(final IPerson person) {
     }
 
     /**
+     * Filter the provided set of user attribute names to contain only those the specified principal
+     * has permissions to view.
+     *
+     * @param principal
+     * @param generallyPermittedAttributes
```
apetro: I don't offhand know what `generallyPermitted` means in this context. Suggest using JavaDoc comment on the parameter to explain.

uportal-war/src/main/java/org/apereo/portal/portlets/lookup/PersonLookupHelperImpl.java
```diff
@@ -440,19 +440,44 @@ protected IAuthorizationPrincipal getPrincipalForUser(final IPerson person) {
     }
 
     /**
+     * Filter the provided set of user attribute names to contain only those the specified principal
+     * has permissions to view.
```
apetro: This JavaDoc looks like a copy-and-paste of that for `getPermittedAttributes`. Suggest using the JavaDoc to differentiate these methods.

uportal-war/src/main/java/org/apereo/portal/portlets/lookup/PersonLookupHelperImpl.java
```diff
@@ -464,9 +489,14 @@ protected IPersonAttributes getVisiblePerson(
...
                         person.getName())) {
 
-            // if the user has permission, filter the person attributes according
-            // to the specified permitted attributes
-            final Map<String, List<Object>> visibleAttributes = new HashMap<String, List<Object>>();
+            // If the user has permission, filter the person attributes according
+            // to the specified permitted attributes;  the collection of permitted
+            // attributes can be different based on whether the user is trying to
+            // access information about him/herself.
+            final Set<String> permittedAttributes = person.getName().equals(principal.getKey())
```
apetro: On skim I'm surprised that `person.getName()` is being compared with `principal.getKey()`. The `name` property of an `IPersonAttributes` matches against the `key` property of an `IAuthorizationPrincipal`?

uportal-war/src/main/java/org/apereo/portal/security/IPermission.java
```diff
@@ -112,42 +112,59 @@
...
+    String VIEW_USER_ACTIVITY = "VIEW_USER";
+
+    /**
+     * Governs which user attributes are visible within the portal, applies
+     * to attributes of others as well as one's own attributes.
+     */
+    String VIEW_USER_ATTRIBUTE_ACTIVITY = "VIEW_USER_ATTRIBUTE";
+
+    /**
+     * Governs the visibility of one's own user attributes.
```
apetro: Opportunity here to clarify that the `VIEW_USER_ATTRIBUTE` activity is a superset of this activity.

uportal-war/src/main/java/org/apereo/portal/portlets/lookup/PersonLookupHelperImpl.java
```diff
@@ -427,9 +427,9 @@ protected IAuthorizationPrincipal getPrincipalForUser(final IPerson person) {
      * @param attributeNames
      * @return
      */
-    protected Set<String> getAvailableAttributes(
+    protected Set<String> getPermittedAttributes(
```
apetro: Suggest `@since` documenting when this method came into the codebase.

uportal-war/src/main/java/org/apereo/portal/portlets/lookup/PersonLookupHelperImpl.java
```diff
@@ -440,19 +440,44 @@ protected IAuthorizationPrincipal getPrincipalForUser(final IPerson person) {
...
 
     /**
+     * Filter the provided set of user attribute names to contain only those the specified principal
+     * has permissions to view.
+     *
+     * @param principal
+     * @param generallyPermittedAttributes
+     * @return
+     */
+    protected Set<String> getPermittedOwnAttributes(
```
apetro: Suggest `@since` documenting when this method came into codebase.

uportal-war/src/main/java/org/apereo/portal/portlets/lookup/PersonLookupHelperImpl.java
apetro: Would love to see a unit test or three, verifying no regression and addition of the new complexity.

uportal-war/src/main/java/org/apereo/portal/portlets/lookup/PersonLookupHelperImpl.java
drewwills: Yes, I was surprised about that as well.  Apparently all 3 of these are equivalent...

  - `IPerson.getUserName()`
  - `IPersonAttributes.getName()`
  - `IAuthorizationPrincipal.getKey()` (if the principal represents a user)

I added javadoc to `IAuthorizationPrincipal.getKey()` to document this fact.

uportal-war/src/main/java/org/apereo/portal/portlets/lookup/PersonLookupHelperImpl.java
drewwills: @apetro -- I'm supportive of the general practice of using `@since` in this way, but I'm in doubt about this case.

This isn't a new method, but a name change for clarity.  (The old name does not suggest that permissions checking is performed, leaving the the caller wondering if it needs doing.)

uportal-war/src/main/java/org/apereo/portal/security/IPermission.java
drewwills: Adding `@since` here as well.

uportal-war/src/main/java/org/apereo/portal/portlets/lookup/PersonLookupHelperImpl.java
apetro: I'm not objecting to the name change. But note that the name change breaks the API in a non-private method. To the extent that anything depended upon this API, it broke with this name change. To the extent that anything depends on this API going forward, it's depending on an API that exists only in uPortal v5 and later, presumably. To the extent that this isn't actually an API that should be exposed beyond private, let's mark it `private` and I won't want it to have an `@since` tag. :smile:

`@since` tags are meta-valuable in that they encourage thinking about what APIs are being exposed and the cost-benefits of those changes. This name change is fine -- the benefit of clearer naming probably does outweigh the cost. At least I think so, and you do too, which is why you changed it. :+1:. But let's acknowledge that it creates an API that exists only where it exists.

PR #882, opened 2017-05-11
UP-4850: Add "Skip to page content" button in the portal header

CR96:
https://issues.jasig.org/browse/UP-4850
(Indirectly related to https://issues.jasig.org/browse/UP-2599)

##### Checklist

##### Description of change
This button will allow easier access to uPortal's content when using screen reading software.
This was a feature way back in 2010 with Universality - this brings the button into Respondr.

### DIFF
--- a/uportal-war/src/main/resources/layout/theme/respondr/navigation.xsl
+++ b/uportal-war/src/main/resources/layout/theme/respondr/navigation.xsl
@@ -384,4 +384,14 @@
   <!-- =================================================== -->
 
 
+  <!-- ========== TEMPLATE: SKIPNAV ========== -->
+  <!-- =================================================== -->
+  <!--
+   | This template renders the "Skip to page content" button at th top of the page.
+  -->
+  <xsl:template name="skipnav">
+    <a role="button" class="portal-header-skip-nav btn" href="#portalPageBody"><xsl:value-of select="upMsg:getMessage('skip.to.page.content', $USER_LANG)"/></a>
+  </xsl:template>
+  <!-- =================================================== -->
+
 </xsl:stylesheet>
--- a/uportal-war/src/main/resources/layout/theme/respondr/regions.xsl
+++ b/uportal-war/src/main/resources/layout/theme/respondr/regions.xsl
@@ -107,6 +107,7 @@
     -->
     <xsl:template name="region.pre-header">
         <xsl:if test="//region[@name='pre-header']/channel">
+            <a role="button" class="portal-header-skip-nav btn" href="#portalPageBody"><xsl:value-of select="upMsg:getMessage('skip.to.page.content', $USER_LANG)"/></a>
             <div id="region-pre-header" class="portal-user">
                 <xsl:for-each select="//region[@name='pre-header']/channel">
                     <xsl:call-template name="regions.portlet.decorator" />
--- a/uportal-war/src/main/resources/layout/theme/respondr/regions.xsl
+++ b/uportal-war/src/main/resources/layout/theme/respondr/regions.xsl
@@ -107,7 +107,6 @@
     -->
     <xsl:template name="region.pre-header">
         <xsl:if test="//region[@name='pre-header']/channel">
-            <a role="button" class="portal-header-skip-nav btn" href="#portalPageBody"><xsl:value-of select="upMsg:getMessage('skip.to.page.content', $USER_LANG)"/></a>
             <div id="region-pre-header" class="portal-user">
                 <xsl:for-each select="//region[@name='pre-header']/channel">
                     <xsl:call-template name="regions.portlet.decorator" />
--- a/uportal-war/src/main/resources/layout/theme/respondr/respondr.xsl
+++ b/uportal-war/src/main/resources/layout/theme/respondr/respondr.xsl
@@ -718,6 +718,7 @@
             <script src="/uPortal/scripts/respond-1.4.2.min.js" type="text/javascript"></script>
         </head>
         <body class="up dashboard portal fl-theme-mist">
+          <xsl:call-template name="skipnav" />
           <div class="row-offcanvas">
             <div id="up-notification"></div>
             <div id="wrapper">
@@ -862,6 +863,7 @@
         <script src="/uPortal/scripts/respond-1.4.2.min.js" type="text/javascript"></script>
     </head>
     <body class="up dashboard portal fl-theme-mist detachedHeader">
+        <xsl:call-template name="skipnav" />
         <div id="wrapper">
             <xsl:call-template name="region.hidden-top" />
             <xsl:call-template name="region.page-top" />
--- a/uportal-war/src/main/webapp/media/skins/respondr/common/less/header.less
+++ b/uportal-war/src/main/webapp/media/skins/respondr/common/less/header.less
@@ -17,6 +17,27 @@
  * under the License.
  */
 
+/* Hide the "Skip to page content" button offscreen until it's brought into focus */
+.portal-header-skip-nav {
+    position: absolute;
+    top: -1000px;
+    left: -1000px;
+    height: 1px;
+    width: 1px;
+    text-align: left;
+    color: @header-link-color;
+    overflow: hidden;
+
+    &:active, &:focus, &:hover {
+        left: 0;
+        top: .5rem;
+        z-index: 1;
+        width: auto;
+        height: auto;
+        overflow: visible;
+   }
+}
+
 .portal-header {
     background: @header-background-color;
     color: @header-text-color;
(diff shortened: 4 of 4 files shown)


drewwills:
@CR96,

Excellent idea.

Is this button visible to users who are not using a screen reader?  Should it be?

FWIW I'm inclined to believe it shouldn't be inside the pre-header region.  (Regions are only present in the DOM when they contain content.)

drewwills:
Assuming the message is visible only to screen readers, you could put it just withing the `<body>` tag here...

  - https://github.com/Jasig/uPortal/blob/master/uportal-war/src/main/resources/layout/theme/respondr/respondr.xsl#L720

and here...

  - https://github.com/Jasig/uPortal/blob/master/uportal-war/src/main/resources/layout/theme/respondr/respondr.xsl#L864

Even better would be to set up a (single) XSL template that knows how to render that link, then call it from those 2 places.

drewwills:
### A Couple Additional Points

If, rather, the intent is for this link to be visible to "regular" browser users...

  - Why?
  - It should probably be converted to a portlet, and _then_ added to the pre-header

CR96:
@drewwills This isn't intended to be visible to "normal" users - it only appears when brought into focus by a screen reader. 

Try it with GitHub - I based my button on how they handled it.

drewwills:
@CR96 -- Gotcha, I see.  It seems very reasonable.

If we want it to work like that -- where it can become visible to regular users -- I think a portlet in the `pre-header` is a good call.  The example of the `portal-greeting` portlet would be a decent one to follow...

  - https://github.com/Jasig/uPortal/blob/master/uportal-war/src/main/data/default_entities/portlet-definition/portal-greeting.portlet-definition.xml

apetro:
Point of order: let's try not to use normative terms like "regular" and "normal" when talking about the absence of specific disabilities and challenges. Try instead "typical" or, better, "typically sighted", "vision-typical", etc. This is an opportunity to be more precise anyway, and that can be helpful. Is the distinction really about the user, or is it about the browser? "typical visual browsers". Or maybe even, very specifically, what accessibility-enhancing browsers and approaches will be enabled by a given change? As in, "this change allows a MacOS-VoiceOver-in-Chrome user to skip past 40 words of stuff to get to the interesting content".

apetro:
I favor merging this PR. Ability to skip navigation is better than not having this feature. Progress is progress. :+1:

I also favor continuing to wrestle with this post-merge, making more progress after that as feasible. Opportunities to refactor the implementation. Opportunities to further nuance the functionality. Looking around for best practices:

+ Skip-navigation [aligns with `WCAG 2.0 Guideline 2.4.1`](http://accessibility.psu.edu/skipnav/).
+ "It should be visible to sighted users on a keyboard so that audience can take advantage of it." (same source). Interesting. Skip-navigation may not just serve vision-atypical users -- it may also serve mobility-atypical users who can see the later page content just fine thanks but will find tabbing down past the header tedious. As-implemented may be fine for this since the Skip-nav control becomes visible when tabbed to, so keyboard users may come across it. [WebAIM](http://webaim.org/techniques/skipnav/) has more to say about techniques for helping keyboard users to notice the skip link.

drewwills:
In Respondr, we do everything we can to avoid baking UI features into the theme.  It keeps the theme small, and it's easier for individual adopters to change/move/remove elements as they wish.  This notion is especially important -- very important -- when it comes to regions:  there should be no non-pluggable (portlet or soffit) content in a region.

So if we want to bake it directly into the theme XSL (which I would advise against), we really should place it outside of any regions.

But there's an issue with that idea -- the `pre-header` is the most logical/appropriate place for this content to appear.  If we place it above the `pre-header` -- and if it's visible -- the tendency would be for this content to push the `pre-header` down.  (We could play tricks with absolute positioning, but we would then have other things to worry about like forground/background & whether other content might appear under it.)

I can make time to convert this item to a portlet in the `pre-header`.

bpowell:
@drewwills because this is an ADA issue it would be better for this to be baked into the product with no chance of it being accidentally removed. I think if this was made into a portlet, there is a chance it can be removed or messed with. It's a static button only for screen readers. I agree with @apetro let's merge this is now and hash out the rest later if needed.

drewwills:
I'm trying to update this contribution only to take the link out of a region, but I'm getting really bizarre test failures...

```
ERROR:  'The first argument to the non-static Java function 'hasPermission' is not a valid object reference.'
FATAL ERROR:  'Cannot convert data-type 'void' to 'boolean'.'
           :Cannot convert data-type 'void' to 'boolean'.
```

cousquer:
@CR96 @drewwills @apetro @bpowell @ChristianMurphy 
_Here is how to test this PR in pure accessibility audit_
**Understanding criterion of WCAG 2.0** :[ 2.4.1](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms.html) - [2.4.3](https://www.w3.org/WAI/WCAG20/quickref/#qr-navigation-mechanisms-focus-order) - [3.2.3](https://www.w3.org/WAI/WCAG20/quickref/#qr-consistent-behavior-consistent-locations)
**Sufficient Techniques and/or Failures in WCAG 2.0** : [G1](https://www.w3.org/TR/WCAG-TECHS/G1.html) - [G59](https://www.w3.org/TR/WCAG-TECHS/G59.html) - [G123](https://www.w3.org/TR/WCAG-TECHS/G123.html) - G124 - SCR28 - F66

**The practical test : WCAG 2.0 [Level A] **

> On each Web page, are bypass or skip links to groups of important links and to the main content area available (except in particular cases)?

**With the WebDev (Firefox)**

Open the page in two different browser tabs.
In the "CSS" menu, choose "Disable Styles" then "Disable All Styles".
In the page where CSS styles are disabled, check:
The presence, if necessary (see note), of a link giving direct access to the main menu;
The presence, if necessary (see note), of a link giving direct access to the main content area;
That the links work;
That the links are always at the same place (considering the relative order of the page elements, they are always at the top of the page, for example), in each set of pages.
In the page where CSS styles are enabled, check:
That the links are always, visually, at the same place in each set of pages;
That the links appear at least when they receive focus, when the user tabs through the content with the keyboard.
If true, the test is passed.

**With the WebDev (Chrome)**

Open the page in two different browser tabs.
In the "CSS" menu, choose "Disable All Styles".
In the page where CSS styles are disabled, check:
The presence, if necessary (see note), of a link giving direct access to the main menu;
The presence, if necessary (see note), of a link giving direct access to the main content area;
That the links work;
That the links are always at the same place (considering the relative order of the page elements, they are always at the top of the page, for example), in each set of pages.
In the page where CSS styles are enabled, check:
That the links are always, visually, at the same place in each set of pages;
That the links appear at least when they receive focus, when the user tabs through the content with the keyboard.
If true, the test is passed.

**Note:** it is not mandatory that all skip or bypass links are systematically proposed. Generally, the link to main content is always available to skip the navigation menu, however links to the menu or the search form are useless if these elements are located immediately in the top of the page, for example.

drewwills:
I figured out my build errors -- this PR is against `rel-4-3-patches`, not `master`.

I agree it should be applied to both, but the highest priority is `master`.

drewwills:
@CR96, @ChristianMurphy, @bpowell, @cousquer, @ChristianMurphy, et al.,

I took @CR96's changes and applied them to `master` in #896 EXCEPT that I took the link out of the region (should look & work the same).

If we're happy with that, I can apply it to `rel-4-3-patches` as well.

kdjohnson:
@drewwills So you stole his code and he gets no credit for it?

kdjohnson:
@drewwills I guess I just don't understand the reasoning behind accepting the changes, but then also not merging the branch in. Unless, I totally don't understand the process here? Would it not have made more sense to communicate that you'd like to see some changes like taking the link out of the region before merging. My biggest problem is that @CR96 doesn't really get any sort of recognition for the work he's done here outside of this conversation.

drewwills:
@kdjohnson -- I should be able to merge @CR96's commit.  (And FWIW I had planned to do so all along, as far as the branch this PR targets.)

But there is an issue in that the original PR was opened against `rel-4-3-patches`, and new changes are intended to be applied to `master` first.  (Changes that are eligible for merge into previously released versions of uPortal can be applied from `master`.)

I am not sure what will happen when I attempt to cherry-pick @CR96's changes from `rel-4-3-patches` into `master`, since the affected files are known to be incompatible.  But I will put some energy into that.

drewwills:
@kdjohnson

    Would it not have made more sense to communicate that you'd like to see some changes like taking the link out of the region before merging. 

Yes, that makes more sense.

In this case, I began communicating that desire 8 days ago.

Several smart & capable contributors (whose perspectives I value immensely... and I mean that without even a hint of sarcasm) have posted to this PR to say this change is important -- more important perhaps than the changes I proposed, even if those changes are also valuable -- and that they would like to see it merged ASAP.

I also thing this change is important (and should be merged ASAP), but I still value the changes.  So instead of insisting that @CR96 make them (and not merging until they arrive), I decided to "pitch in."

drewwills:
## Update

I was able to reboot the `master` branch version of this delta in a way that pulls @CR96's commit over from where he applied it, so he will be in the commit history of that branch as well.

kdjohnson:
@drewwills I hope none of this is coming off aggressive or anything. It was not very clear to me about the decisions that were being made regarding accepting/merging this code. I appreciate the clarification and effort you've done to resolve that.

drewwills:
I updated this PR (against `rel-4-3-patches`) as well, preserving @CR96's involvement.

I hope to merge both PRs.

PR #883, opened 2017-05-11
WIP: UP-4851 DLM default adding nodes to first permissible rather than last position.

apetro:
##### Checklist

-   [ ] commit message follows [commit guidelines][] (intend squash merge, so only that commit message will matter)
-   [x] tests are included (sort of. testing the behavior change is hard and isn't tested. Test coverage increased on some code in the neighborhood refactored in the course of this enhancement)
-   [x] documentation is changed or added (JavaDoc)

##### Description of change

### DIFF
--- a/uportal-war/src/main/java/org/apereo/portal/layout/dlm/DistributedLayoutManager.java
+++ b/uportal-war/src/main/java/org/apereo/portal/layout/dlm/DistributedLayoutManager.java
@@ -929,6 +929,8 @@ public class DistributedLayoutManager implements IUserLayoutManager, Initializin
             final String nextSiblingId)
             throws PortalException {
 
+        if ( isFragmentOwner ) return false;
+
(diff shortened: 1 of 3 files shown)


apetro:
WIP. Have tried this a few ways. Sleeping on it, realized this way too is fatally flawed. (The current API doesn't distinguish between "I really want you to add this node as the last sibling, on purpose, for reals" and "I just want you to add this node as a child of this other node and I don't care about sibling order I just want to accept the default".  So if I change the default behavior, I break the ability to add a node to the end on purpose. MyUW has no need for ability to add node to the end on purpose, but I imagine others do.

Next idea is to give up on changing the *default* and instead add a new method (and new REST endpoint or so) that more clearly expresses "I don't care about sibling order" to distinguish this from "I care very much that this new node have no next sibling."

apetro:
Superseded by #885 .

PR #885, opened 2017-05-12
UP-4851: Add nodes first-feasible rather than last when order unspecified.

apetro:
Supersedes #883 .

See also [uportal-dev@ thread](https://groups.google.com/a/apereo.org/d/topic/uportal-dev/aspMtOThps8/discussion).

What this does:

+ Adds a layout manager API method for adding a node with the sibling order left unspecified, an implementation detail for the layout manager.
+ Implements the method in DLM with a policy of adding the node to the first sibling position consistent with applicable DLM restrictions.
+ Changes user preferences servlet to use the new less-specific layout manager API add method when the request has not specified the order in which a node should be added.

Why?

Often user layout modifying UIs will want to add an item with that item landing in some kind of "default" order rather than specifying that the item add last. This "often" is evidenced by [Respondr preferences adding portlets into first position](https://groups.google.com/a/apereo.org/d/msg/uportal-dev/aspMtOThps8/gWu0m1UhAQAJ) and by `uPortal-home` wanting to make its "Add to home" UI control add an item to the first rather than last position on the home.

So this change adds an API method for adding in a default order, implements this in DLM with an add-to-first-feasible position policy, and uses this method in the user preferences servlet methods when the add order wasn't otherwise explicitly communicated. This should make it easier to implement uPortal UIs achieving this "just-add-it-to-a-default-spot-under-this-parent" effect, since those UIs won't themselves have to compute the desired (and permissible) add position, they can leave it unspecified and let the layout manager figure it out.

To the extent that the default ordering ought to differ across uPortal implementations, the default ordering is an implementation detail of the layout manager so DLM could evolve or a successor layout manager could be implemented that chooses a different default behavior or makes it configurable.

##### Checklist

-   [ ] commit message follows [commit guidelines][] (get this right on the squash commit)
-   [ ] tests are included (partial. tests supporting refactoring and understanding neighboring code included.)
-   [x] documentation is changed or added (JavaDoc)

### DIFF
--- a/uportal-war/src/main/java/org/apereo/portal/layout/IUserLayoutManager.java
+++ b/uportal-war/src/main/java/org/apereo/portal/layout/IUserLayoutManager.java
@@ -98,6 +98,21 @@ public interface IUserLayoutManager {
             throws PortalException;
 
     /**
+     * Add a node under a parent without specifying in what order it ought to be placed among
+     * siblings. Use addNode(node, parentId, nextSiblingId) to express exactly where among
+     * current children of the parent node the new node ought to be placed. Use this method to
+     * leave sibling order to the discretion of the underlying layout manager.
+     *
+     * @param nodeToAdd non-null node to add to the layout
+     * @param parentId non-null ID of parent under which to add the node
+     * @return new description of the added node, now with a new or updated ID,
+     * or null if not added.
+     * @since 5.0.0
+     */
+    IUserLayoutNodeDescription addNodeInAnyOrder(
+        IUserLayoutNodeDescription nodeToAdd, String parentId);
+
+    /**
      * Move a node (channel or folder) from one location to another.
      *
      * @param nodeId a <code>String</code> value of a node Id.
--- a/uportal-war/src/main/java/org/apereo/portal/layout/dlm/DistributedLayoutManager.java
+++ b/uportal-war/src/main/java/org/apereo/portal/layout/dlm/DistributedLayoutManager.java
@@ -929,6 +929,8 @@ public class DistributedLayoutManager implements IUserLayoutManager, Initializin
             final String nextSiblingId)
             throws PortalException {
 
+        if ( isFragmentOwner ) return false;
+
         if (node == null) { // cannot add a null node
             return false;
         }
@@ -970,8 +972,6 @@ public class DistributedLayoutManager implements IUserLayoutManager, Initializin
         // todo if isFragmentOwner should probably verify both node and parent are part of the
         // same layout fragment as the fragment owner to insure a misbehaving front-end doesn't
         // do an improper operation.
-
-        if ( isFragmentOwner ) return false;
 
         if (parent instanceof IUserLayoutFolderDescription
                 && !(((IUserLayoutFolderDescription) parent).isAddChildAllowed())
--- a/uportal-war/src/main/java/org/apereo/portal/layout/dlm/DistributedLayoutManager.java
+++ b/uportal-war/src/main/java/org/apereo/portal/layout/dlm/DistributedLayoutManager.java
@@ -929,6 +929,10 @@ public class DistributedLayoutManager implements IUserLayoutManager, Initializin
             final String nextSiblingId)
             throws PortalException {
 
+        if (node == null) { // cannot add a null node
+            return false;
+        }
+
         if (parent == null) { // cannot add a node without a parent to add it under
             return false;
         }
--- a/uportal-war/src/main/java/org/apereo/portal/layout/dlm/DistributedLayoutManager.java
+++ b/uportal-war/src/main/java/org/apereo/portal/layout/dlm/DistributedLayoutManager.java
@@ -463,6 +463,70 @@ public class DistributedLayoutManager implements IUserLayoutManager, Initializin
         return null;
     }
 
+    /*
+     * DLM implementation adds to first allowable position among existing children of the parent
+     * node, if any.
+     */
+    public IUserLayoutNodeDescription addNodeInAnyOrder(
+        final IUserLayoutNodeDescription nodeToInsert, final String parentId) {
+
+        final IUserLayoutNodeDescription parentNode = getNode(parentId);
+
+        final String firstAllowableNextSiblingId =
+            firstAllowableNextSiblingId(parentNode, nodeToInsert);
+
+        return addNode(nodeToInsert, parentId, firstAllowableNextSiblingId);
+    }
+
+    /**
+     * Compute the siblingId representing the earliest allowable placement for adding a child node.
+     * @param parent potential parent of the node to be added
+     * @return siblingId representing earliest allowable nextSiblingId for adding a node,
+     * or blank if no such nextSiblingId
+     * @since 5.0.0
+     */
+    private String firstAllowableNextSiblingId(
+        final IUserLayoutNodeDescription parent,
+        final IUserLayoutNodeDescription nodeToInsert) {
+
+        final Enumeration sibIds = getVisibleChildIds(parent.getId());
+        final List sibs = Collections.list(sibIds);
+
+        if (sibs.size() == 0) { // no existing children, so no first allowable next sibling ID
+            return "";
+        }
+
+        // Algorithm:
+        // Starting at the last sibling position.
+        // Try to promote to the next earlier position. Repeat zero to N times,
+        // until failure to promote or until reaching first position.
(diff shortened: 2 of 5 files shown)


uportal-war/src/main/java/org/apereo/portal/layout/IUserLayoutManager.java
```diff
@@ -98,6 +98,21 @@ public IUserLayoutNodeDescription addNode(
...
+     * current children of the parent node the new node ought to be placed. Use this method to
+     * leave sibling order to the discretion of the underlying layout manager.
+     *
+     * @param nodeToAdd non-null node to add to the layout
+     * @param parentId non-null ID of parent under which to add the node
+     * @return new description of the added node, now with a new or updated ID,
+     * or null if not added.
+     * @since 5.0.0
+     */
+    IUserLayoutNodeDescription addNodeInAnyOrder(
```
drewwills: Suggest name change -- I think simply a new overload of `addNode` is a better here...

  - The "any order" part is misleading;  to me it implies rendomness
  - It's weird that this method has a longer/more complicated name than the version that accepts 3 arguments

uportal-war/src/main/java/org/apereo/portal/layout/dlm/DistributedLayoutManager.java
```diff
@@ -452,9 +453,80 @@ public IUserLayoutNodeDescription addNode(
 
             return node;
         }
+
+        if (LOG.isTraceEnabled()) {
```
drewwills: - Suggest `DEBUG` level here
  - Do you have the bandwidth to convert this file to slf4j, since you're in here making changes?

uportal-war/src/main/java/org/apereo/portal/layout/dlm/DistributedLayoutManager.java
```diff
@@ -452,9 +453,80 @@ public IUserLayoutNodeDescription addNode(
...
+    }
+
+    /**
+     * Compute the siblingId representing the earliest allowable placement for adding a child node.
+     * @param parent potential parent of the node to be added
+     * @return siblingId representing earliest allowable nextSiblingId for adding a node,
+     * or blank if no such nextSiblingId
+     * @since 5.0.0
+     */
+    private String firstAllowableNextSiblingId(
```
drewwills: Suggest `calculateFirstAllowableNextSiblingId` or something akin.  Adding a verb to the name will help with clarity.

uportal-war/src/main/java/org/apereo/portal/layout/dlm/DistributedLayoutManager.java
```diff
@@ -452,9 +453,80 @@ public IUserLayoutNodeDescription addNode(
...
+        final IUserLayoutNodeDescription parent,
+        final IUserLayoutNodeDescription nodeToInsert) {
+
+        final Enumeration sibIds = getVisibleChildIds(parent.getId());
+        final List sibs = Collections.list(sibIds);
+
+        if (sibs.size() == 0) { // no existing children, so no first allowable next sibling ID
+            return "";
```
drewwills: An empty String?  Wouldn't `null` be a better response here?

uportal-war/src/main/java/org/apereo/portal/layout/dlm/DistributedLayoutManager.java
```diff
@@ -908,19 +980,35 @@ private Map getPublishedChannelParametersMap(String channelPublishId) throws Por
         }
     }
 
+    @Override
```
drewwills: :+1:

uportal-war/src/main/java/org/apereo/portal/layout/dlm/DistributedLayoutManager.java
```diff
@@ -908,19 +980,35 @@ private Map getPublishedChannelParametersMap(String channelPublishId) throws Por
...
     protected boolean canAddNode(
-            IUserLayoutNodeDescription node,
-            IUserLayoutNodeDescription parent,
-            String nextSiblingId)
+            final IUserLayoutNodeDescription node,
+            final IUserLayoutNodeDescription parent,
+            final String nextSiblingId)
             throws PortalException {
+
+        if ( isFragmentOwner ) return false;
```
drewwills: Fragment owners cannot use this API?

apetro: That's a refactor to make more explicit [a weirdness already in the code](https://github.com/Jasig/uPortal/blob/ce925e5ed28f05585fa3287b869f18bae151c405/uportal-war/src/main/java/org/apereo/portal/layout/dlm/DistributedLayoutManager.java#L952). ( [mentioned on up-dev@](https://groups.google.com/a/apereo.org/d/msg/uportal-dev/aspMtOThps8/8jrhLxlwBAAJ) ). I don't understand why `canAdd` automatically returns false for fragment owners. It might be that this is intended to be defensive because it doesn't currently check whether it's a proposal to add to the managed fragment layout or a personal layout. But it's weird, because one way or another fragment owners are actually allowed to add nodes to their layouts, as I understand it.

uportal-war/src/main/java/org/apereo/portal/layout/dlm/DistributedLayoutManager.java
apetro: Keeping the verb out of it keeps focus on what the method returns? It's not just procedural (calculate the next sibling ID), it's functional (return the next sibling ID). Caller doesn't need to specify the implementation details -- it might calculate it, it might pull it from a cache, it might ...

A more object-oriented approach to this might push this method into the folder-type Node API such that could ask a folder what's the first allowable next sibling ID for an item to be added to it. I'd have done that except DLM already has this procedural approach going on such that a bunch of helper methods would have to come with it or so.

drewwills: It doesn't have to be "calculate;"  some other ideas...

  - choose
  - select
  - get

Its noun-ness makes me expect it to be a class (rather than a method).  Also, it leaves me to wonder if the "siblingId" mentioned in the name is something I'm passing in or receiving.  (I can figure it out by looking a bit deeper... but I think it's better if it just jumps out at you.)

uportal-war/src/main/java/org/apereo/portal/layout/dlm/DistributedLayoutManager.java
ChristianMurphy: > That's a refactor to make more explicit a weirdness already in the code.

:+1:

drewwills: Yeah if that's the case, it's a good change.  :+1:

drewwills:
@apetro are you still planning on squashing?

apetro:
Yup, I'll eventually get back to this. Context switched to other priorities. There's a conference coming up I'm not ready for.

drewwills:
@apetro -- I know the feeling.

I'm trying to get it merged in coordination with the "Decompose Sources" work... trying to be mindful not to create extra work for you (or me).

uportal-war/src/test/java/org/apereo/portal/layout/dlm/DistributedLayoutManagerTest.java
```diff
@@ -0,0 +1,122 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.*;
+
+/**
+ * Unit tests for DLM.
+ */
+public class DistributedLayoutManagerTest {
+
```
bjagg: :1st_place_medal: for adding tests!

bjagg:
I spun up this PR and tried to test it. It started up fine. Used http://localhost:8080/uPortal/api/layout?action=addPortlet&fname=announcements&tabName=Development but my REST tool isn't picking up my session.

ChristianMurphy:
Hey @apetro , it looks like a merge conflict has cropped up. Would you be able to resolve the conflict?

ChristianMurphy:
Hey @apetro, thanks for the pull request!
Unfortunately are merge conflicts that have cropped up that are blocking these changes from being merged.
I'm going to close this PR, feel free to re-open it if you have time to resolve the merge conflicts.

PR #886, opened 2017-05-12
WIP UP-4830:  Decompose sources in uportal-war into Gradle-based submodules

drewwills:
https://issues.jasig.org/browse/UP-4830

##### Checklist

##### Description of change

This is continuing work on UP-4830 (Gradle-based rework).

This effort is picking up steam;  it's likely I can make it all the way through in the next several days.

https://wiki.jasig.org/display/UPC/uPortal+5+Project+Overhaul


drewwills:
I found this post of StackOverflow that seems to match the CI issues we're seeing...

  - http://stackoverflow.com/questions/41998178/gradle-compilejpamodelgenjava-failed

drewwills:
@ChristianMurphy, @apetro, @bjagg, et al. -- This is a good/clean break point.  I'll be looking to merge this work ASAP (later today... certainly after the tests finish).

I'll start a new PR for the next round of these updates.  This next one may be the last -- only 1298 files remain in uportal-war/src/main/java/.  (The more we move, the easier it gets.)

Between this & that, I'll be looking to process as many of the other PRs as possible.

.gitignore
```diff
@@ -417,7 +417,7 @@ modules.xml
 !bootstrap/lib/*.jar
 !uportal-portlets-overlay/CalendarPortlet/3rd-party/repository/org/osaf/caldav4j/0.7/*.jar
 !uportal-portlets-overlay/CalendarPortlet/3rd-party/repository/slide/jakarta-slide-webdavlib/2.2pre1-httpclient-3.0/*.jar
-!uportal-war/src/main/java/org/apereo/portal/permission/target/
-!uportal-war/src/main/webapp/WEB-INF/tags/
+!**/src/**/target/
+!**/src/**/tags/
```
ChristianMurphy: :+1:

drewwills:
Folks,

The CI is complaining of this...

```
':uPortal-concurrency:verifyGoogleJavaFormat'
```
I would greatly prefer it if that would be a warning, rather than red ink.

ChristianMurphy:
@drewwills two reasons:
1. the files were added to the gradle build which is checked
2. the code was checked with Google Java Formatter 1.3, while the project was previously formatted with version 1.2

drewwills:
the code was checked with Google Java Formatter 1.3, while the project was previously formatted with version 1.2

So we bump the version of a dependency and invalidate all our sources?

Are we really expecting that -- going forward -- the sources will essentially always be formatted by an algorithm?  Do we think that makes sense?  (I don't.)

ChristianMurphy:
> So we bump the version of a dependency and invalidate all our sources?

Not really we just need to re-run the autofixer, and everything is good.

> Are we really expecting that -- going forward -- the sources will essentially always be formatted by an algorithm?

Sure, why manually fight with a code style if you can automatically join the club?
Let the creativity be in the solution not in the code style. :smile:

jhelmer-unicon:
I get the need for consistency, but I have to admit, I'm nervous about the specter of massive diffs every time that code formatting tool changes.   :(    I'd almost say pick a version and stick with it forever -- or at least until it no longer works for us.   I don't think we want to arbitrarily reformat a bunch of code *again* just because some tool changed.

drewwills:
@jhelmer-unicon's point is a good one -- if we adopt practices that produce loads of extra deltas, those deltas will come back to haunt adopters when they merge.

jhelmer-unicon:
As someone who is looking down the barrel of a merge to uPortal 5 in the future....   Please don't make my life any harder than it is already going to be...   :(

BTW:  I'd personally prefer to see checkstyle used to ensure coding standards.   I don't like how inflexible this tool seems to be.   I get that we want *a standard*, but let's be honest...  some of the google rules are arbitrary and ugly (2 spaces.  ugh).   The idea is to promote readability.   As a project, we should have the flexibility to choose to deviate from the google style in minor ways if we find the google style to be less readable.  I don't understand what the upside to the javadoc changes that were made here is.    Checkstyle at least allows you to demote some of this stuff to a warning vs. blowing up the build.

drewwills:
I think @jhelmer-unicon summed up my feelings on this topic pretty well.  :+1:

uportal-war/src/main/java/org/apereo/portal/rest/PagsRESTController.java
```diff
@@ -48,9 +48,6 @@
      Tried to use ResponseEntities but the ran into conflicts when using <mvc:annotation-driven/>
     */
 
-    /** Help other Java classes use this API well. */
-    public static final String URL_FORMAT_STRING = "/api/v4-3/pags/%s.json";
```
apetro: Losing a public field?

drewwills: Moved to another class.

There was a class I was trying to move that referenced this field.

bjagg:
Styling wars, yeah! I'll start by saying I am not an Alphabet fan. Nuf said on that, I guess.

Great that we can use their tool to automate this. Maybe post-decompose, someone can go through and fix the source with it, but it certainly should NOT add to the work for this effort today.

When a release is cut, license headers, NOTICE file and JavaDocs are checked and addressed. This can be onerous, but it alleviates that effort from contributors. JavaDocs checking is the worst part of this effort since it requires changing source by hand as opposed to automatic tooling for the other efforts.

We can go all the way and have stylechecker and licensing run on every build. My IDE does some of this. It's the main reason I use it before pushing PRs. Otherwise, I could return to Vim almost exclusively. 

My main concern is the timing and lack of discussion on this change with the community ... or did I miss it? This is not something to be decided in PR comments.

ChristianMurphy:
@bjagg There has been ongoing discussion in <https://groups.google.com/a/apereo.org/d/topic/uportal-dev/FYiSDvmum1I/discussion>
The original Checkstyle + Google code style was merged in https://github.com/Jasig/uPortal/pull/733 the transition to use the Google Java formatter was made in https://github.com/Jasig/uPortal/pull/887.

bjagg:
Oh I missed that thread, lol!

So the move of source files from Maven managed directories to Gradle ones is what triggered the check ... as you've mentioned to me before, @ChristianMurphy

PR #889, opened 2017-05-16
UP-4834 gradle task clean shared

stalele:
##### Checklist

-   [ ] tests are included

##### Description of change

Added cleanShared and cleanTomcat tasks in build.gradle. 
You need to set serverbase property in gradle.properties correctly so that it can find corresponding files and perform clean operations on those


build.gradle
```diff
@@ -44,7 +63,8 @@ subprojects {
         mavenLocal()
         mavenCentral()
     }
-
+    
+//}
```
ChristianMurphy: Left over test comment?

gradle.properties
```diff
@@ -55,3 +55,5 @@ junitVersion=4.12
 mockitoVersion=2.7.6
 slf4jVersion=1.7.21
 springVersion=3.2.9.RELEASE
+#set serverbase property here
+serverbase=
```
ChristianMurphy: Should this be coming from a different properties file? My understanding was that `gradle.properties` would be used for properties that can be used for all projects. `serverbase` is usually custom for each local environment.

build.gradle
```diff
@@ -70,5 +90,9 @@ subprojects {
          */
         compile.extendsFrom provided
     }
+    
+  //  compile name: 'uPortal-classes'
```
ChristianMurphy: Left over test comment?

build.gradle
```diff
@@ -33,6 +35,23 @@ buildScan {
     licenseAgree = 'yes'
 }
 
+	task cleanShared() {
+		group 'Configuration'
+		description 'Removes ALL shared libraries from the servlet container.'
+		def libs = serverbase+"/shared/lib"
```
ChristianMurphy: [Groovy code style recommends using string interpolation over concatenation.](http://groovy-lang.org/style-guide.html#_gstrings_interpolation_multiline)

```groovy
def libs = "${serverbase}/shared/lib"
```

build.gradle
```diff
@@ -33,6 +35,23 @@ buildScan {
...
+		FileTree tree = fileTree(dir: libs)
+		delete fileTree(dir:libs , include: '*')
+	}
+
+	 task cleanTomcat() {
+		group 'Configuration'
+		description 'Removes the deployed uPortal from the servlet container.'
+		def uPortal = serverbase+"/webapps/uPortal"
```
ChristianMurphy: [Groovy code style recommends using string interpolation over concatenation.](http://groovy-lang.org/style-guide.html#_gstrings_interpolation_multiline)

build.gradle
```diff
@@ -33,6 +35,23 @@ buildScan {
...
+		printf uPortal
```
ChristianMurphy: Left over print statement?

stalele: This is not left over I want to print what directory it is deleting when gralde task is executed

gradle.properties
```diff
@@ -34,7 +34,7 @@ projectLicenseDistribution=repo
 projectIssueSystem=JIRA
 projectIssueUrl=https://issues.jasig.org/browse/UP
 
-gradleVersion=3.0
+gradleVersion=3.5
```
ChristianMurphy: Thanks for updating this!

build.gradle
```diff
@@ -69,5 +93,5 @@ subprojects {
          */
         compile.extendsFrom provided
     }
-
-}
+
```
ChristianMurphy: trailing spaces?

build.gradle
```diff
@@ -44,7 +68,7 @@ subprojects {
         mavenLocal()
         mavenCentral()
     }
-
+
```
ChristianMurphy: trailing spaces?

build.gradle
```diff
@@ -3,15 +3,23 @@ description = "Apereo uPortal $version"
...
+task printProps() {
+    println props.getProperty("server.home")
+}
 buildscript {
     repositories {
         maven {
             url 'https://plugins.gradle.org/m2/'
         }
+
```
ChristianMurphy: trailing spaces?

build.gradle
stalele: I do not see any trialing space here... I deleted the }

build.gradle
stalele: I do not see trialing space here

build.gradle
drewwills: :+1: , but could you also tell the user what it is?

```
Removing deployed uPortal from servlet container at location:  {uPortalDeployDir}
```

stalele: I changed the statements a bit "Configured deployed uPortal from servlet container at location:  {uPortalDeployDir}"

The reason I did that when you execute cleanShared you will see that statement even though you have not executed cleanTomcat task. Gradle is prints the printf statment during configuration. I do not want to mislead the developer if they see "Removing...." while gradle is only configuring the task

drewwills:
@stalele 

    You need to set serverbase property in gradle.properties correctly so that it can find corresponding files and perform clean operations on those

Is this still the case?  This requirement is not a good one, because the `gradle.properties` file will (therefore) permanently appear in the "Changes not staged for commit:" list in Git.

Please source the `serverbase` property from build.properties (which is listed in the .gitignore).

build.gradle
```diff
@@ -3,10 +3,24 @@ description = "Apereo uPortal $version"
...
 apply plugin: 'com.gradle.build-scan'
+apply plugin: 'java'
 
 // Adds support for Node.js scripts
 apply plugin: 'com.moowork.node'
 
+def props = new Properties()
+file("build.properties").withInputStream { props.load(it) }
+
+task copySampleProperties(type: Copy) {
```
ChristianMurphy: @stalele If `build.properties` does not exist, Gradle will fail before it can copy it.
This is what is currently happening on CI.
https://travis-ci.org/Jasig/uPortal/builds/233427091#L187-L192

drewwills: How is the CI meant to get its `build.properties` file?

ChristianMurphy: It could copy over the `build.sample.properties`. :man_shrugging:

ChristianMurphy:
@drewwills Using `cp` would also work, adding it as a gradle task mirrors the functionality in https://github.com/Jasig/uPortal/blob/c5d425de9738e73432bf4c2d066b67782495b001/assembly/build.xml#L273-L277

drewwills:
@ChristianMurphy...

    adding it as a gradle task mirrors the functionality in https://github.com/Jasig/uPortal/blob/c5d425de9738e73432bf4c2d066b67782495b001/assembly/build.xml#L273-L277

Does it?  (I don't see that.)

The Ant task is for the Quickstart, and it knows how to change the value of `server.home` in the copy it makes.  This Gradle task does not appear to make that change.  Without that part, I don't think the Gradle version is needed.  (It's not sufficiently valuable to bother including.)

build.gradle
```diff
@@ -3,10 +3,26 @@ description = "Apereo uPortal $version"
...
+    into '.'
+    include 'build.properties.sample'
+    rename { String fileName ->
+    fileName.replace("build.properties.sample", "build.properties")
+  }
+  doLast {
+	  file("build.properties").withInputStream { props.load(it) }
+  }
+}
+
```
bjagg: Is copySampleProperties task needed? build.properties should exist. If not, you can gracefully fail.

build.gradle
```diff
@@ -69,5 +103,5 @@ subprojects {
          */
         compile.extendsFrom provided
     }
-
-}
+    
+}
```
bjagg: Remove whitespace from end of line

build.gradle
stalele: It is needed for CI environment @ChristianMurphy  am I right?

ChristianMurphy: @bjagg @stalele CI either needs to be able to lazy load properties or to copy the file over.
Currently the copy approach is being taken, and gradle gives a cross platform approach to handle copying.

gradle.properties
ChristianMurphy: @stalele the gradle reversion was dropped back to 3.4 in https://github.com/Jasig/uPortal/commit/fbafedf445408bdc91abbaf919c32d2a66398dc9

build.gradle
```diff
@@ -26,6 +27,21 @@ repositories {
     mavenLocal()
 }
 
+def props = new Properties()
+
+task copySampleProperties(type: Copy) {
+ printf "Configuring copying the build.properties.sample into build.properties"
+ from '.'
```
ChristianMurphy: :art: could `printf` and `from` be indented to match line 35?

bjagg:
Seems that the issue is not that the new tasks are called during CI, rather it is that the file does not exist when the gradle files are processed. If so, the cleaner solution is to add `if (!file.exists()) {...}` around loading the properties. Really, the task should fail if the properties file is missing.

build.gradle
```diff
@@ -29,6 +30,27 @@ repositories {
...
+        doLast {
+            file("build.properties").withInputStream { props.load(it) }
+        }
+    }
+    else {
+    println("ERROR:${sampleBuildFile} does not exist") 
+    }
+}
+
 buildscript {
```
stalele: Aren't we suppose to copy contents(/replace) from build.properties.sample to build.properties?

bjagg: No. Only the 'Quick Start' does that as part of the uPortal demo. Devs and implementers need to create and custom the values.

PR #899, opened 2017-05-23
NOJIRA: Document how to integrate build with IDE

ChristianMurphy:
##### Checklist

##### Description of change

Document how to integrate build system with IDE.


docs/developer/other/IDE_INTEGRATION.md
```diff
@@ -0,0 +1,31 @@
...
+1. Open a terminal
+2. `cd` to the uPortal folder
+3. run
+```sh
+./gradlew eclipse
+```
+4. Open eclipse
+6. Select file > import
+7. Search for "Existing Projects into Workspace"
+![open in intellij](../../images/eclipse_import.png)
```
vertein: consider 'open in eclipse' alt text for the open in eclipse picture.

ChristianMurphy: updated

docs/developer/other/IDE_INTEGRATION.md
```diff
@@ -0,0 +1,31 @@
+# IDE Integration
+
```
apetro: This documentation says what to do. I can follow these steps. But why? What do these steps accomplish? How is it better to do this rather than just open the project via IDE without having first done these steps?

drewwills: Good question.

Also -- does this content document anything that covered in very superficial Idea and/or Eclipse documentation?  Does it document anything that Idea and/or Eclipse users would be unlikely to know?

ChristianMurphy: > How is it better to do this rather than just open the project via IDE without having first done these steps?

In olden days it was because IDE imports for gradle missed a lot (things like subprojects would be ignored).
Now there isn't much difference whether use using the plugin or use import.

> Does it document anything that Idea and/or Eclipse users would be unlikely to know?

Unlikely, this would more likely help a developer new to either of the IDEs.

----

Taking the conversation a different direction.
The main reason I even considered adding this was to note a piece of functionality which is not normally run as part of the `gradle install`, developers have to go out of their way to run `gradle idea` or `gradle eclipse` to leverage it.

Would a better approach be to remove both these plugins entirely?

apetro: I'd favor keeping the plugins, keeping this documentation, just enhancing it to describe and maybe link more information about what these plugins do and why one would want to use them.

I was unaware of `gradle idea` before this PR. Ran it before my latest attempt to get a uPortal development environment going in IntelliJ-IDEA. It was significantly less miserable than a recent previous attempt. So, maybe a lot to be gained in guiding developers to run it before opening.

bjagg: I set up copies of uPortal often. Even knowing what to set in the setup wizard, it's a bit of a chore. I appreciate this improvement. :1st_place_medal:

ChristianMurphy: Sounds good, I've added a brief summary of benefits at the top of `IDE_INTEGRATION.md`.

PR #903, opened 2017-05-25
Add layoutObject property to MarketplaceEntry.

apetro:
No longer blocked since [circular dependency][uportal-dev@ thread] resolved.

Enrichens `MarketplaceEntry` with the `LayoutPortlet` for the underlying `PortletDefinition`. Makes the by-fname lookup of `MarketplaceEntry` information (e.g. exposed via JSON web service) more useful and require less cross-indexing with layout data.

Backstory: it turns out `uPortal-home` depends upon this currently MyUW-local modification. Adding this to the MarketplaceEntry (and thus to its resulting JSON) gets `uPortal-home` and `uPortal` `master` slightly closer to integrating.

These web service APIs need thoughtful refactoring, documentation, design. They should probably be substantially different, and they should definitely be substantially less complicated and more performant, than they are. But at this moment in the evolution of uPortal, successful interop with incubating `uPortal-home` might be more important than any of that.  Let's get these products a little closer to working with one another.

-----

[uportal-dev@ thread]: https://groups.google.com/a/apereo.org/d/topic/uportal-dev/egT3gFFB2ZM/discussion


apetro:
`./gradlew install` works for me locally with Tim's 1405900 .

PR #904, opened 2017-05-25
UP-4830: Decompose sources in uportal-war into Gradle-based submodules (episode 5)

drewwills:
https://issues.jasig.org/browse/UP-4830

##### Checklist

-   [x] the [individual contributor license agreement][] is signed
-   [x] commit message follows [commit guidelines][]

##### Description of change

Fifth installment of the move to Gradle.

[individual contributor license agreement]: https://github.com/Jasig/uPortal/blob/master/CONTRIBUTING.md#individual-contributor-license-agreement
[commit guidelines]: https://github.com/Jasig/uPortal/blob/master/CONTRIBUTING.md#commit
[message properties]: https://github.com/Jasig/uPortal/tree/master/uportal-war/src/main/resources/properties/i18n
[WCAG 2.0 AA]: https://www.w3.org/WAI/WCAG20/quickref/?levels=aaa&technologies=smil%2Cpdf%2Cflash%2Csl

PR #901, opened 2017-05-23
UP-4830:  Decompose sources in uportal-war into Gradle-based submodules (chapter 4)

drewwills:
https://issues.jasig.org/browse/UP-4830

##### Checklist

-   [x] the [individual contributor license agreement][] is signed
-   [x] commit message follows [commit guidelines][]

##### Description of change

Fourth installment of UP-4830;  this work is nearing completion.

[individual contributor license agreement]: https://github.com/Jasig/uPortal/blob/master/CONTRIBUTING.md#individual-contributor-license-agreement
[commit guidelines]: https://github.com/Jasig/uPortal/blob/master/CONTRIBUTING.md#commit
[message properties]: https://github.com/Jasig/uPortal/tree/master/uportal-war/src/main/resources/properties/i18n
[WCAG 2.0 AA]: https://www.w3.org/WAI/WCAG20/quickref/?levels=aaa&technologies=smil%2Cpdf%2Cflash%2Csl

PR #897, opened 2017-05-20
WIP UP-4830:  Decompose sources in uportal-war into Gradle-based submodules (installment 3)

drewwills:
https://issues.jasig.org/browse/UP-4830

##### Checklist

-   [x] the [individual contributor license agreement][] is signed
-   [x] commit message follows [commit guidelines][]

##### Description of change

Another installment of UP-4830;  I hope to finish this week.

Previous: #886 #860

[individual contributor license agreement]: https://github.com/Jasig/uPortal/blob/master/CONTRIBUTING.md#individual-contributor-license-agreement
[commit guidelines]: https://github.com/Jasig/uPortal/blob/master/CONTRIBUTING.md#commit
[message properties]: https://github.com/Jasig/uPortal/tree/master/uportal-war/src/main/resources/properties/i18n
[WCAG 2.0 AA]: https://www.w3.org/WAI/WCAG20/quickref/?levels=aaa&technologies=smil%2Cpdf%2Cflash%2Csl

PR #860, opened 2017-04-18
[WiP] UP-4830: Decompose sources in uportal-war into Gradle-based submodules

drewwills:
https://issues.jasig.org/browse/UP-4830

##### Checklist

-   [x] the [individual contributor license agreement][] is signed
-   [x] commit message follows [commit guidelines][]

##### Description of change

See [plan for uP5](https://wiki.jasig.org/display/UPC/uPortal+5+Project+Overhaul)

[individual contributor license agreement]: https://github.com/Jasig/uPortal/blob/master/CONTRIBUTING.md#individual-contributor-license-agreement
[commit guidelines]: https://github.com/Jasig/uPortal/blob/master/CONTRIBUTING.md#commit
[message properties]: https://github.com/Jasig/uPortal/tree/master/uportal-war/src/main/resources/properties/i18n
[WCAG 2.0 AA]: https://www.w3.org/WAI/WCAG20/quickref/?levels=aaa&technologies=smil%2Cpdf%2Cflash%2Csl

PR #906, opened 2017-05-31
UP-4830:  Decompose sources in uportal-war into Gradle-based submodules (final elements)

drewwills:
https://issues.jasig.org/browse/UP-4830

### Checklist

-   [x] tests are included

### Description of change

All sources in `uportal-war/src/main/java` have been previously moved (and merged).  This PR is for finishing the task defined in the JIRA.

#### Remaining items for JIRA ticket complete

- [x] Unit tests "decomposed" into the modules  that contain the classes they are designed to test (as much as feasible)
- [x] Add a Gradle module that produces a uPortal war artifact
- [x] Add a maven overlay that adds `maven-pluto-plugin` and (possibly) some filtering to the Gradle-built uPortal war
- [x] Update the Ant-based deploy system to deploy the overlay uPortal war instead of `uportal-war` 
- [x] Troubleshoot the Gradle-built uPortal war to the point where the app context starts
- [x] Troubleshoot the Gradle-built uPortal war to the point where it renders a page (instead of the uPortal error page) 
- [x] Troubleshoot the Gradle-built uPortal war to the point where there aren't alarming stack traces in the logs
- [x] Add LESS compilation and resource aggregation to the Gradle-built uPortal war (with this change, portal should look & function normally)
- [x] Review Maven plugins in use within `uportal-war` to make sure we're not leaving behind anything valuable
- [x] Integrate the Gradle-built uPortal war with administrative shell tools (instead of `uportal-war`)
- [x] Remove `uportal-war`

Not all the tests, however, are moved in this delta.  Some of the existing unit tests are written in a way that makes it difficult to separate them from the complete portal configuration & runtime classpath.


ChristianMurphy:
Previous #860 #886 #897 #901 #904

drewwills:
@apetro I know I moved that file (and the same test is passing locally).  I suspect this issue stems from a shortsighted definition in `.gitignore`.

ChristianMurphy:
These are the (possibly over) specific custom that were brought into the new gitignore.

``` gitignore
!uportal-portlets-overlay/CalendarPortlet/3rd-party/repository/org/osaf/caldav4j/0.7/*.jar
!uportal-portlets-overlay/CalendarPortlet/3rd-party/repository/slide/jakarta-slide-webdavlib/2.2pre1-httpclient-3.0/*.jar
!uportal-war/src/test/resources/org/apereo/portal/io/xml/import_archive.*
!uportal-war/src/test/resources/org/apereo/portal/utils/DirScannerData.zip
```
I suspect these two could be the culprits.

``` gitignore
!uportal-war/src/test/resources/org/apereo/portal/io/xml/import_archive.*
!uportal-war/src/test/resources/org/apereo/portal/utils/DirScannerData.zip
```

drewwills:
IMHO we probably don't need a rule in `.gitignore` that ignores all `.zip` files in the entire project.

Probably no files in `**/src/**` should be ignored.

ChristianMurphy:
> IMHO we probably don't need a rule in `.gitignore` that ignores all `.zip` files in the entire project

`zip`, `jar`, `war`, and `ear` files are artifacts. We really shouldn't being checking any artifacts into the source control management.

drewwills:
> zip, jar, war, and ear files are artifacts.
> We really shouldn't being checking any artifacts into the source control management.
> If we need to bring one its okay, but it should be the exception, not the rule.

It already is the exception.  No one is suggesting that we start adding more of these files or change our opinion of when/where they should be added to the repo in any way.

The current configuration in `.gitignore` is error-prone.  We don't add these files under `{module.name}/src/` except if we intend to add them.  We don't use any processes or human practices that would cause unwanted files to be in that location.

drewwills:
@apetro & @ChristianMurphy -- I added the parts I'm working on now (even though they are WIP) b/c @ChristianMurphy may be able to help me with LESS compilation, and he will need this stuff.

ChristianMurphy:
> No one is suggesting that we start adding more of these files or change our opinion of when/where they should be added to the repo in any way.

:+1: 

> We don't add these files under `{module.name}/src/` except if we intend to add them. We don't use any processes or human practices that would cause unwanted files to be in that location.

I'm less worried about human practices, and more worries about human mistakes.
Artifacts can get accidentally get moved to a `src` folder, I would prefer for that file to be only checked in if there is also an explicit exception added to the `.gitignore`.

ChristianMurphy:
For less work see https://github.com/drewwills/uPortal/pull/3

.gitignore
```diff
@@ -419,5 +419,8 @@ modules.xml
 !uportal-portlets-overlay/CalendarPortlet/3rd-party/repository/slide/jakarta-slide-webdavlib/2.2pre1-httpclient-3.0/*.jar
 !**/src/**/target/
 !**/src/**/tags/
-!uportal-war/src/test/resources/org/apereo/portal/io/xml/import_archive.*
-!uportal-war/src/test/resources/org/apereo/portal/utils/DirScannerData.zip
+!**/src/test/**/*.jar
+!**/src/test/**/*.tar
+!**/src/test/**/*.tar.gz
+!**/src/test/**/*.tgz
+!**/src/test/**/*.zip
```
ChristianMurphy: I'm still not fond of the approach of including artifacts if they happen to be under a folder `src` and `test`.
At least there is still some restriction of where they can be added.

drewwills:
@ChristianMurphy, @bjagg, @apetro, @asgrant, @cousquer, @bpowell, et al...

This PR is ready to merge.  I will do so shortly (after lunch) barring new information.

PR #909, opened 2017-06-01
UP-4857: Upgrade gradle wrapper to Gradle 4

ChristianMurphy:
##### Checklist

##### Description of change

<https://issues.jasig.org/browse/UP-4857>

:arrow_up: Brings better logging :scroll:  and caching :racehorse: , as well as some new APIs to leverage :new:.
Removes Java 5 and 6 support and public access to some internal APIs.

Release Notes <https://docs.gradle.org/4.0/release-notes.html>

:notebook: Gradle wrappers 3.5 and 4.0 use `sh` instead of `bash`


ChristianMurphy:
One new feature the could help in the near term would be native lazily evaluated properties in the build <https://docs.gradle.org/4.0-rc-1/release-notes.html#public-type-for-representing-lazily-evaluated-properties>

Could be used as we port tasks from Ant to Gradle (e.g. #889).

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Sat May 20 14:19:53 MST 2017
+#Thu Jun 01 10:45:08 MST 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.4-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.0-rc-1-bin.zip
```
apetro: Concerned about tradeoffs in depending upon RC-level rather than GA-release-level dependencies. Ought uPortal to be waiting for gradle 4.0 to go GA?

ChristianMurphy: Mild concerns. I anticipate an RC 2 before GA, so GA is still probably a month out. I don't anticipate any new breaking changes or major new features to be added before then.

My primary concerns are that:
* uPortal 5 could be cut without Gradle 4 and the possibility of being trapped on Gradle 3 until the next major uPortal release.
* During the transition to Gradle some custom tasks are being created that could be handled by new tasks bundled in Gradle 4. Those will take time and energy to rework.

ChristianMurphy: Marking PR as blocked for now.

ChristianMurphy: > If no regressions are reported, a final [Gradle 4] release will follow later this week.

https://github.com/gradle/gradle/releases/tag/v4.0.0-RC3

ChristianMurphy: Gradle 4 has a general release! :star2:

drewwills: It's great that there's a GA release of Gradle 4.  FWIW, I don't feel that updating the version of the Gradle Wrapper -- even a major version -- would in any way violate (on it's own... all else being equal...) the principals of Semantic Versioning.

We would (I believe) be able to make an update like that w/o breaking any APIs, configs, database schemas, etc.

drewwills:
I tested this PR locally.  I'm not having any issues, even with the move to `sh`.

:+1:

PR #913, opened 2017-06-02
Resolves marketplace entry feed almost always having a null url

vertein:
##### Checklist

##### Description of change

The marketplace entry feed would almost always return an object with a null url (would return valid url if altmaxurl was set).  This creates a consistent url in the feed.

Doesn't resolve that three constructors are using more or less the same code.


Doug-Reed:
:+1:

apetro:
So the consequence of this changeset is that `MarketplaceEntry`s always have an `alternativeMaximizedLink` JavaBean property value? Which makes it less "alternative". Maybe "bestAvailableMaximizedLink" would better capture the spirit.

Naming things is hard. Thanks for getting this working, that's a huge win. Big :+1: from me.

drewwills:
What about simply `maximizedUrl` or `effectiveMaximizedUrl`?

PR #920, opened 2017-06-06
UP-4861: Adds the Error Prone compiler checks to build

ChristianMurphy:
##### Checklist

-   [x] tests are included

##### Description of change

<https://issues.jasig.org/browse/UP-4861>

Add Error Prone compiler checks to the JavaCompilation process.
http://errorprone.info
Static analysis focusing purely on possible errors similar to FindBugs.
Error Prone focuses on compile time source checks, while FindBugs focuses on ByteCode analysis.
Meaning it detects a different class of problems than FindBugs does.


uPortal-layout/uPortal-layout-core/src/main/java/org/apereo/portal/layout/dao/jpa/StylesheetDescriptorImpl.java
```diff
@@ -304,7 +304,7 @@ public ILayoutAttributeDescriptor getLayoutAttributeDescriptor(String name) {
         }
 
         //Remove all old data entries that were not updated
-        dataMap.keySet().remove(oldDataKeys);
+        dataMap.keySet().removeAll(oldDataKeys);
```
ChristianMurphy: `dataMap.keySet()` and `oldDataKeys` are both`Set<String>`.
Remove would try find a `Set<String>` in a `String` which will never be found.
`removeAll` removes the sub`Set` from the `Set`.

uPortal-marketplace/src/main/java/org/apereo/portal/portlet/marketplace/MarketplacePortletDefinition.java
```diff
@@ -741,10 +741,9 @@ public boolean equals(Object other) {
             return false;
         }
         final MarketplacePortletDefinition otherDefinition = (MarketplacePortletDefinition) other;
-        if (getFName() == otherDefinition.getFName()) {
+        if (getFName() == null && otherDefinition.getFName() == null) {
```
ChristianMurphy: explicitly show that it is checking if both values are null.

uPortal-rendering/src/main/java/org/apereo/portal/url/UrlSyntaxProviderImpl.java
```diff
@@ -431,6 +431,7 @@ public IPortalRequestInfo getPortalRequestInfo(HttpServletRequest request) {
                                 break;
                             }
                         }
+                        break;
```
ChristianMurphy: It is unclear if these breaks were accidentally or intentionally missed. These may need to be removed and replaced by a `@SuppressWarnings("FallThrough")`.

ChristianMurphy:
Travis CI failure due to an exception in Findbugs rather than a compilation or test failure.

ChristianMurphy:
closing in favor of #936

PR #922, opened 2017-06-15
UP-4865:  Once favorited, the Favorites portlet continues to list por…

drewwills:
…tlets even when a user no longer has permission to view them

https://issues.jasig.org/browse/UP-4865

##### Checklist

##### Description of change

Favorite Apps will remain in the list even when the user loses access to an app. Clicking on a favorite that the user lost access to will refresh the page and display the following message at the top of the page: “It looks like the page you were attempting to go to does not exist so we redirected you here.” It would be better if apps to which the user no longer has access are not displayed in the list.

### DIFF
--- a/uPortal-portlets/src/main/java/org/apereo/portal/portlets/favorites/FavoritesController.java
+++ b/uPortal-portlets/src/main/java/org/apereo/portal/portlets/favorites/FavoritesController.java
@@ -106,8 +106,11 @@ public class FavoritesController extends AbstractFavoritesController {
         final List<IUserLayoutNodeDescription> rawFavorites = FavoritesUtils.getFavoritePortlets(userLayout);
 
         /*
-         * Filter the collection by SUBSCRIBE permission.  (NOTE:  I wish this filtering were done
-         * by the Layout Manager, before now, but it isn't.)
+         * Filter the collection by SUBSCRIBE permission.
+         *
+         * NOTE:  In the "regular" (non-Favorites) layout, this permissions check is handled by
+         * the rendering engine.  It will refuse to spawn a worker for a portlet to which you
+         * cannot SUBSCRIBE.
          */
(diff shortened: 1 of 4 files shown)


uPortal-portlets/src/main/java/org/apereo/portal/portlets/favorites/FavoritesController.java
```diff
@@ -87,11 +99,32 @@ public String initializeView(Model model) {
...
+        final List<IUserLayoutNodeDescription> collections =
                 FavoritesUtils.getFavoriteCollections(userLayout);
         model.addAttribute("collections", collections);
 
-        List<IUserLayoutNodeDescription> favorites = FavoritesUtils.getFavoritePortlets(userLayout);
+        final List<IUserLayoutNodeDescription> rawFavorites = FavoritesUtils.getFavoritePortlets(userLayout);
+
+        /*
+         * Filter the collection by SUBSCRIBE permission.  (NOTE:  I wish this filtering were done
+         * by the Layout Manager, before now, but it isn't.)
```
ChristianMurphy: I feel like I'm missing something here, if this belongs in the layout manager, why not add it to the layout manager?

drewwills: A couple of reasons (or maybe more)...

- Effort;  the `FavoritesController` appears to be closer to the information necessary to make these calculations
- Unintended consequences;  the Layout Manager (and the layout from it) is used extensively throughout the portal;  the "old school" (Simple) Layout Manager is actually the underpinnings of the DLM ("Distributed Layout Manager");  a change like that, if make too quickly, could have an adverse effect on fragments (and probably a few other things I can't immediately think of)

ChristianMurphy: Makes sense, could some of that reasoning be added to the code comments?

PR #929, opened 2017-06-19
NOJIRA: Can now publish to local maven

bpowell:
##### Checklist

##### Description of change
For some reason the ability to publish to maven local was removed. I added it back

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -124,6 +124,13 @@ subprojects {
     artifacts {
         archives javadocJar, sourcesJar
     }
+    publishing {
+        publications {
+            maven(MavenPublication) {
+                from components.java
+            }
+        }
+    }
     uploadArchives {
         repositories {
             mavenDeployer {
--- a/build.gradle
+++ b/build.gradle
@@ -67,6 +67,7 @@ subprojects {
     apply plugin: 'java'
     apply plugin: 'maven'
     apply plugin: 'signing'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -67,6 +67,7 @@ subprojects {
     apply plugin: 'java'
     apply plugin: 'maven'
     apply plugin: 'signing'
+    apply plugin: 'maven-publish'
```
drewwills: What does this plugin do?

I thought it was (only) for the previous (non-working) approach at publishing to sonatype... so when the approach was changed, I removed it.

bpowell: The plugin published artifacts to the local m2 repo on the machine. For some reason it no longer works. If you go back to commit 4564e65 and do a `./gradlew clean assemble publishToMavenLocal` it will push all artifacts to the local m2. The task `publishToMavenLocal` is brought in by `maven-publish` plugin. I would hold off on merging this until we can figure out why artifacts are not being published with this task. I thought this would fix it, but further testing shows it does not.

bpowell:
@drewwills it now works correctly in that doing a `./gradlew clean assemble publishToMavenLocal` it will take all artifacts and push them to the local m2 repo.

drewwills:
@bpowell et al. -- I believe you can publish artifacts to local m2 with `./gradlew install`.

PR #939, opened 2017-07-12
UP-4885: Ensure supported Java version is used to compile portal

ChristianMurphy:
##### Checklist

##### Description of change

<https://issues.jasig.org/browse/UP-4885>

Previously `sourceCompatility` set in `gradle.properties` was creating a dynamic property, which cannot be picked up by the `JavaCompile` task, so the version check was being ignored.

Same goes for `targetCompatibility`, however `targetCompatibility` can be removed since it defaults to `sourceCompatility`.

`gradleVersion` is competing against version set in `.gradle/gradle-wrapper.properties`

:white_check_mark: Example warning output:

```log
* What went wrong:
Execution failed for task ':uPortal-core:compileJava'.
> Could not target platform: 'Java SE 8' using tool chain: 'JDK 7 (1.7)'.
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,6 @@ description = "Apereo uPortal $version"
 apply plugin: 'com.gradle.build-scan'
 apply plugin: 'com.github.kt3k.coveralls'
 apply plugin: me.champeau.gradle.buildscans.RecipesPlugin
-apply plugin: 'java'
 apply plugin: 'build-announcements'
 apply plugin: 'org.standardout.versioneye'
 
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -58,9 +57,12 @@ versioneye {
 }
 
 allprojects {
+    apply plugin: 'java'
     apply plugin: 'jacoco'
     apply plugin: 'nebula.lint'
 
+    sourceCompatibility = 1.8
```
drewwills: This item looks like something that we can move into `gradle.properties` (and that would be a better fit there).

ChristianMurphy: Ya I don't like this being here.
But what I learned working on this is:
```groovy
sourceCompatibility = 1.8 // Does nothing
apply plugin: 'java'
```
Or
```groovy
sourceCompatibility = 1.8 // Does nothing
apply plugin: 'java'
sourceCompatibility = sourceCompatibility // Still does nothing
```
While
```groovy
apply plugin: 'java'
sourceCompatibility = 1.8 // Checks the Java version
```

ChristianMurphy: Also `gradle.properties` are loaded before plugins are applied.

PR #942, opened 2017-07-13
UP-4887:  Overhaul of data model and supporting code for portlet life…

drewwills:
…cycle in uP5

https://issues.jasig.org/browse/UP-4887

##### Checklist

-   [x] tests are included

##### Description of change

uPortal has Portlet Lifecycle features to support workflow and "delegated administration" use cases WRT portal content, but there are a number of issues with them...
  - MAINTENANCE mode is tracked with a parameter (instead of a column), and has a very surprising approach to evaluating whether a portlet is in maintenance (viz. the presence of the parameter means a portlet is in maintenance, even if the value is false)
  - The current approach -- based on columns of UP_PORTLET_DEF -- is woefully in-extensible (you can't change columns every time you want to add or remove a lifecycle state)

For uPortal 5, we should do the painful work of overhauling this system.

Instead of storing all lifecycle info in the portlet's (only) row in UP_PORTLET_DEF, we should move lifecycle info to its own table and establish a one-to-many relationship between the 2 tables.  This approach will allow us to...
  - Store all lifecycle info in the same table
  - Extend the system with new lifecycle states in the future, if needed

This plan has impacts on Import/Export;  we need to
  - Represent all lifecycle info in the <lifebcycle> element
  - Provide a new portlet-definition XSD that supports allows that
  - Provide an "upgrader" that supports backwards-compatibility with portlet entity files that reference the existing XSD


drewwills:
I'll be adding some unit tests shortly.

uPortal-content/uPortal-content-portlet/src/main/java/org/apereo/portal/portlet/dao/jpa/PortletDefinitionImpl.java
```diff
@@ -714,18 +665,59 @@ public EntityIdentifier getEntityIdentifier() {
...
-            return PortletLifecycleState.MAINTENANCE;
-        } else if (expirationDate != null && expirationDate.before(now)) {
-            return PortletLifecycleState.EXPIRED;
-        } else if (publishDate != null && publishDate.before(now)) {
-            return PortletLifecycleState.PUBLISHED;
-        } else if (approvalDate != null && approvalDate.before(now)) {
-            return PortletLifecycleState.APPROVED;
-        } else {
-            return PortletLifecycleState.CREATED;
+        for (IPortletLifecycleEntry entry : lifecycleEntries) {
```
ChristianMurphy: :art:  Maybe use the Java Stream API?

```java
return lifecycleEntries
  .stream()
  .filter(entry -> entry.getDate().before(now))
  .reduce((_, entry) -> entry.getLifecycleState())
  .orElse(null);
```

uPortal-content/uPortal-content-portlet/src/main/java/org/apereo/portal/portlet/dao/jpa/PortletDefinitionImpl.java
```diff
@@ -714,18 +665,59 @@ public EntityIdentifier getEntityIdentifier() {
...
+
+    @Override
+    public void setLifecycleState(PortletLifecycleState lifecycleState, IPerson user) {
+        setLifecycleState(lifecycleState, user, new Date());
+    }
+
+    @Override
+    public void setLifecycleState(PortletLifecycleState lifecycleState, IPerson user, Date timestamp) {
+        // Avoid a ConcurrentModificationException
+        final List<IPortletLifecycleEntry> list = new ArrayList<>(lifecycleEntries);
```
ChristianMurphy: :art: Streams are pretty cool :snowflake: 

``` java
lifecycleEntries = lifecycleEntries
  .stream()
  .filter(entry -> entry.getDate().before(timestamp))
  .collect(Collectors.toList());
```

uPortal-core/src/main/java/org/apereo/portal/portlet/om/PortletLifecycleState.java
```diff
@@ -15,21 +15,34 @@
...
 
+    public static PortletLifecycleState forOrderValue(int orderValue) {
+        PortletLifecycleState rslt = null;
+        for (PortletLifecycleState state : PortletLifecycleState.values()) {
+            if (state.getOrder() == orderValue) {
+                rslt = state;
+                break;
+            }
+        }
+        if (rslt == null) {
```
ChristianMurphy: if `state` is `null` won't an exception be thrown on line 34 `state.getOrder()`?

uPortal-security/uPortal-security-permissions/src/main/java/org/apereo/portal/security/provider/AuthorizationImpl.java
```diff
@@ -279,45 +282,34 @@ public boolean canPrincipalManage(IAuthorizationPrincipal principal, String port
...
+                activity = IPermission.PORTLET_MANAGER_MAINTENANCE_ACTIVITY;
+                break;
         }
 
-        activity = IPermission.PORTLET_MANAGER_ACTIVITY;
-        if ((order <= PortletLifecycleState.PUBLISHED.getOrder()
-                        || portlet.getPublishDate() != null)
-                && doesPrincipalHavePermission(principal, owner, activity, target)) {
-            return true;
+        if (activity == null) {
```
ChristianMurphy: could this be moved the the `default` case of the `switch`?

drewwills:
@apetro, @jimhelwig, @vertein -- Guys, I don't want to be pushy, but...

I'd like to close the loop on this item soon.  UW has an important relationship with this work, so I'd really love to get some feedback from you.

There are several changed files.  If it's not possible to fit in a line-by-line analysis this week, could we maybe discuss the changes at a high level?  This PR takes several columns off the `UP_PORTLET_DEF` table and replaces them with a `UP_PORTLET_LIFECYCLE` table, together with a one-to-many relationship between the two tables.

uPortal-core/src/main/java/org/apereo/portal/portlet/om/IPortletDefinition.java
```diff
@@ -59,8 +60,31 @@
...
+     * The current lifecycle state is equal to the last entry in the lifecycle history that is not
+     * in the future.
+     */
     PortletLifecycleState getLifecycleState();
 
+    /**
+     * Adds the specified state to the lifecycle history at the present moment, removing any
+     * lifecycle changes scheduled for the future.
+     */
+    void setLifecycleState(PortletLifecycleState lifecycleState, IPerson user);
```
apetro: This method is silently destructive. Better to return `boolean` indicating whether scheduled future lifecycle changes were removed?  Would a caller ever care? I guess could add the boolean in the future if eventually someone cares.

uPortal-core/src/main/java/org/apereo/portal/portlet/om/IPortletDefinition.java
```diff
@@ -59,8 +60,31 @@
...
+
+    /**
+     * Schedules the specified state change for the specified moment, removing any lifecycle
+     * changes scheduled beyond that.
+     */
+    void setLifecycleState(PortletLifecycleState lifecycleState, IPerson user, Date timestamp);
```
apetro: This method is silently destructive. Better to return `boolean` indicating whether scheduled future lifecycle changes were removed? Would a caller ever care? I guess could add the boolean in the future if eventually someone cares.

uPortal-content/uPortal-content-portlet/src/main/java/org/apereo/portal/portlet/dao/jpa/PortletDefinitionImpl.java
```diff
@@ -714,18 +665,59 @@ public EntityIdentifier getEntityIdentifier() {
...
+
+    }
+
+    @Override
+    public void setLifecycleState(PortletLifecycleState lifecycleState, IPerson user) {
+        setLifecycleState(lifecycleState, user, new Date());
+    }
+
+    @Override
+    public void setLifecycleState(PortletLifecycleState lifecycleState, IPerson user, Date timestamp) {
```
apetro: Intended concurrency semantics? What's the intention around e.g. multiple threads trying to set lifecycle state concurrently?

apetro:
Proposed a couple JavaDoc copy edits against the branch behind this PR, at https://github.com/drewwills/uPortal/pull/4 .

uPortal-core/src/main/java/org/apereo/portal/portlet/om/IPortletDefinition.java
drewwills: Perhaps change the name to `updateLifecycleState` to avoid (human) confusion with a simple setter.

apetro: Indeed. It's not a simple JavaBean mutator, so avoiding `setX()` naming convention is a good call.

uPortal-content/uPortal-content-portlet/src/main/java/org/apereo/portal/portlet/dao/jpa/PortletDefinitionImpl.java
drewwills: @ChristianMurphy -- my own reading suggests that Streams are a terrific addition to Java 8 and will be a welcome tool in our uPortal developer toolkit, but also that they are not a "slam dunk" in every case.  They can be used in such a way where the performance is several times worse than the equivalent `for` loop (though it's not clear to me we would ever notice the difference, since the difference between -- for example -- 2 millis and 8 millis is only 6 millis).  Also, there are times when the Stream version of a loop is more clear (remarkably more clear) than the equivalent `for` loop, and there are circumstances where the opposite is true.

This is my vote on this subject (and many similar subjects):  no black-and-white rules about Streams.  If it possible to use a Stream, that **does not mean that using a Stream is a requirement** for merging.

This is an area where suggestions are very helpful, but rules are hurtful.

drewwills: That said, I don't mind looking at using a Stream in this particular case.  I simply want to be clear that there are a few things we require -- e.g. WCAG 2.0 Level AA, lack of security vulnerabilities, _etc_. -- and then there are a great many things we value.

uPortal-core/src/main/java/org/apereo/portal/portlet/om/PortletLifecycleState.java
drewwills: BTW, I think this is excellent feedback.  I love having an extra pair of eyes to help catch the "boneheaded" moments.

In this case, we're good.  Certainly -- if `state` were null on line 34 we'd get an NPE.

But we are checking `rslt` for null on this line (#39).  We set `rslt` to null when we defined it on line 32.  We are checking if it is _still_ null, which would indicate that we don't have an instance of `PortletLifecycleState` that matches the specified `orderValue`.

uPortal-security/uPortal-security-permissions/src/main/java/org/apereo/portal/security/provider/AuthorizationImpl.java
drewwills: This is a good change.  Done.

uPortal-content/uPortal-content-portlet/src/main/java/org/apereo/portal/portlet/dao/jpa/PortletDefinitionImpl.java
drewwills: I had to chuckle -- the notion that this method would ever have to worry about multiple threads modifying the `lifecycleEntries` collection at the same time is beyond my imagination.

I guess I should hope this feature (and also uPortal itself) becomes so widely used that this item is a concern?

I can easily `synchronize` the method.  It should not impact anyone negatively.

uPortal-core/src/main/java/org/apereo/portal/portlet/om/PortletLifecycleState.java
ChristianMurphy: Okay, thank makes sense. :+1: 

It still feels a bit round-about to me, the intent is filtering the list, checking if the filtered list is empty, then getting the last element.
With a for loop the order is: filtering and getting the last element the same time, then checking is empty through null check.
Streams can put those operations in an order that gives some more clarity.

uPortal-content/uPortal-content-portlet/src/main/java/org/apereo/portal/portlet/dao/jpa/PortletDefinitionImpl.java
apetro: Some kinds of feedback are best expressed as a Pull Request. I encourage @ChristianMurphy to go ahead and code the refactor to streams, proffer it as a Pull Request, and dare us all not to agree that the streamifying Pull Request is worth merging.

drewwills: That's a decent suggestion.

ChristianMurphy:
Hey @drewwills, it looks like a merge conflict has cropped up. Would you be able to resolve the conflict?

uPortal-content/uPortal-content-portlet/src/main/java/org/apereo/portal/portlet/dao/jpa/PortletDefinitionImpl.java
drewwills: Working on this one now...

IIRC, this suggestion has some good ideas, but not all the ideas are good.  In Hibernate, it's unwise (again -- IIRC) to replace member variables that are collections of other Hibernate-managed entities.  I believe Hibernate wraps some tooling around those collections, and you can run into issues.  If memory serves, it's better to add/remove elements from the collection.

drewwills: I came up with this approach...

```
        /*
         * Lifecycle entries that are scheduled for a
         * date following the new entry must be cleared.
         */
        final Set<IPortletLifecycleEntry> canceledStateChanges =
                lifecycleEntries.stream()
                .filter(entry -> entry.getDate().after(timestamp))
                .collect(Collectors.toSet());
        lifecycleEntries.removeAll(canceledStateChanges);
```

uPortal-content/uPortal-content-portlet/src/main/java/org/apereo/portal/portlet/dao/jpa/PortletDefinitionImpl.java
drewwills: FWIW "Use of '_' as a lambda parameter name is not allowed." _~Idea_  ([further info](https://stackoverflow.com/questions/34521690/why-is-a-single-underscore-character-an-illegal-name-for-a-lambda-parameter))

drewwills: Came up with this approach...

```
        final Date now = new Date();
        final IPortletLifecycleEntry currentEntry = lifecycleEntries.stream()
                .filter(entry -> entry.getDate().before(now)) // Not entries in the future
                .reduce((e1, e2) -> e1.getDate().after(e2.getDate()) ? e1 : e2) // Only the latest
                .orElse(null); // Possible if the portlet is not yet fully created
        return currentEntry != null ? currentEntry.getLifecycleState() : null;
```

drewwills:
@ChristianMurphy I'm taking care of it.

drewwills:
Folks,

I've pushed my updates, but there's a bit of weirdness on export that I'd like to address.

PR #948, opened 2017-07-14
UP-4892:  (rel-4-3-patches) AnyUnblockedGrantPermissionPolicy does not properly leverage the UsernameTaggedCacheEntryPurger...

drewwills:
... to clear permissions evaluations when a user re-authenticates

https://issues.jasig.org/browse/UP-4892

The cache 'org.jasig.portal.security.provider.AnyUnblockedGrantPermissionPolicy.HAS_UNBLOCKED_GRANT' declares the tagTrackingCacheEventListener, but it doesn't do any of the necessary things to benefit from its features.

We need to implement username tracking of cache keys in this cache so that a user's cached permissions evaluations will be leared when s/he re-authenticates.

##### Checklist

##### Description of change

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/security/provider/AnyUnblockedGrantPermissionPolicy.java
+++ b/uportal-war/src/main/java/org/jasig/portal/security/provider/AnyUnblockedGrantPermissionPolicy.java
@@ -38,6 +38,8 @@ import org.jasig.portal.security.IAuthorizationPrincipal;
 import org.jasig.portal.security.IAuthorizationService;
 import org.jasig.portal.security.IPermission;
 import org.jasig.portal.security.IPermissionPolicy;
+import org.jasig.portal.utils.cache.CacheKey;
+import org.jasig.portal.utils.cache.UsernameTaggedCacheEntryPurger;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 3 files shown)


drewwills:
Folks -- this PR began as a bug fix for a client and is for `rel-4-3-patches`.  _Please don't merge_ until I have a chance to organize a version of this fix for `master`.

uportal-war/src/main/java/org/jasig/portal/security/provider/AnyUnblockedGrantPermissionPolicy.java
```diff
@@ -373,86 +367,36 @@ private boolean containsType(final Set<IPermission> permissions, final String so
...
-                    + ", target=" + target + "]";
+    protected CacheKey getCacheKey(IAuthorizationPrincipal principal, IPermissionOwner owner,
+                                   IPermissionActivity activity, IPermissionTarget target) {
+
+        CacheKey rslt;
+        if (principal.isGroup()) {
+            /*
+             * Untagged keys for groups...
+             */
+            rslt = CacheKey.build(
```
ChristianMurphy: :art Using `return` instead of `rlst` could allow this to become a guard expression.

ref: https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html

drewwills: @ChristianMurphy -- The page you linked is badly in need of more explanation.  It's clear what he's suggesting, but the rationale is entirely missing.  (Perhaps he just imagines that once you see his brilliance, no explanation is required?... I don't buy it.)

In the history of Java programming, there have been many luminaries (through the years) who have staunchly advocated a style where methods should have at most one `return` statement (also for the purpose of "clarity").  That advice is starkly contradictory to the advice you linked.  You really can follow only one or the other.

Now then...
  - I don't feel more obligation to follow the "one return per method" folks than the advice you posted;  I didn't bring it up because I'm a "one return per method" zealot, only to show that what you posted is not universally accepted.
  - I don't feel obligated to follow _anyone's_ Java style advice, simply because s/he said it;  I _deeply enjoy_ opportunities for growth and a programmer... but if you want to persuade me of something, you have to make a compelling case.

This link does not make a compelling case.  Do you perhaps have another?  I'm open minded... but also always critical.

ChristianMurphy: I am also familiar with the one `return`, no `continue` or `break` style.
But I know if from C and C++ where that style helps guarantee memory cleanup. [reference 1](https://softwareengineering.stackexchange.com/questions/118703/where-did-the-notion-of-one-return-only-come-from)
With a garbage collector to manage cleanup, multiple returns can reduce logical lines of code and cyclomatic complexity. [reference 2](https://sourcemaking.com/refactoring/replace-nested-conditional-with-guard-clauses) [reference 3](https://blog.codefx.org/techniques/multiple-return-statements/)

drewwills:
I may be able to wrap this one up today... incl. applying it to `master`.

PR #969, opened 2017-07-26
UP-4913:  Access settings defined in security.properties EXCLUSIVELY through the Spring Environment

drewwills:
https://issues.jasig.org/browse/UP-4913

##### Checklist

-   [x] tests are included

##### Description of change

For uP5, we need to be sourcing properties ONLY through the Spring Environment.


uPortal-security/uPortal-security-mvc/src/main/java/org/apereo/portal/security/mvc/LogoutController.java
```diff
@@ -138,113 +108,96 @@ public void service(HttpServletRequest request, HttpServletResponse response)
...
-                                        + " Looking for redirect string for the root context");
-                    }
-                    redirect = this.redirectMap.get("root");
-                    if (redirect != null && !redirect.equals("")) {
-                        return redirect;
-                    }
-                }
+    private String selectRedirectionUrl(HttpServletRequest request) {
+
+        String rslt = null; // default
```
ChristianMurphy: Still not a fan of *one return* style in high level languages like Java.

uPortal-security/uPortal-security-mvc/src/main/java/org/apereo/portal/security/mvc/LogoutController.java
```diff
@@ -43,15 +40,19 @@
...
 @Controller
 @RequestMapping("/Logout")
-public class LogoutController implements InitializingBean {
+public class LogoutController {
+
+    /* package-private */ static final String LOGOUT_REDIRECT_PREFIX = "logoutRedirect.";
+    /* package-private */ static final String LOGOUT_REDIRECT_ROOT = "root";
 
-    private static final Log log = LogFactory.getLog(LogoutController.class);
+    private final Logger logger = LoggerFactory.getLogger(getClass());
```
ChristianMurphy: :art: It would be nice to use [`@Slf4j`](https://projectlombok.org/features/log)

uPortal-security/uPortal-security-mvc/src/main/java/org/apereo/portal/security/mvc/LogoutController.java
```diff
@@ -67,63 +68,32 @@ public void setPortalEventFactory(IPortalAuthEventFactory portalEventFactory) {
...
-            throws ServletException, IOException {
-        String redirect = this.getRedirectionUrl(request);
+    public void doLogout(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+        String redirect = this.selectRedirectionUrl(request);
         final HttpSession session = request.getSession(false);
 
         if (session != null) {
-            // Record that an authenticated user is requesting to log out
+            // Record that an authenticated user is requesting to logger out
```
apetro: Nope.

uPortal-security/uPortal-security-mvc/src/main/java/org/apereo/portal/security/mvc/LogoutController.java
drewwills: Am I missing something?  -- this change is a change to using slf4j.

uPortal-security/uPortal-security-mvc/src/main/java/org/apereo/portal/security/mvc/LogoutController.java
drewwills: lol

drewwills:
@ChristianMurphy,

    What else is planned for this?

It's simple:  anywhere that uPortal code directly loads one of the 3 properties files listed in the JIRA, that needs to change.  Until it does, adopters will not be able to rely on defining their settings "externally." (And they will fall back on overlaying properties files, which is a brittle, risky practice.)

There are a few more places that `security.properties` is loaded directly.

I think `rdbm.properties` is not loaded directly.

I bet `portal.properties` is loaded directly at least once.

uPortal-security/uPortal-security-mvc/src/main/java/org/apereo/portal/security/mvc/LogoutController.java
ChristianMurphy: You are already using `org.slf4j.Logger`.
Lombok handles the magic of handling the imports and injecting a `log` instance for you.
Wherever it is leveraged is guarantees a standard implementation.

drewwills: Ah! -- this is a comment about Lombok, not slf4j.  (I didn't follow your link.)

drewwills:
Folks,

Based on what I learned when further testing & troubleshooting this change, I removed the feature to redirect to different locations based on security context on logout.

AFAIK, the feature...
  - Wasn't working
  - (Probably) hadn't worked in a long time
  - Isn't used
  - Would be difficult to "get right" with support for the PropertySourcesPlaceholderConfigurer

So I removed some complexity from the `LogoutController`.

drewwills:
@folks -- I need to dial this PR back a bit.  I had wanted to eliminate direct reads of...

  - `rdbm.properties`
  - `security.properties`
  - `portal.properties`

But I'm not going to have time to complete all of that immediately.  Here's the status...

  - `rdbm.properties` didn't have any direct reads in the first place;  it's only sourced via the `PropertySourcesPlaceholderConfigurer`
  - `security.properties` had several direct reads;  I eliminated them in this PR
  - `portal.properties` has (iirc) only one direct read:  the `PortalPropertiesManager`;  this thing is used by over a hundred classes, and I can't get through all of that (not now)

Also, the changes for `security.properties` mount to a significant refactor of how `ISecurityContextFactory` instances are configured and created -- they are now 100% Spring-managed beans!  (That should be a welcome change.)  Many, many lines of custom cruft have been removed.

uPortal-webapp/src/main/webapp/WEB-INF/web.xml
```diff
@@ -90,15 +90,15 @@
         <filter-class>org.jasig.cas.client.validation.Cas20ProxyReceivingTicketValidationFilter</filter-class>
         <init-param>
           <param-name>casServerUrlPrefix</param-name>
-          <param-value>${environment.build.cas.protocol}://${environment.build.cas.server}${environment.build.cas.context}</param-value>
+          <param-value>http://localhost:8080/cas</param-value>
```
bpowell: Why are we hardcoding localhost here? If this can be controlled via uPortal-start that is okay, but if not this is really bad. We don't run CAS on the same machine as uPortal (it's also a bad idea to do that).

drewwills: Yeah, this item is tougher than other items.

I included this change because (ATM) this value is _literally_ hard-coded to...

```
${environment.build.cas.protocol}://${environment.build.cas.server}${environment.build.cas.context}
```

There are no more Maven Filters, so the replacement tokens end up in the deployment.  :(  That's why the 'Sign In' link is broken on `master` now.

At the present moment, you can (100%) configure this setting to a local value by overlaying uPortal's `web.xml` file in uPortal-start.  That works, but I hope to do better (just not in this PR.)

The reading I've done suggests that we can (ultimately) avoid overlaying `web.xml` using JVM args (but not environment variable):  https://stackoverflow.com/questions/1404384/referencing-environment-variables-in-web-xml

bpowell: Did we move to gradle too quickly? Because filtering seems like it is the main issue we have been running into lately and gradle does not help us in any way solve this problem. I think it is worth waiting on merging this until we can get the JVM args side of things figured out so it is not a problem.

apetro: `Cas20ProxyReceivingTicketValidationFilter` could be enhanced to accept configuration via environment variables preferentially over its configuration via filter init-params. 

Or using Spring technology put a delegating filter in web.xml that delegates to a Spring bean wiring up the `Cas20ProxyReceivingTicketValidationFilter` instance, with that bean then configured in the preferred way via Spring.

drewwills: I don't believe it's a Gradle issue really -- Gradle can do filtering at least as well as Maven.
It's really about how non-portable builds are a blood-curdlingly serious issue for the long-term health of the project & community.  We really must get away from building uPortal specifically for DEV, TEST, PROD, _etc_.

So we "cut the cord" with filtering when we moved to Gradle (because it seemed a sensible opportunity).  Technically every previous portal runtime feature is still supported because you can "overlay" any file in uPortal to do exactly what you want.  But (of course), we want to discourage that too.

We will refactor every ounce of commonly-customized configuration such that it can be controlled externally.  This change is a work-in-progress.  That's okay -- we're only cutting milestone builds of uP5 at present.  We will "complete" that transformation before we have a GA release.

I changed `web.xml` in this PR becuase -- as it stands -- it's broken 100% of the time.  The change I made (so far) only fixes it in a standard, local dev deployment.

Both of @apetro's suggestions have attractiveness, though I lean toward the 2nd.  Perhaps I can get to that shortly and include it here.

drewwills:
For some context on these changes -- and how you'll be able to configure authentication in uP5 -- please take a look at [this PR in uPortal-start](https://github.com/Jasig/uPortal-start/pull/30).

drewwills:
@bpowell, @apetro, and @ChristianMurphy -- the latest commit (5a6ba3fd677b4d3fd695bc2a319ad4082de71451) fixes CAS configuration in uPortal-start _via_ the primaryPropertyPlaceholderConfigurer.

At this point, this PR includes all the intended functional updates.

uPortal-security/uPortal-security-authn/src/main/java/org/apereo/portal/security/provider/SimpleSecurityContext.java
```diff
@@ -34,9 +34,7 @@
 
     private final int SIMPLESECURITYAUTHTYPE = 0xFF02;
 
-    SimpleSecurityContext() {
-        super();
-    }
+    /* package-private */ SimpleSecurityContext() {}
```
ChristianMurphy: Isn't package private implied by default access level modifier?

drewwills: It is.

It's also arguably the least used, least commonly appropriate access level in Java.  Many practitioners (including myself) believe it was a poorly-chosen default.  (It's too late to change it now.)  Many Java programmers (again incl. me) imagine that the package-private access level is commonly used by mistake, since it is the default.  For this reason, many Java developers add a comment when they use it to indicate that they actually intended to use it.

PR #975, opened 2017-08-05
UP-4918:  Upgrade the person-directory dependency to version 1.8.5 and refactor to support uP5-style configuration better

drewwills:
https://issues.jasig.org/browse/UP-4918

##### Checklist

##### Description of change

Requires repackage (in configs) 'org.jasig' -> 'org.apereo'.

This change is about refactoring how user attributes are configured so we do not need to overlay personDirectoryContext.xml in uPortal-start, which would be difficult and error-prone (even moreso than the uP4 system).


drewwills:
Tomorrow I will enhance the documentation [here](https://jasig.github.io/uPortal/implement/) to include some information about defining your `IPersonAttributeDao` beans.

Once that's in, I'll be looking to merge and to cut a `5.0.0-M3` build soon so that we can configure User Attributes for some uP5 early adopters.

I'll make sure the CI completes.

drewwills:
This work should be complete enough to merge.  I'll be looking to do that soon.

ChristianMurphy:
@drewwills would it make sense to add this documentation to `uPortal-start/docs`?
Since uPortal-start is where this will be configured from now?

drewwills:
@ChristianMurphy -- It's a good question, but my sense is that it's not the right choice... at least not at this time.

That's a "slippery slope."  As an adopter of uPortal 5, all of your implementation tasks would/should be managed through `uPortal-start`.  If we move this item for that reason, we'll soon fall into moving the majority of the manual over to `uPortal-start`.  We only just got the GitHub-based manual underway... I'd hate to introduce so much incremental chaos.

IMHO -- and for the present -- the `uPortal-start` documentation should be limited to...

 * Information on working with `uPortal-start` itself, such as Gradle tasks
 * A single page:  the `README`

If we moved it, moreover, I'm thinking the proper call would be to create the `uPortal-project.github.io` repo and move it there.  One issue with that approach -- which is also an issue with moving the manual to `uPortal-start` -- is that we would no longer be able to submit a single PR with documentation changes together with the code changes to which they apply.

ChristianMurphy:
Interesting, it sounds like this could warrant some wider discussion.
Could you post your thoughts to <https://groups.google.com/a/apereo.org/d/topic/uportal-dev/uS14KJFyJ6A/discussion>?

drewwills:
It could use some wider discussion... ultimately.

I have many thoughts on this topic, but my #1 is "don't rock the boat."

PR #993, opened 2017-09-06
ci: preserve global npm cache

ChristianMurphy:
##### Checklist

##### Description of change

local `node_modules` folder will still need to be rebuilt each build.
npm can save some time by reusing a copy from its cache folder, rather than downloading from network.
similar to how caching `.gradle` allows CI to preserve its local maven instance between builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,4 +16,5 @@ script: ./gradlew -u -i -q -S build jacocoAggregateReport coveralls
 
 cache:
   directories:
+    - $HOME/.npm/
     - $HOME/.gradle/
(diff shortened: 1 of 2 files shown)


ChristianMurphy:
MacOS image is failing due to timeout. :apple: :no_entry: 
Timeout it due to network instability on Travis CI today. :construction: 

> **Investigating** - We’re investigating an increased rate of internal restarts of macOS builds, resulting in longer boot times for both public and private repositories. This has resulted in an increased backlog for macOS builds at travis-ci.org
> *Sep 6, 09:00 UTC*
>
> **Update** - In addition to longer boot times, users are experiencing an increase in errored builds due to log timeouts when running macOS builds. We are investigating networking issues and will update again as soon as we know more. 
> *Sep 6, 10:45 UTC*

PR #995, opened 2017-09-07
Adds portlet preferences to bearer token

bpowell:
##### Checklist

##### Description of change
Gives the bearer token access to the current portlets preferences.

### DIFF
--- a/uPortal-soffit/uPortal-soffit-connector/src/main/java/org/apereo/portal/soffit/AuthorizationHeaderProvider.java
+++ b/uPortal-soffit/uPortal-soffit-connector/src/main/java/org/apereo/portal/soffit/AuthorizationHeaderProvider.java
@@ -72,6 +72,9 @@ public class AuthorizationHeaderProvider extends AbstractHeaderProvider {
         logger.debug(
                 "Found the following user attributes for username='{}':  {}", username, attributes);
 
+        // Preferences
+        final Map<String, String[]> preferences = renderRequest.getPreferences().getMap();
+
         // Groups
         final List<String> groups = new ArrayList<>();
         final IGroupMember groupMember = GroupService.getGroupMember(username, IPerson.class);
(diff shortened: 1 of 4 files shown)


drewwills:
I hate to play devil's advocate (I am normally a pretty positive person)... why is this change needed?

There's a whole separate JWT that contains just the portlet preferences.  Preferences does not logically seem to be a part of the user.  If we need to make this change, to me it suggests we may have the data model completely wrong.

bpowell:
Is that JWT being used at all anywhere? I was not able to find it.

drewwills:
It's listed here (though the link may need updating)...

  - https://github.com/Jasig/uPortal/blob/master/docs/developer/soffits/soffit_data_model.md

It should also appear in the snooper.

bpowell:
Playing around with the `Preferences` JWT, it doesn't actually have any preferences from the portlet in it. We can print out the JWT on the JSP and it contains an empty map of preferences.

bpowell:
@drewwills what do we want to do about this?

drewwills:
@bpowell -- Let's schedule a quick call to walk through it.  (We should connect ahead of 2017/09/25 anyway.)

bpowell:
@drewwills okay sounds good. What day works for you? I can't do this Friday

drewwills:
Tomorrow 1pm your time?

bpowell:
@drewwills works for me

apetro:
Suggestion: @drewwills , all: when there's going to be a 📞  like this, announce it on [`uportal-dev@`](https://groups.google.com/a/apereo.org/forum/#!forum/uportal-dev) ✉️  and hold it in a way that others can join 🌐  . Cf. [The Open Source Way on direct conversations](https://www.theopensourceway.org/book/The_Open_Source_Way-Stuff_everyone_knows_and_forgets_anyway-Take_extra_extra_extra_care_to_have_all_discussions_in_the_open.html#The_Open_Source_Way-Stuff_everyone_knows_and_forgets_anyway-Radically_visible_meetings_at_all_times) 📖  .

Looks like there was a decision here that resulted in closing this PR. Could you post a summary of what path forward was decided after what considerations?

drewwills:
The summary is essentially already in the threaded discussion here:  the `Preferences` object contains (well...) the preferences.

There was some understandable confusion about it, owing to the fact that the `Preferences` object filters out values with a certain prefix... it considers them "internal."

PR #994, opened 2017-09-07
UP-4930:  Add ability for a given Principal to configure a portlet

cbeach47:
##### Checklist

-   [x] the [individual contributor license agreement][] is signed
-   [x] commit message follows [commit guidelines][]
-   [x] tests are included
-   [x] documentation is changed or added
-   [x] [message properties][] have been updated with new phrases
-   [x] view conforms with [WCAG 2.0 AA][]

##### Description of change

Part of the UP-4930 effort - add the ability for a given Principal to configure a portlet in the same way Portlet Administrators can. 

There is a uPortal-start component to this effort:  https://github.com/Jasig/uPortal-start/pull/42

[individual contributor license agreement]: https://github.com/Jasig/uPortal/blob/master/CONTRIBUTING.md#individual-contributor-license-agreement
[commit guidelines]: https://github.com/Jasig/uPortal/blob/master/CONTRIBUTING.md#commit
[message properties]: https://github.com/Jasig/uPortal/tree/master/uportal-war/src/main/resources/properties/i18n
[WCAG 2.0 AA]: https://www.w3.org/WAI/WCAG20/quickref/?levels=aaa&technologies=smil%2Cpdf%2Cflash%2Csl

PR #1006, opened 2017-09-16
UP-4930:  Moved over changes from PR#994 to clean branch

cbeach47:
https://issues.jasig.org/browse/UP-4930

##### Checklist

-   [x] tests are included

##### Description of change

Adds the ability for a given Principal to configure a portlet in the same way Portlet Administrators can. 

Historical related PRs:
https://github.com/Jasig/uPortal-start/pull/42
https://github.com/Jasig/uPortal/pull/994


docs/sysadmin/portlet-admin/Portlet-setup.md
```diff
@@ -0,0 +1,12 @@
+# Portlet Setup
+After a portlet-definition has been imported into uPortal, users in the Portal Adminstrators group can incorporate portlets into the user experience.
+
+To access, click to the _Portlet Administration Launcher_.  Select a portlet and click _Edit_.
```
drewwills: Okay...

Traditionally, uPortal has not provided much (if any) of this sort of documentation.  This is documentation about how to use the UI, and it's a bit of a can of worms:

  - It is brittle;  it's likely to become incorrect as the platform evolves
  - It is a truly massive undertaking, if we start to hold ourselves to the standard of providing it;  we probably don't have the bandwidth to tackle that

We normally aim to make the UI simple & intuitive enough that we don't need this kind of documentation.  (The question of whether we succeed at that is another matter.)  Typically, I would be happier to see investment in making the UI easier to "figure out" than in documentation about what to click.

uPortal-io/uPortal-io-types/src/main/java/org/apereo/portal/io/xml/permission/PermissionSetsDataFunction.java
```diff
@@ -44,6 +44,9 @@
                     + "WHERE NOT (UPP.ACTIVITY = 'SUBSCRIBE' AND UPP.TARGET LIKE '"
                     + IPermission.PORTLET_PREFIX
                     + "%')\n "
+                    + "AND NOT (UPP.ACTIVITY = 'PORTLET_MODE_CONFIG' AND UPP.TARGET LIKE '"
+                    + IPermission.PORTLET_PREFIX
+                    + "%')\n "
```
drewwills: It looks like you added yours correctly;  but this is an opportunity to improve the whole thing, like...

```
private static final String QUERY =
        "SELECT DISTINCT UPP.OWNER, UPET.ENTITY_TYPE_NAME, UPP.PRINCIPAL_KEY, UPP.ACTIVITY, UPP.PRINCIPAL_TYPE\n "
                + "FROM UP_PERMISSION UPP\n "
                + "LEFT JOIN UP_ENTITY_TYPE UPET ON UPP.PRINCIPAL_TYPE = UPET.ENTITY_TYPE_ID\n "
                + "WHERE NOT (UPP.ACTIVITY = '" + IPermssion.PORTLET_SUBSCRIBER_ACTIVITY + "' AND UPP.TARGET LIKE '"
                + IPermission.PORTLET_PREFIX
                + "%')\n "
                + "AND NOT (UPP.ACTIVITY = '" + IPermission.PORTLET_MODE_CONFIG + "' AND UPP.TARGET LIKE '"
                + IPermission.PORTLET_PREFIX
                + "%')\n "
                + "AND NOT (UPP.ACTIVITY = '" + IPermission.PORTLET_BROWSE_ACTIVITY + "' AND UPP.TARGET LIKE '"
                + IPermission.PORTLET_PREFIX
                + "%')\n ";
```

uPortal-portlets/src/main/java/org/apereo/portal/portlets/portletadmin/PortletAdministrationHelper.java
```diff
@@ -115,10 +115,14 @@
...
 
-    public static final String[] PORTLET_SUBSCRIBE_ACTIVITIES = {
+    public static final String[] PORTLET_PRINCIPAL_SUBSCRIBE_ACTIVITIES = {
         IPermission.PORTLET_SUBSCRIBER_ACTIVITY, IPermission.PORTLET_BROWSE_ACTIVITY
     };
 
+    public static final String[] PORTLET_PRINCIPAL_PUBLISH_ACTIVITIES = {
+        IPermission.PORTLET_MODE_CONFIG
+    };
+
```
drewwills: I believe you can tackle this change in a simpler way...

It appears that the `PORTLET_SUBSCRIBE_ACTIVITIES` (before you renamed it) array is only used in this class.  It appears, furthermore, that the role of the array is only about what things to add to a web form and process.

The name is misleading.  It is in no way important that the members of the array are all `SUBSCRIBE` permissions -- it's merely a coincidence.

I would...

  - Make the variable `private`
  - Change the name to something like `PORTLET_PERMISSIONS_ON_FORM`
  - Add your new one to the original array.

uPortal-portlets/src/main/java/org/apereo/portal/portlets/portletadmin/PortletAdministrationHelper.java
```diff
@@ -253,9 +262,11 @@ private PortletDefinitionForm createNewPortletDefinitionForm() {
         JsonEntityBean everyoneBean =
                 new JsonEntityBean(everyoneGroup, groupListHelper.getEntityType(everyoneGroup));
         form.addPrincipal(everyoneBean);
-        for (String activity : PORTLET_SUBSCRIBE_ACTIVITIES) {
+        for (String activity : PORTLET_PRINCIPAL_SUBSCRIBE_ACTIVITIES) {
             form.addPermission(everyoneBean.getTypeAndIdHash() + "_" + activity);
         }
+        // Purposefully did not add the permissions in PORTLET_PRINCIPAL_PUBLISH_ACTIVITIES for 'everyone'
+        // since they are more of an admin-level set of activities.
```
drewwills: Good point and good comment.

uPortal-portlets/src/main/java/org/apereo/portal/portlets/portletadmin/PortletAdministrationHelper.java
```diff
@@ -268,7 +279,9 @@ private PortletDefinitionForm createNewPortletDefinitionForm() {
      */
     public PortletDefinitionForm savePortletRegistration(
             IPerson publisher, PortletDefinitionForm form) throws Exception {
-
+        if (this.logger.isTraceEnabled()) {
+            logger.trace("In savePortletRegistration() - for: " + form.getPortletName());
+        }
```
drewwills: Can you take a moment to update the class to `org.slf4j.Logger`, which doesn't need any of that `isTraceEnabled()` nonsense?

uPortal-security/uPortal-security-permissions/src/main/java/org/apereo/portal/security/provider/AuthorizationImpl.java
```diff
@@ -241,7 +241,15 @@ public boolean canPrincipalConfigure(
...
         final String activity = IPermission.PORTLET_MODE_CONFIG;
-        return doesPrincipalHavePermission(principal, owner, activity, target);
+
+        boolean isAllowed = doesPrincipalHavePermission(principal, owner, activity, target);
+        if (this.log.isTraceEnabled()) {
+            log.trace(
+                    String.format(
+                            "In canPrincipalConfigure() - principal.key=[%s], is allowed?=[%s]",
+                            principal.getKey(), isAllowed));
+        }
```
drewwills: The `org.slf4j.Logger` turns this mess into 1 line -- it handles both `isTraceEnabled` and `String.format` automatically (though the syntax is different).

uPortal-spring/src/main/java/org/apereo/portal/spring/security/evaluator/PortalPermissionEvaluator.java
```diff
@@ -87,6 +91,13 @@ else if (permission instanceof String) {
                             + targetDomainObject.getClass());
         }
 
+        if (this.logger.isTraceEnabled()) {
+            logger.trace(
+                    String.format(
+                            "In hasPermission() - owner=[%s], activity=[%s], targetId=[%s] ",
+                            activity.getOwnerFname(), activity.getActivityFname(), targetId));
+        }
```
drewwills: Simply `logger.trace("In hasPermission() - owner=[{}], activity=[{}], targetId=[{}]", activity.getOwnerFname(), activity.getActivityFname(), targetId);`

uPortal-webapp/src/main/resources/layout/theme/respondr/content.xsl
```diff
@@ -435,7 +435,8 @@
           </xsl:variable>
           <xsl:variable name="permissionChannelId">PORTLET_ID.<xsl:value-of select="@chanID"/></xsl:variable>
           <xsl:variable name="canConfigure">
-            <!-- This option is special in that it evaluates both whether (1) the portlet supports CONFIG mode and (2) this user is allowed to access it. -->
+            <!-- This option is special in that it evaluates:
+             whether the portlet supports CONFIG mode AND if this user is a Portal Admin (or in a principal that has been granted permission) -->
```
drewwills: Is this language somehow clearer?  (I think it's less.)

uPortal-webapp/src/main/webapp/WEB-INF/flows/edit-portlet/portletConfig.jsp
```diff
@@ -464,13 +464,19 @@ PORTLET DEVELOPMENT STANDARDS AND GUIDELINES
...
                                                     </a>
                                                 </td>
                                                 <td class="col-sm-3 text-nowrap">
                                                     <spring:message code="edit.subscribe"/>
                                                     <a href="javascript:void(0)" data-toggle="tooltip" data-placement="top" title="<spring:message code='edit.subscribe.tooltip'/>">
-                                                        <i class="fa fa-info-circle"></i>
+                                                        <i class="fa fa-info-circle" aria-hidden="true"></i> <span class="sr-only"><spring:message code='edit.subscribe.tooltip'/></span>
+                                                     </a>
+                                                </td>
+                                                <td class="col-sm-3 text-nowrap">
```
drewwills: You just added a bonus `col-sm-3` to an existing HTML structure -- is that okay?  (IIRC Bootstrap columns are meant to add up to 12.)

drewwills:
This PR looks pretty good over all.

uPortal-webapp/src/main/resources/layout/theme/respondr/content.xsl
cbeach47: Good point - I'll revert that change.

docs/sysadmin/portlet-admin/Portlet-setup.md
cbeach47: Ok, I'll back off that change.

uPortal-portlets/src/main/java/org/apereo/portal/portlets/portletadmin/PortletAdministrationHelper.java
drewwills: In case you haven't already figured it out, I was full of it in this remark.  What you have is pretty good.

uPortal-webapp/src/main/webapp/WEB-INF/flows/edit-portlet/portletConfig.jsp
drewwills: Check with @ChristianMurphy on this one.

uPortal-portlets/src/main/java/org/apereo/portal/portlets/portletadmin/PortletAdministrationHelper.java
cbeach47: It was a bit brittle.  I'm currently revising it to use an enum with the plan to have both a single group of possible portlet permissions and the ability to differentiate the owners.

uPortal-webapp/src/main/webapp/WEB-INF/flows/edit-portlet/portletConfig.jsp
ChristianMurphy: Bootstrap 3 grid is percentage based, so this is the equivalent of
``` css
@media (min-width: 768px) { /* smaller screen sizes will apply no width rule */
    .col-sm-4 {
        width: 33.33333333%;
    }
}
```

functionally it works.
Semantically, using `col-sm-3` here is a bit wonky.

cbeach47:
@drewwills - Changes as per your commits are now included.  Based on our side conv, I also added the behavior where the 'Configure?' column will be hidden if the Portlet doesn't support the CONFIG mode.  There are some oddities in the test results (some portlets are hidden when I don't think they should be, and the 'Edit Portlet > Controls > configurable' default value doesn't match the supportMode of CONFIG at times (Google Docs)).

I'm working with @bjagg to see if I've just mis-configured / misunderstood something, or if there's a deeper issue.

cbeach47:
@drewwills after discussing with @bjagg and further testing, the additional behavior of hiding the 'Configure?' column does work as expected - the portlet.xml needs to specify a support type of 'config'.  

I did find what looks like a UI oddity and opened up https://issues.jasig.org/browse/UP-4941 .

drewwills:
I did a `portalInit` with a `SNAPSHOT` based on this branch and brought up `portal-logo` in the Portlet Manager...

configure

Why is Configure checked by default?  (It shouldn't be.)

drewwills:
I have this issue diagnosed & fixed locally.  Shortly I'll push a modest change to the PR.

uPortal-portlets/src/main/java/org/apereo/portal/portlets/portletadmin/PortletAdministrationHelper.java
```diff
@@ -115,18 +116,31 @@
...
-        final String activity;
+        private final String owner;
+        private final String activity;
 
         PortletPermissionsOnForm(String owner, String activity) {
             this.owner = owner;
             this.activity = activity;
         }
+
+        public String getOwner() {
```
drewwills: I realize `owner` and `activity` were/are `final`, but they should have getter methods (otherwise they _look_ like a risky approach).

uPortal-portlets/src/main/java/org/apereo/portal/portlets/portletadmin/PortletDefinitionForm.java
```diff
@@ -457,13 +457,45 @@ public void setPortletPreferenceReadOnly(
...
-    public void addPrincipal(JsonEntityBean principal) {
-        this.principals.add(principal);
+    /**
+     * Sets the default collection of permissions for newly-added principals.  They are BROWSE and
+     * SUBSCRIBE.
+     *
+     * @since 5.0
+     */
+    /* package-private */ void initPermissionsForPrincipal(JsonEntityBean principal) {
+        permissions.add(principal.getTypeAndIdHash() + "_" + PortletAdministrationHelper.PortletPermissionsOnForm.BROWSE.getActivity());
```
drewwills: I want to implement this logic in the Java, not the jsp.  (see below.)

uPortal-webapp/src/main/webapp/WEB-INF/flows/edit-portlet/portletConfig.jsp
```diff
@@ -478,52 +473,35 @@ PORTLET DEVELOPMENT STANDARDS AND GUIDELINES
...
-                                                      <c:when test="${portlet.permissions.contains(principal_subscribe_perm) || !portlet.permissions.contains(principal_browse_perm)}">
-                                                        <td class=text-center><input type="checkbox" name="${principal_subscribe_perm}" checked></td>
-                                                      </c:when>
-                                                      <c:otherwise>
-                                                        <td class=text-center><input type="checkbox" name="${principal_subscribe_perm}"></td>
-                                                      </c:otherwise>
-                                                    </c:choose>
-                                                    <c:if test="${supportsConfig}">
-                                                        <c:choose>
-                                                            <c:when test="${portlet.permissions.contains(principal_configure_perm) || !portlet.permissions.contains(principal_browse_perm)}">
```
drewwills: The main issue/bug with the PR before commit #3 was with this line (above).

It's not a good idea to determine `checked` based on `portlet.permissions.contains(principal_configure_perm) || !portlet.permissions.contains(principal_browse_perm)` -- what does the second part of the test have to do with it???

I realize you were just following the pattern from the previous 2 `<td>`s, but those weren't a good idea either.

What the previous 2 `test` expressions were trying to accomplish was:  "if I either have the permission explicitly, or if I don't have any permissions at all, default this checkbox to checked."

It was a hackish way of tricking the jsp into checking both checkboxes for a new principal (since principals with no checkboxes checked are removed with any page submit).  The approach was brittle -- your update would have broken it... the hack (as it was) only works if there are only 2 permissions in the form.

cbeach47: Thanks @drewwills - that is a cleaner approach.

PR #1014, opened 2017-09-28
Added jgroups support for PostgreSQL

bpowell:
##### Checklist

-   [ x] the [individual contributor license agreement][] is signed

##### Description of change
Allows jgroups to be used in a PostgreSQL environment

### DIFF
--- a/uPortal-webapp/src/main/resources/properties/contexts/applicationContext.xml
+++ b/uPortal-webapp/src/main/resources/properties/contexts/applicationContext.xml
@@ -124,11 +124,27 @@
         </constructor-arg>
     </bean>
 
+    <!-- And JGROUPS DDL used by PostgresSQL Server -->
+    <bean id="jgroupsPostgreSQLInitSql" class="java.lang.String">
+        <constructor-arg>
+            <value>
+                CREATE TABLE IF NOT EXISTS JGROUPSPING (
+                    own_addr varchar(200) NOT NULL,
+                    cluster_name varchar(200) NOT NULL,
+                    ping_data BYTEA DEFAULT NULL,
+                    PRIMARY KEY (own_addr, cluster_name)
+                )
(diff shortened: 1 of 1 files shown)


uPortal-webapp/src/main/resources/properties/contexts/applicationContext.xml
```diff
@@ -124,11 +124,27 @@
         </constructor-arg>
     </bean>
 
+    <!-- And JGROUPS DDL used by PostgresSQL Server -->
+    <bean id="jgroupsPostgreSQLInitSql" class="java.lang.String">
+        <constructor-arg>
+            <value>
+                CREATE TABLE IF NOT EXISTS JGROUPSPING (
```
ChristianMurphy: :question: would adding `IF NOT EXISTS` to the default query break a different database?

bjagg: The SQL is specific to Postgres via dialect mapping in `jgroupsInitSqlMap` below.

bjagg:
:+1:

uPortal-webapp/src/main/resources/properties/contexts/applicationContext.xml
ChristianMurphy: Right, could it be added to the default query, to avoid a special condition for Postgres?

bjagg: No. There are databases that do not recognize `IF NOT EXISTS`, unfortunately. I believe Oracle is one.

Also, the Postgres code is custom for the ping_data column type (?).

ChristianMurphy: :white_check_mark:

PR #1019, opened 2017-10-05
refactor: release tag

ChristianMurphy:
##### Checklist

##### Description of change

use plain semantic version tag without any prefixing.
This makes automated version checking simpler, since it matches the semver regex.
https://github.com/sindresorhus/semver-regex/blob/master/index.js#L3

*edit* SemVerTag does encourage using a `v` prefix

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -209,7 +209,7 @@ subprojects {
  * Configuration for the 'gradle-release' plugin (https://github.com/researchgate/gradle-release)
  */
 release {
-    tagTemplate = 'uportal-${version}' // Looks like it should be a GString, but not necessary
+    tagTemplate = 'v${version}' // Looks like it should be a GString, but not necessary
     git {
         requireBranch = ''
     }


drewwills:
What tools are we using that use this regex?  (I don't believe I'm using any.)

Tags without a prefix do not seem to appear in the UI under the "Branch" pulldown in GitHub, either when I browser the "Tags" tab or use the search.  That bothers me -- it's the reason I restored the prefix.

ChristianMurphy:
The primary reason is that it is to spec (or rather to the SemVerTags sub-spec).
There is some disagreement on whether `v` should be prefixed or not, I'd be fine either way.

> This sub-specification SHOULD be used if you use a version control system (Git, Mercurial, SVN, etc) to store your code. Using this system allows automated tools to inspect your package and determine SemVer compliance and released versions.
> 1. When tagging releases in a version control system, the tag for a version MUST be “vX.Y.Z” e.g. “v3.1.0”.
> 2. The first revision that introduces SemVer compliance SHOULD be tagged “semver”. This allows pre-existing projects to assume compliance at any arbitrary point and for automated tools to discover this fact.

http://semver.org/spec/v1.0.0.html#tagging-specification-semvertag

> What tools are we using that use this regex?

VersionEye uses this pattern when calculating upgrade risk.

> Tags without a prefix do not seem to appear in the UI under the "Branch" pulldown in GitHub, either when I browser the "Tags" tab or use the search. That bothers me -- it's the reason I restored the prefix.

again the SemVerTag spec does allow a `v` prefix.
Tags prefixed with `v` do show in the jump to tag feature on GitHub.

github-tags

drewwills:
The `v` prefix should be fine, as long as the tags show up in the GitHub UI.

PR #1020, opened 2017-10-06
UP-4950:  Regression in the Groups Selector stemming from Porta…

drewwills:
…lPreAuthenticatedProcessingFilter

https://issues.jasig.org/browse/UP-4950

##### Checklist

##### Description of change

This change fixes the regression -- specifically `clearSecurityContextPriorToPortalAuthentication=true` (most of the changes are updates to logging) -- but probably undoes the fix for [UP-4613](https://issues.jasig.org/browse/UP-4613).

  - See also https://github.com/Jasig/uPortal/commit/b55488d17550091500d5fcbd9b581484386efb1a

I need to connect with @groybal next week.

### DIFF
--- a/uPortal-spring/src/main/java/org/apereo/portal/spring/security/preauth/PortalPreAuthenticatedProcessingFilter.java
+++ b/uPortal-spring/src/main/java/org/apereo/portal/spring/security/preauth/PortalPreAuthenticatedProcessingFilter.java
@@ -56,14 +56,11 @@ public class PortalPreAuthenticatedProcessingFilter
 
     private static final String SWAPPER_LOG_NAME = "org.jasig.portal.portlets.swapper";
 
-    /**
-     * Log for identity swapper activity.
-     */
+    /** Log for identity swapper activity. */
     private final Logger swapperLog = LoggerFactory.getLogger(SWAPPER_LOG_NAME);
 
     /**
-     * "Regular" log.  This variable covers a commons-logging log of the same name in the
-     * superclass.
+     * "Regular" log. This variable covers a commons-logging log of the same name in the superclass.
      */
     private final Logger logger = LoggerFactory.getLogger(getClass());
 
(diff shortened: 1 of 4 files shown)


uPortal-spring/src/main/java/org/apereo/portal/spring/security/preauth/PortalPreAuthenticatedProcessingFilter.java
```diff
@@ -54,7 +54,15 @@
...
-    private final Log swapperLog = LogFactory.getLog("org.jasig.portal.portlets.swapper");
+    private static final String SWAPPER_LOG_NAME = "org.jasig.portal.portlets.swapper";
+
+    /** Log for identity swapper activity. */
+    private final Logger swapperLog = LoggerFactory.getLogger(SWAPPER_LOG_NAME);
+
+    /**
+     * "Regular" log. This variable covers a commons-logging log of the same name in the superclass.
+     */
+    private final Logger logger = LoggerFactory.getLogger(getClass());
```
apetro: Since `PortalPreAuthenticatedProcessingFilter` is non-final, should this logger be `protected` and so available to potential sub-classes?

drewwills: I usually prefer it when only methods are `protected`, not instance members at all.

drewwills: Also, I'd say that `PortalPreAuthenticatedProcessingFilter` is non-final for the sake of not interfering with Spring proxy magic... not b/c there's any intent to support a human subclassing it.

groybal:
The change from 'false' to 'true' for clearSecurityContextPriorToPortalAuthentication is fine.  The value should have been 'true' to begin with, that was a mistake on my part for UP-4613.  The change will not affect UP-4613, since that flag was only intended to help support a two-step authentication used by a client.

PR #1031, opened 2017-10-16
build(gradle): add license plugin

ChristianMurphy:
##### Checklist

##### Description of change

Adds gradle license plugin, the Apache 2 short license notice, and formats all files in the project.

:notebook: N.B. Java files have been switched from JavaDoc style to SlashStar style to prevent conflicts between the license plugin and the static analysis plugin.


build.gradle
```diff
@@ -15,6 +15,7 @@ plugins {
     id 'com.github.sherter.google-java-format' version '0.6' apply false
     id 'nebula.lint' version '8.3.1' apply false
     id 'net.ltgt.errorprone' version '0.0.11' apply false
+    id 'com.github.hierynomus.license' version '0.14.0' apply false
```
apetro: <Insert standard grumpiness about production-grade products (uPortal) depending on pre-release `0.whatever` versioned stuff.>

ChristianMurphy: :man_shrugging: the [other plugin](https://github.com/jaredsburrows/gradle-license-plugin) I found is still at `0.6.0` with only 1 contributor. 
If you have a more stable plugin that could be used, suggestions are welcome :smile:

ChristianMurphy: Its also worth noting that https://github.com/hierynomus/license-gradle-plugin
is a wrapper around the classes from http://code.mycila.com/license-maven-plugin, which is at a very stable 3.0

apetro: Yup. I do wish projects would heed [SemVer advice about when to 1.0.0](http://semver.org/#how-do-i-know-when-to-release-100).

ChristianMurphy:
There are merge conflicts and build errors blocking this PR.
This out, will open a revised version as time permits.

PR #1032, opened 2017-10-17
UP-4957: CORS Filter

bjagg:
##### Checklist

-   [ ] tests are included

##### Description of change

https://issues.jasig.org/browse/UP-4957

### DIFF
--- a/docs/implement/security.md
+++ b/docs/implement/security.md
@@ -13,6 +13,11 @@ enables cross-origin requests.
 ### CORS Filter Configuration Options
 
 <dl>
+    <dt><code>cors.enabled</code></td>
+    <dd>
+      <p>A flag to control if this filter is enabled or not.
+      <strong>Defaults:</strong> <code>false</code></p>
+    </dd>
     <dt><code>cors.allowed.origins</code></td>
     <dd>
       <p>A list of <a href="http://tools.ietf.org/html/rfc6454">origins</a>
--- a/docs/implement/security.md
+++ b/docs/implement/security.md
@@ -1,0 +1,90 @@
+# Security in uPortal
+
+## Table of Contents
+
+1. [CORS Filter](#cors-Filter)
+
+## CORS Filter
+
+The CORS Filter is an implementation of W3C's CORS 
+(Cross-Origin Resource Sharing) specification, which
+enables cross-origin requests.
+
+### CORS Filter Configuration Options
+
+<table>
+  <tr>
+    <th width="15%">Attribute</th>
+    <th width="85%">Description</th></tr>
+  <tr>
+    <td><code class="attributeName">cors.allowed.origins</code></td>
+    <td>
+      <p>A list of <a href="http://tools.ietf.org/html/rfc6454">origins</a>
+      that are allowed to access the resource. A <code>*</code> can be
+      specified to enable access to resource from any origin. Otherwise, a
+      whitelist of comma separated origins can be provided. Eg: <code>
+      http://www.w3.org, https://www.apache.org</code>.
+      <strong>Defaults:</strong> <code>*</code> (Any origin is allowed to
+      access the resource).</p>
+    </td></tr>
+  <tr>
+    <td><code class="attributeName">cors.allowed.methods</code></td>
+    <td>
+      <p>A comma separated list of HTTP methods that can be used to access the
+      resource, using cross-origin requests. These are the methods which will
+      also be included as part of <code>Access-Control-Allow-Methods</code> 
+      header in pre-flight response. Eg: <code>GET, POST</code>.
+      <strong>Defaults:</strong> <code>GET, POST, HEAD, OPTIONS</code></p>
+    </td></tr>
+  <tr>
+    <td><code class="attributeName">cors.allowed.headers</code></td>
+    <td>
+      <p>A comma separated list of request headers that can be used when
+      making an actual request. These headers will also be returned as part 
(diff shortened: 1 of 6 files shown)


uPortal-webapp/src/main/resources/properties/contexts/securityContext.xml
```diff
@@ -96,6 +99,19 @@
     </bean>
 
     <!--
+     | Security-related filters
+     +-->
+    <bean id="corsFilter" class="org.apereo.portal.security.filter.CorsFilter">
+        <property name="allowedOrigins" value="${cors.allowed.origins:*}" />
+        <property name="allowedHttpMethods" value="${cors.allowed.methods:GET,POST,HEAD,OPTIONS}" />
```
drewwills: Considering the defaults -- If `allowedOrigins=*`, should we permit POST in a CORS request at all?  How about only `GET` and `HEAD`?  These are described as "safe methods" here:  https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html

jgribonvald: In my mind the problem is more on the property allowedOrigins=* all used method should be authorized in any case. The problem with the property allowed origin is that everybody should consider it for his case. And in example for my portal deployment (one instance for several domain use) I will need to set origin to * as the property is not well exploited/evolved and that we can't specify several URL or a URL pattern (we can but it's not compatible with all browsers). The other way is to use jsonp instead of classic ajax request.

drewwills: One of the goals of this effort is to preserve "deep linking" as a feature in the portal.  It should be possible to put a link like 'http://my.school.edu/uPortal/p/announcements' on another web site, and expect the user to arrive at the Announcements portlet.

bjagg: Yeah, we can remove POST and OPTIONS from the defaults.

With the default of '*' for Allowed Origins, deep linking is still preserved. This filter simply allows implementers the option to tighten access if required.

jgribonvald: What about the use of REST api from another site and maybe from mobile app ?

docs/implement/security.md
```diff
@@ -0,0 +1,90 @@
...
+      access the resource).</p>
+    </td></tr>
+  <tr>
+    <td><code class="attributeName">cors.allowed.methods</code></td>
+    <td>
+      <p>A comma separated list of HTTP methods that can be used to access the
+      resource, using cross-origin requests. These are the methods which will
+      also be included as part of <code>Access-Control-Allow-Methods</code> 
+      header in pre-flight response. Eg: <code>GET, POST</code>.
+      <strong>Defaults:</strong> <code>GET, POST, HEAD, OPTIONS</code></p>
```
ChristianMurphy: :scroll: Documented defaults no longer match real defaults.

docs/implement/security.md
```diff
@@ -0,0 +1,90 @@
...
+
+## CORS Filter
+
+The CORS Filter is an implementation of W3C's CORS 
+(Cross-Origin Resource Sharing) specification, which
+enables cross-origin requests.
+
+### CORS Filter Configuration Options
+
+<table>
```
apetro: Consider [`<dl>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) for a more semantic and less presentation-oriented way to model this documentation. May have better accessibility properties in making more of the semantics of the document available to the browser.

docs/implement/security.md
bjagg: Good catch!

docs/implement/security.md
bjagg: Hmm, I agree!

bjagg: Wish we were using AsciiDoc!

drewwills:
I'm planning to pull this one down and kick the tires a bit today, FWIW.

bjagg:
If feasible, I'd like to merge this tonight.

drewwills:
@bjagg -- can we connect a bit and talk about what problem this change solves?  I wasn't tracking any needs that this change addresses.

I'm not saying that it doesn't do something valuable -- or that we don't want it -- I just can't connect it to any of the items we've previously discussed.

bjagg:
https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet documents how CORS is the first part in prevention.

> ... Although it is usually trivial to spoof any header from a browser using JavaScript, it is generally impossible to do so in the victim's browser during a CSRF attack, except via an XSS vulnerability in the site being attacked with CSRF. More importantly for this recommended Same Origin check, a number of HTTP request headers can't be set by JavaScript because they are on the 'forbidden' headers list. Only the browsers themselves can set values for these headers, making them more trustworthy because not even an XSS vulnerability can be used to modify them.
>
> The Source Origin check recommended here relies on three of these protected headers: Origin, Referer, and Host, making it a pretty strong CSRF defense all on its own.

bjagg:
Added ability to disable CORS filter. Will be disabled by default to keep default browser restrictions of "SAME ORIGIN" in place.

bjagg:
My latest commit replaces the new flag with the optional filter decorator I wrote. Also ran goJF successfully.

PR #1033, opened 2017-10-17
UP-4675:  Added "var" to jQuery assignment

allanjackson:
This prevents jQuery from overwriting the global $ variable if it exists. Had to move the assignment because the rest of the function was relying on the global var.
https://issues.jasig.org/browse/UP-4675

##### Checklist

-   [ ] tests are included

##### Description of change

### DIFF
--- a/uPortal-webapp/src/main/webapp/WEB-INF/jsp/BackgroundPreference/viewBackgroundPreference.jsp
+++ b/uPortal-webapp/src/main/webapp/WEB-INF/jsp/BackgroundPreference/viewBackgroundPreference.jsp
@@ -146,12 +146,8 @@
 </div>
 
 <script id="backgroundScript">
-(function(){
-    var BackgroundChanger = function() {
-        if (typeof up !== 'undefined') {
-            var $ = up.jQuery;
-        }
-        
+(function($){
+    var BackgroundChanger = function() {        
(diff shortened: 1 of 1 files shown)


uPortal-webapp/src/main/webapp/WEB-INF/jsp/BackgroundPreference/viewBackgroundPreference.jsp
```diff
@@ -148,6 +148,10 @@
 <script id="backgroundScript">
 (function(){
```
ChristianMurphy: :notebook: N.B. a common pattern within the portal is to use an [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) with `up.jQuery` passed as a param.

E.G.
```js
(function ($) {
   // some business logic here
})(up.jQuery);

drewwills: This is a great pattern.  :+1:

drewwills:
Welcome @allanjackson

allanjackson:
Okay, I filled out the ICLA form. Then I updated the code to pass up.jQuery as a parameter to the function expression.

PR #1034, opened 2017-10-17
UP-4958:  Remove 'Subscribed Fragments' feature

drewwills:
https://issues.jasig.org/browse/UP-4958

##### Checklist

##### Description of change

A feature for opt-in fragments was added to uPortal several years ago.

It's not a bad idea (it's a good idea, FWIW), but it was implemented in a less-than-optimal way.

The solution for this feature that was added touches almost every aspect of the portal, including...
  - Database schema
  - Import/Export
  - UI

We really only needed the ability to give yourself a user attribute (w/in boundaries established by an admin).  If we had that, everything else would fall in place:  PAGS group > DLM audience > presto! opt-in fragment.


apetro:
> We really only needed the ability to give yourself a user attribute

cf. [Grouper features for self-service end-user opt-in to group membership](https://spaces.internet2.edu/display/Grouper/Grouper+new+UI#GroupernewUI-GroupsIcanjoin).

PR #1039, opened 2017-10-23
UP-4957:  The default value of cors.allowed.origins should probably b…

drewwills:
…e blank, meaning no other origins may access uPortal resources with JS;  if you want to enable it, it's simple enough to specify the origin(s) from which you want to access it

https://issues.jasig.org/browse/UP-4957

##### Checklist

##### Description of change

@bjagg 

### DIFF
--- a/uPortal-webapp/src/main/resources/properties/contexts/securityContext.xml
+++ b/uPortal-webapp/src/main/resources/properties/contexts/securityContext.xml
@@ -102,7 +102,7 @@
      | Security-related filters
      +-->
     <bean id="corsFilter" class="org.apereo.portal.security.filter.CorsFilter">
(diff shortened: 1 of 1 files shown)


uPortal-webapp/src/main/resources/properties/contexts/securityContext.xml
```diff
@@ -102,7 +102,7 @@
      | Security-related filters
      +-->
     <bean id="corsFilter" class="org.apereo.portal.security.filter.CorsFilter">
-        <property name="allowedOrigins" value="${cors.allowed.origins:*}" />
+        <property name="allowedOrigins" value="${cors.allowed.origins:}" />
```
bjagg: Yes, this is a better default

PR #1043, opened 2017-10-31
UP-4962:Manipulate origin and method values to make checks case-inse…

bjagg:
…nsitive

##### Checklist

-   [x] the [individual contributor license agreement][] is signed
-   [x] commit message follows [commit guidelines][]
-   [ ] documentation is changed or added

##### Description of change

https://issues.jasig.org/browse/UP-4962

[individual contributor license agreement]: https://github.com/Jasig/uPortal/blob/master/CONTRIBUTING.md#individual-contributor-license-agreement
[commit guidelines]: https://github.com/Jasig/uPortal/blob/master/CONTRIBUTING.md#commit
[message properties]: https://github.com/Jasig/uPortal/tree/master/uportal-war/src/main/resources/properties/i18n
[WCAG 2.0 AA]: https://www.w3.org/WAI/WCAG20/quickref/?levels=aaa&technologies=smil%2Cpdf%2Cflash%2Csl

PR #1045, opened 2017-11-01
Cors post null content type

bpowell:
##### Checklist
-   [ x ] the [individual contributor license agreement][] is signed

##### Description of change
If a 'POST' request comes in with a null content-type, default it to `CORSRequestType.ACTUAL`


bpowell:
@bjagg this fixes the actionURL problem that we have been discussing on the dev list.

ChristianMurphy:
alternative to #1043

bpowell:
@ChristianMurphy not an alternative, this should really be in addition to #1043

bjagg:
:+1:

PR #1053, opened 2017-11-22
ADA patch

kdjohnson:
##### Checklist

-   [x] the [individual contributor license agreement][] is signed
-   [x] view conforms with [WCAG 2.0 AA][]

##### Description of change

Patches problem with the wrong ARIA attributes being applied to elements

[individual contributor license agreement]: https://github.com/Jasig/uPortal/blob/master/CONTRIBUTING.md#individual-contributor-license-agreement
[commit guidelines]: https://github.com/Jasig/uPortal/blob/master/CONTRIBUTING.md#commit
[message properties]: https://github.com/Jasig/uPortal/tree/master/uportal-war/src/main/resources/properties/i18n
[WCAG 2.0 AA]: https://www.w3.org/WAI/WCAG20/quickref/?levels=aaa&technologies=smil%2Cpdf%2Cflash%2Csl

PR #1054, opened 2017-11-22
ADA Patch 

kdjohnson:
##### Checklist

##### Description of change

Patches problem with the wrong ARIA attributes being applied to elements

  #1053 Had a dangling commit so just reopening the pull request. 


kdjohnson:
Regarding the javascript file it is necessary for fixing the incorrect ARIA attributes.

drewwills:
> Regarding the javascript file it is necessary for fixing the incorrect ARIA attributes.
Are you saying there's a substantive (non-whitespace) change in `jstree.js`?

kdjohnson:
Yes.

drewwills:
I only mean "are there code changes -- differences that affect what the code does?"

If there are, what do you think would be the best way to review those changes?

IIRC the `jstree.js` was a 3rd-party open source library.  On the whole, we try not to customize those.

Is there an update to jstree that perhaps addresses our needs?

kdjohnson:
Ah OK, I see what your saying. I'll look into if there is an update to `jstree`. I agree that it would be preferable not to fork this if possible. If not the change that I made was that I removed this [if statement](https://github.com/Jasig/uPortal/pull/1054/files#diff-0facc9ca0d2c5d7038c2693da9532eefL503) as the `aria-multiselectable` attribute that is being added is not a valid ARIA attribute.

drewwills:
Is it the tool that says it isn't valid?  I can find info on it by googling... https://www.w3.org/TR/wai-aria-1.1/#aria-multiselectable

drewwills:
An interesting thread:  https://github.com/GoogleChrome/accessibility-developer-tools/issues/289

kdjohnson:
Yes [aXe](https://www.deque.com/products/axe/). I'm using the Firefox add-on.

drewwills:
My gut inclination is to revert the changes to `jstree.js`, re-add `aria-multiselect="true"` to the jsp, and merge the rest.

kdjohnson:
I'll look into the newer version of `jstree` and see if that hopefully resolves the problem. Maybe @asgrant will have some thoughts on this.

drewwills:
Yes @asgrant -- can you shed some light on this issue?

To me it looks like the flag on `aria-multiselect` is a "false positive."

ChristianMurphy:
#1056 should help address the issue, by removing jstree from the uPortal source entirely.

asgrant:
> Is it the tool that says it isn't valid? I can find info on it by googling... https://www.w3.org/TR/wai-aria-1.1/#aria-multiselectable

Sorry for being late to the party here is the info I have:

@drew I think the problem here is the role being used, main, is not compatible with aria-multiselectable. This can only be used with grid, listbox, tablist, tree

<div id="portalNavigation" class="fl-widget" **role="main" aria-multiselectable="false"** aria-readonly="false" aria-disabled="false" tabindex="0">

I believe this is why we are also seeing aXe have issue with aria-readonly and various other ones too: Here is the list of ones ok to use according to ARIA 1.1 for role, main: https://www.w3.org/TR/wai-aria-1.1/#main

By looking at the jsTree's code It appears v3.3.4 that Christian updated to uses role=tree which should work for aXe: https://www.w3.org/TR/wai-aria-1.1/#tree

kdjohnson:
@ChristianMurphy I've pulled in your changes from #1056.

drewwills:
@asgrant 

> I think the problem here is the role being used, main, is not compatible with aria-multiselectable. This can only be used with grid, listbox, tablist, tree

So when I look at the [Files Changed tab](https://github.com/Jasig/uPortal/pull/1054/files) I see...

role-tablist

It looks like `aria-multiselectable` is allowable here, so I just want to revert this part of the change.

CR96:
@drewwills 

I just took a look at this tablist in the Edit Portlet dialog. It looks like the list contains only a single expandable tab panel (the "Advanced Options" panel). The `aria-multiselectable` attribute doesn't apply as there are not multiple tab panels to select.

Take a look at [this example](http://www.oaa-accessibility.org/examplep/accordian1/). Each tab panel can be expanded independently, and multiple can be open at once - this is when `aria-multiselectable` would be added and set to `true`. In our case, there's only one tab panel in the list, so we don't need to use this attribute.

PR #1078, opened 2017-12-16
UP-4986:  Handle missing activity data in permissions checking more g…

drewwills:
…racefully

https://issues.jasig.org/browse/UP-4986

##### Checklist

##### Description of change

Sometimes a piece of the fundamental Permissions data expected by the code is missing in the database.  It normally happens when a newer version of the uPortal code is run against an existing database, and a required data update was overlooked.  This condition is not great, but doesn't need to be catastrophic;  it should simply mean that no one will (or can) have the new permission.  This method returns false.

Administrators, however, have permission to do anything, including this unknown activity.  It's most common in uPortal for only Administrators to have access to exotic activities, so in most cases this omission is a wash.

We need to log a WARNing, but this method is invoked a lot, and we don't want to do it incessantly.

### DIFF
--- a/uPortal-security/uPortal-security-permissions/src/main/java/org/apereo/portal/security/provider/AuthorizationImpl.java
+++ b/uPortal-security/uPortal-security-permissions/src/main/java/org/apereo/portal/security/provider/AuthorizationImpl.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 import net.sf.ehcache.Ehcache;
 import net.sf.ehcache.Element;
 import net.sf.ehcache.constructs.blocking.CacheEntryFactory;
@@ -80,10 +81,7 @@ public class AuthorizationImpl implements IAuthorizationService {
      * missing data, currently 5 minutes. Without this throttle, the logs would quickly fill with
      * the same message when there is missing data.
      */
-    private static final long MISSING_DATA_LOG_PERIOD_MILLIS =
-            1000L * // Millis in a second
-                    60L * // Seconds in a minute
-                    5L; // Minutes in five minutes
+    private static final long MISSING_DATA_LOG_PERIOD_MILLIS = TimeUnit.MINUTES.toMillis(5L);
 
     /** Instance of log in order to log events. */
     protected final Logger logger = LoggerFactory.getLogger(getClass());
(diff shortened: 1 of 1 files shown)


uPortal-security/uPortal-security-permissions/src/main/java/org/apereo/portal/security/provider/AuthorizationImpl.java
```diff
@@ -73,8 +75,18 @@
 @Service("authorizationService")
 public class AuthorizationImpl implements IAuthorizationService {
 
+    /**
+     * Period during which this service will not complain (in the logs) about the same item of
+     * missing data, currently 5 minutes. Without this throttle, the logs would quickly fill with
+     * the same message when there is missing data.
+     */
+    private static final long MISSING_DATA_LOG_PERIOD_MILLIS =
```
ChristianMurphy: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/TimeUnit.html

```java
TimeUnit.Minutes.toMillis(5L);
```

uPortal-security/uPortal-security-permissions/src/main/java/org/apereo/portal/security/provider/AuthorizationImpl.java
```diff
@@ -571,22 +587,54 @@ public boolean doesPrincipalHavePermission(
...
+             * exotic activities, so in most cases this omission is a wash.
+             *
+             * We need to log a WARNing, but this method is invoked a lot, and we don't want to do
+             * it incessantly.
+             */
+            final Long now = System.currentTimeMillis();
+            final String missingDataTrackerKey = owner + ":" + activity;
+            final Long lastLogMessageTime = missingDataLogTracker.get(missingDataTrackerKey);
+            if (lastLogMessageTime == null
+                    || lastLogMessageTime < now - MISSING_DATA_LOG_PERIOD_MILLIS) {
```
ChristianMurphy: Non-blocker to merging:
I still think that having a clearly defined data validation layer would be better for uPortal long term.

Without a validation layer, the business logic, in places like this, becomes more complex trying to handle the expected data flow, as well as a variety of acceptable unexpected flows.
And gets work-arounds/hacks like the `MISSING_DATA_LOG_PERIOD_MILLIS` conditional, which are easy to add, but will be hard to remove.

uPortal-security/uPortal-security-permissions/src/main/java/org/apereo/portal/security/provider/AuthorizationImpl.java
drewwills: Needs to be...

```
TimeUnit.MINUTES.toMillis(5L);
```

but not a bad suggestion.

PR #1080, opened 2017-12-20
UP-4987:  Make Group search sensitive to case

cbeach47:
##### Checklist

-   [ ] tests are included

##### Description of change

https://issues.jasig.org/browse/UP-4987

Changed the Group search method constants to an enum, and added the ability to handle case sensitive searching.  

Please review, but do not merge yet - this is a WIP, and I need to have a deeper look if the case sensitive / insensitive calls were correctly separated.

### DIFF
--- a/uPortal-api/uPortal-api-internal/src/main/java/org/apereo/portal/layout/dlm/remoting/GroupListHelperImpl.java
+++ b/uPortal-api/uPortal-api-internal/src/main/java/org/apereo/portal/layout/dlm/remoting/GroupListHelperImpl.java
@@ -21,7 +21,6 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
-
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apereo.portal.EntityIdentifier;
@@ -48,9 +47,9 @@ public class GroupListHelperImpl implements IGroupListHelper {
     /*
      * (non-Javadoc)
      * @see org.apereo.portal.layout.dlm.remoting.IGroupListHelper#search(java.lang.String, java.lang.String)
-     * 
-     * External search, thus case insensitive. 
-     * 
+     *
+     * External search, thus case insensitive.
+     *
      */
     @SuppressWarnings("unchecked")
     @Override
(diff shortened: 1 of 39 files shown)


ChristianMurphy:
Looks good so far @cbeach47, do you have timing information to give some context on the performance impact this will have?

uPortal-api/uPortal-api-internal/src/main/java/org/apereo/portal/layout/dlm/remoting/GroupListHelperImpl.java
```diff
@@ -47,6 +47,9 @@
     /*
      * (non-Javadoc)
      * @see org.apereo.portal.layout.dlm.remoting.IGroupListHelper#search(java.lang.String, java.lang.String)
+     *
+     * External search, thus case insensitive.
```
drewwills: What do you mean by "external?"  (Searches based on user input are the ones that need to be insensitive... I imagine that's what you mean.)

cbeach47: Yeah 'external' meant searches based on user input. I'll switch around that wording to be more clear.

cbeach47:
@ChristianMurphy - I haven't set up timing comparisons, especially since the change effectively made the query more simple.

uPortal-groups/uPortal-groups-core/src/main/java/org/apereo/portal/groups/RDBMEntityGroupStore.java
```diff
@@ -86,7 +86,7 @@
     private static String insertMemberSql;
 
     // SQL group search string
-    private static String searchGroupsPartial =
+    private static String searchGroupsPartialCaseInsensitive =
```
drewwills: These look as though they should be both `final` and `IN_ALL_UPPERCASE`.  I see you were following what was there; but since we're touching it, we should improve it in this no-brainer way (barring some issue I can't see).

drewwills:
Looks pretty thorough.  I'm going to kick the tires locally.

drewwills:
I did a bit of micro-benchmarking...

## Before

before

## After

after

The times are almost identical.  I suspect the data set I used wasn't large enough for this change to make a perceptible difference.

drewwills:
@asgrant -- At some point (after we merge it), I'd love to hear if this change made New Relic any happier.

PR #1107, opened 2018-01-09
UP-4988:  Disable JGroups initialize_sql in favor of handling schema …

drewwills:
…generation with dbloader

https://issues.jasig.org/browse/UP-4988

##### Checklist

##### Description of change

This is a fix I think many folks will enjoy...

DbLoader may appear to be an unattractive piece of ancient, custom uPortal tech -- and on a superficial level it is, just a bit -- but "under the hood" it has much more modern and better-maintained features than the JGroups initialize_sql setting (which generates the JGROUPSPING table, used by uPortal).

As of several years ago, DbLoader was refactored to be a thin wrapper around Hibernate;  it leverages Hibernate's features for dialects & JDBC type mapping.  It's actually pretty slick at it's job, and using it to create the JGROUPSPING table in a platform-agnostic way is a lot less messy than doing so with initialize_sql (i.e. the native JGroups feature).

In addition to being more elegant, it enjoys this important advantage:  it participates in schema drop & re-creation within the dataInit task.  Thiis difference means that the table will be "scrubbed" of any orphaned data on each dataInit or portalInit.  (Orphaned data in the table is the source of JGroups-related slowness when starting a new uPortal app context in Tomcat or within the CLI tools.)

### DIFF
--- a/uPortal-webapp/src/main/resources/properties/jgroups.xml
+++ b/uPortal-webapp/src/main/resources/properties/jgroups.xml
@@ -42,13 +42,14 @@
          thread_pool.keep_alive_time="30000"
     />
 
-    <!-- user database for discovery -->
+    <!-- User database for discovery;  initialize_sql to blank prevents
+         table creation, which we handle with Hibernate via dbloader -->
     <JDBC_PING
             connection_url="${hibernate.connection.url}"
             connection_username="${hibernate.connection.username}"
             connection_password="${hibernate.connection.password}"
             connection_driver="${hibernate.connection.driver_class}"
-            initialize_sql="${jgroupsInitSql}"
+            initialize_sql=""
     />
     <MERGE3  min_interval="5000"
              max_interval="30000"
--- a/uPortal-webapp/src/main/resources/properties/contexts/applicationContext.xml
+++ b/uPortal-webapp/src/main/resources/properties/contexts/applicationContext.xml
@@ -92,62 +92,6 @@
     <!-- Resolver that can be used to dynamically look up property values at run-time -->
     <bean id="propertyResolver" factory-bean="primaryPropertyPlaceholderConfigurer" factory-method="getPropertyResolver"/>
 
-    <!-- Need to capture this value in a bean for map key, using el -->
-    <bean id="hibernateDialect" class="java.lang.String">
-        <constructor-arg value="${hibernate.dialect}"/>
-    </bean>
-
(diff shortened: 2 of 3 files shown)


drewwills:
@bjagg -- Are you in a position to try this PR quickly on something non-HSQL?

bjagg:
I will kick the tires shortly. I think this is a great alternative to the built-in JGroups approach.

Will this resolve the issue with left over records?

drewwills:
@bjagg 

Thank you!  :+1: 

> Will this resolve the issue with left over records?

Yes, to a degree.

I don't know what's causing the left-over records (though I speculate it happens when the JVM terminates abruptly or possibly when the host changes network identity with the portal running), and whenever you have these orphaned records you will experience some delay (more delay the more records you have).  BUT after this change, whenever you `portalInit` or `dataInit` you will reset the table, and you won't have any orphaned records any more (at least for a time).

It's not a total fix, but it's a significant mitigation.  (It's very significant if my conjecture about how they happen is accurate... since server deployments will be less likely to do those things.)  IMHO it's probably all the mitigation we need.  At least I propose a "wait & see" approach as far as further measures to deal with this issue.

The next step would be something like writing custom code to remove rows from the table based on some rule -- either all of them (periodically), or based on age (for which we'd probably have to introduce a new column).

I _don't_ support changing the JGroups behavior so that it won't operate at all in local environments.  (I think there's some value lost in that approach, and I think server environments may experience it as well.)  I believe it's a data problem;  let's tackle it at that level.

uPortal-webapp/src/main/resources/properties/jgroups.xml
```diff
@@ -42,13 +42,14 @@
...
-    <!-- user database for discovery -->
+    <!-- User database for discovery;  initialize_sql to blank prevents
+         table creation, which we handle with Hibernate via dbloader -->
     <JDBC_PING
             connection_url="${hibernate.connection.url}"
             connection_username="${hibernate.connection.username}"
             connection_password="${hibernate.connection.password}"
             connection_driver="${hibernate.connection.driver_class}"
-            initialize_sql="${jgroupsInitSql}"
+            initialize_sql=""
```
bjagg: Looks good. `initialize_sql` needs to be set to empty string to skip default action of creating a table with MS SQL Server syntax

ChristianMurphy:
backported to `rel-5-0-patches` as f163a486f011ce981b31abfc8c32f1b4280dafa9

PR #1115, opened 2018-01-15
UP-4992:  Add Swagger for REST API documentation and tooling

drewwills:
https://issues.jasig.org/browse/UP-4992

##### Checklist

##### Description of change

Swagger is the world’s largest framework of API developer tools for the OpenAPI Specification(OAS), enabling development across the entire API lifecycle, from design and documentation, to test and deployment.

swagger


docs/developer/other/API.md
```diff
@@ -0,0 +1,17 @@
...
+## REST APIs
+
+uPortal includes an extensive suite of REST APIs that it uses to access and manipulate portal data
+from within the UI.  Swagger-based documentation and tooling is available for these APIs, but is
+disabled (by default) as a security precaution.
+
+To enable Swagger API documentation, add the following property to `uPortal.properties`:
+
+```properties
+org.apereo.portal.rest.swagger.SwaggerConfiguration.enabled=true
```
ChristianMurphy: Without the `org.apereo.portal.rest.swagger.SwaggerConfiguration.enabled=true` option I am still able to get to

screencapture-localhost-8080-uportal-api-swagger-ui-html-1516036926868

when I visit */uPortal/api/swagger-ui.html*

drewwills: Yep -- I noticed that as well.

I believe I could make that URL entirely inaccessible (return a 404) using Spring Profiles instead of a property, but...

  - (1) We don't have any other configuration in the portal based on Spring Profiles, as it stands;  and
  - (2) It would be harder to toggle, since it would require a JVM argument or an environment variable

I think I can live with an empty page.  Or if you want to experiment, perhaps it can be done with (conditional) Java-based Spring configuration.

Something like...

```
if (enabled) {
    registry.addResourceHandler("swagger-ui.html")
      .addResourceLocations("classpath:/META-INF/resources/");
 
    registry.addResourceHandler("/webjars/**")
      .addResourceLocations("classpath:/META-INF/resources/webjars/");
}
```

drewwills:
Also @ChristianMurphy -- Documentation & tooling for portlet REST APIs could be done w/ Swagger, but (AFAIK) it's a per-webapp setup.

drewwills:
That's excellent @cousquer :+1:

PR #1120, opened 2018-01-23
[WIP] UP-4994:  Enhance DLM to support pluggable content strategies (beyond 'the content is the layout of the fragment owner')

drewwills:
https://issues.jasig.org/browse/UP-4994

##### Checklist

-   [ ] the [individual contributor license agreement][] is signed
-   [ ] tests are included

##### Description of change

DLM has a lot of flexible features, but it's ultimately restrictive as far as how you can define the content of a fragment:  your only choice is content based on the fragment owner's (personal, database-persisted) layout.

We need more options than that.  We need DLM fragment content based on pluggable strategies -- so the content of a fragment can be the ooutput of a custom algorithm.

This enhancement will allow us to produce fragment content based on things like:
  - Metadata matching (match user context info to portlet metadata)
  - Learning analytics data (users like you are interested in the following portlets)

Within the portal JVM, the Java type that encapsulates the nature of a fragment is UserView.  We need to convert UserView into an interface-based abstraction (viz. IUserView) and support pluggable concrete implementations. 


ChristianMurphy:
:notebook: an example use case for a pluggable strategy could be supporting a "recommended" portlet section in a layout. See https://groups.google.com/a/apereo.org/d/msg/uportal-user/1lgvmGAViys/lBE8Nt4MCgAJ for ongoing discussion.

apetro:
Strategically, I wonder about the tradeoffs between "enhance DLM" and "replace DLM" strategies. But progress is progress. I'm fine with this: pluggable strategies in existing DLM will allow more experimentation with what layout management uPortal would more ideally have.

drewwills:
I thought you might feel that way @apetro.

FWIW I agree that "layouts" in uPortal could benefit from a profound transformation.  Tactically, I'd prefer to tackle that transformation as an evolutionary process.

We provide some new options, they grow in popularity, and we allow the legacy support to recede (in ways).

If I had a week to do it, I bet I could "boil" the `ILayoutStore`/DLM code down -- until it was less than half it's current size -- without loosing a feature.

jgribonvald:
Hey, I will have nearly a such needs... I will need to request to an external app api to get an external user services content list (specific to the user), merge it into user portlet list and the content won't be known from our side, but we will need to get the possibility to add these services to favorites (as example).

PR #1125, opened 2018-01-29
Added indexes for up_user and ss_user

bpowell:
##### Checklist
-   [ x] the [individual contributor license agreement][] is signed

##### Description of change
Adds indexes to `SS_USER_PREF_ID` and `USER_NAME`

### DIFF
--- a/uPortal-layout/uPortal-layout-core/src/main/java/org/apereo/portal/layout/dao/jpa/StylesheetUserPreferencesImpl.java
+++ b/uPortal-layout/uPortal-layout-core/src/main/java/org/apereo/portal/layout/dao/jpa/StylesheetUserPreferencesImpl.java
@@ -49,6 +49,7 @@ import org.hibernate.annotations.Cache;
 import org.hibernate.annotations.CacheConcurrencyStrategy;
 import org.hibernate.annotations.Fetch;
 import org.hibernate.annotations.FetchMode;
+import org.hibernate.annotations.Index;
 import org.hibernate.annotations.NaturalId;
 import org.hibernate.annotations.Type;
 
@@ -66,6 +67,7 @@ import org.hibernate.annotations.Type;
 class StylesheetUserPreferencesImpl implements IStylesheetUserPreferences {
     @Id
     @GeneratedValue(generator = "UP_SS_USER_PREF_GEN")
+    @Index(name = "IDX_SS_USER_PREF_ID")
     @Column(name = "SS_USER_PREF_ID")
     private final long id;
 
--- a/uPortal-webapp/src/main/resources/properties/db/tables.xml
+++ b/uPortal-webapp/src/main/resources/properties/db/tables.xml
@@ -52,6 +52,10 @@
         <name>UPU_DFLT_ID_IDX</name>
         <column-ref>USER_DFLT_USR_ID</column-ref>
     </index>
+    <index>
+        <name>UPU_USER_NAME_IDX</name>
+        <column-ref>USER_NAME</column-ref>
+    </index>
     <unique>
         <name>UPU_USER_NAME</name>
         <column-ref>USER_NAME</column-ref>


drewwills:
Looks great! :+1:

bjagg:
Should this be backported?

ChristianMurphy:
I consider the database to be part of uPortal's API, and adding new rows as a feature, which would make this [SemVer Minor](https://semver.org/spec/v2.0.0.html).

However since uPortal has not clearly defined what it's API is, and whether the database is part of it.
This could arguably be SemVer patch, minor, or major. Depending on the interpretation.

drewwills:
Good question @bjagg.

My impression is that this PR only adds a couple indeces (not new columns, not new data).  We need to test whether (after this change) a previously-created database...

  - Gracefully updates with the index declared using JPA annotations;  and
  - Nevertheless "works" with the code (as well as it ever did) without the index from `tables.xml` applied.

I suspect both hypothesis are true.  In which case, it should be harmless to merge.  As @ChristianMurphy hinted at -- the structure of the portal database is not an outward-facing API.  (Even if it were, this change is unlikely to cause issues with it.)

bjagg:
I would say it is not a new functionality

`Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API.`

but rather correcting a performance issue
`A bug fix is defined as an internal change that fixes incorrect behavior.`

asgrant:
Just my 2 cents, this doesn't change the data model of the application. I see this more as a bug fix or a performance enhancement that has no risk on breaking or changing the API.

asgrant:
@bjagg on backporting. This makes calls to up_user roughly 3 times faster in our environment, it's a small difference, although this table is called often so it adds up. So I could see value in backporting for other uPortal implementors. We are running this currently in uPortal 4 in prod and it is working good.

bpowell:
I can submit the same PR to rel-4-3-patches

bpowell:
Submitted a backport to 4.3 https://github.com/Jasig/uPortal/pull/1126

drewwills:
I'm not sure we need to apply it to 4.3.x;  I think the real question was whether we'd apply it to 5.0.x, or if it needed to be a 5.1-only change.

I'm pretty convinced there's no issue applying it to 5.0.x, but I do plan to run it locally this morning.

bpowell:
@drewwills did you want me to close the PR for 4.3 #1126 ? And then did you want me to put this on rel-5.0-patches?

drewwills:
No thanks @bpowell -- it's very easy to `cherry-pick` into `rel-5-0-patches`.

uPortal-layout/uPortal-layout-core/src/main/java/org/apereo/portal/layout/dao/jpa/StylesheetUserPreferencesImpl.java
```diff
@@ -66,6 +67,7 @@
 class StylesheetUserPreferencesImpl implements IStylesheetUserPreferences {
     @Id
     @GeneratedValue(generator = "UP_SS_USER_PREF_GEN")
+    @Index(name = "IDX_SS_USER_PREF_ID")
```
drewwills: I am a bit confused about this one.  The `@Index` annotation is on the `id` member, which is the Hibernate-managed ID _and primary key_ column for the entity -- it definitely has an index already.

The foreign key index on `up_ss_user_pref_lay_attr(ss_user_pref_id)` also appears to be there, before the change...

up_ss_user_pref_lay_attr

drewwills: Here is what this `@Index` annotation appears to do (in my environment)...

up_ss_user_pref-after

It adds a `IDX_SS_USER_PREF_ID` index on the `SS_USER_PREF_ID` column of the `up_ss_user_pref` table.  (It does exactly what it looked like it did.)  Is that the intent?  (It isn't what I understood.)

uPortal-webapp/src/main/resources/properties/db/tables.xml
```diff
@@ -52,6 +52,10 @@
         <name>UPU_DFLT_ID_IDX</name>
         <column-ref>USER_DFLT_USR_ID</column-ref>
     </index>
+    <index>
+        <name>UPU_USER_NAME_IDX</name>
+        <column-ref>USER_NAME</column-ref>
+    </index>
```
drewwills: This one adds a `UPU_USER_NAME_IDX` index on `up_user`...

up_user-after

bpowell:
Hey @drewwills I screwed up on the `ss_` one. I'll be posting a fix for that in a few. @asgrant is looking at the `up_user` one

drewwills:
Thanks @bpowell & @asgrant for keeping up with this item!

asgrant:
@drewwills @bpowell 

up_ss_user_pref_lay_attr.ss_user_pref_id is still missing from uPortal 5 and needed. 
Table "public.up_ss_user_pref_lay_attr", this shows up for me for the description of that table.
Indexes:
    "up_ss_user_pref_lay_attr_pkey" PRIMARY KEY, btree (up_ss_user_pref_lay_attr_id)

-----

I've confirmed creating an index for up_user.user_name is not needed for uPortal 5, however it is required for uPortal 4. I'm wondering if hibernate versions are working differently for PostgreSQL? This is the code we have for production MySAIL using uPortal 4, which is identical to uPortal 5.

tables.xml
`
    <not-null>USER_ID</not-null>
    <primary-key>USER_ID</primary-key>
    <index>
        <name>UPU_DFLT_ID_IDX</name>
        <column-ref>USER_DFLT_USR_ID</column-ref>
    </index>
    <unique>
        <name>UPU_USER_NAME</name>
        <column-ref>USER_NAME</column-ref>
    </unique>
  </table>
`
This is our uPortal 5 db schema for up_user:

Table "public.up_user"
Indexes:
    "up_user_pkey" PRIMARY KEY, btree (user_id)
    **"upu_user_name" UNIQUE CONSTRAINT, btree (user_name)**
    "upu_dflt_id_idx" btree (user_dflt_usr_id)

This is our uPortal 4 db schema for up_user (both test and prod):
Table "public.up_user"
Indexes:
    "up_user_pkey" PRIMARY KEY, btree (user_id)
    "upu_dflt_id_idx" btree (user_dflt_usr_id)

drewwills:
@asgrant et al.,

I hope you know that I'd be very happy to figure this item out and merge a solution.  (I feel like my input is a bit of a "wet blanket" on this PR.)

When I do...

```
select * from INFORMATION_SCHEMA.SYSTEM_INDEXINFO where table_name = 'UP_SS_USER_PREF_LAY_ATTR'
```

I get...

```

PUBLIC | PUBLIC | UP_SS_USER_PREF_LAY_ATTR | false | PUBLIC | SYS_IDX_SYS_PK_21144_21145 | 3 | 1 | UP_SS_USER_PREF_LAY_ATTR_ID | A |   | 0 | [null] |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
PUBLIC | PUBLIC | UP_SS_USER_PREF_LAY_ATTR | true | PUBLIC | SYS_IDX_FK_959W0XA3YUB0FUA4JOY6EVAOG_21459 | 3 | 1 | SS_USER_PREF_ID | A |   | 0 | [null] |  
```

Two rows -- the first is the PK on `UP_SS_USER_PREF_LAY_ATTR_ID`, and the second is a (non-unique) index on `SS_USER_PREF_ID`.

I'm beginning to thing the behavior is dialect- or RDBMS platform-specific.

bjagg:
I just tested this. This creates an index named IDX_SS_USER_PREF_ID on column SS_USER_PREF_ID for table UP_SS_USER_PREF_LAY_ATTR. There was already a generated index on that column for uP5. I will drop tables and test again tomorrow.

PR #1126, opened 2018-01-30
Added indexes for up_user and ss_user UP-4.3

bpowell:
Checklist

[ x] the [individual contributor license agreement][] is signed

Description of change
Adds indexes to SS_USER_PREF_ID and USER_NAME

### DIFF
--- a/uportal-war/src/main/java/org/jasig/portal/layout/dao/jpa/StylesheetUserPreferencesImpl.java
+++ b/uportal-war/src/main/java/org/jasig/portal/layout/dao/jpa/StylesheetUserPreferencesImpl.java
@@ -84,7 +84,6 @@ import com.google.common.collect.ImmutableMap.Builder;
 class StylesheetUserPreferencesImpl implements IStylesheetUserPreferences {
     @Id
     @GeneratedValue(generator = "UP_SS_USER_PREF_GEN")
-    @Index(name = "IDX_SS_USER_PREF_ID")
     @Column(name = "SS_USER_PREF_ID")
     private final long id;
     
@@ -130,6 +129,7 @@ class StylesheetUserPreferencesImpl implements IStylesheetUserPreferences {
     private final Map<String, String> parameters = new LinkedHashMap<String, String>(0);
 
     @OneToMany(targetEntity = LayoutNodeAttributesImpl.class, cascade = CascadeType.ALL, fetch = FetchType.EAGER, orphanRemoval = true)
+    @Index(name = "IDX_SS_USER_PREF_ID")
     @MapKey(name="nodeId")
     @JoinColumn(name = "SS_USER_PREF_ID", nullable=false)
     @Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)
--- a/uportal-war/src/main/java/org/jasig/portal/layout/dao/jpa/StylesheetUserPreferencesImpl.java
+++ b/uportal-war/src/main/java/org/jasig/portal/layout/dao/jpa/StylesheetUserPreferencesImpl.java
@@ -50,6 +50,7 @@ import org.hibernate.annotations.Cache;
 import org.hibernate.annotations.CacheConcurrencyStrategy;
 import org.hibernate.annotations.Fetch;
 import org.hibernate.annotations.FetchMode;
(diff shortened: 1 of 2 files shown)


asgrant:
Copying and pasting from the other PR: 

I've confirmed creating an index for up_user.user_name is not needed for uPortal 5, however it is required for uPortal 4. I'm wondering if hibernate versions are working differently for PostgreSQL? This is the code we have for production MySAIL using uPortal 4, which is identical to uPortal 5.

tables.xml

```
<not-null>USER_ID</not-null>
    <primary-key>USER_ID</primary-key>
    <index>
        <name>UPU_DFLT_ID_IDX</name>
        <column-ref>USER_DFLT_USR_ID</column-ref>
    </index>
    <unique>
        <name>UPU_USER_NAME</name>
        <column-ref>USER_NAME</column-ref>
    </unique>
  </table>
```

This is our uPortal 5 db schema for up_user:

Table "public.up_user"
Indexes:
    "up_user_pkey" PRIMARY KEY, btree (user_id)
    **"upu_user_name" UNIQUE CONSTRAINT, btree (user_name)**
    "upu_dflt_id_idx" btree (user_dflt_usr_id)

This is our uPortal 4 db schema for up_user (both test and prod):
Table "public.up_user"
Indexes:
    "up_user_pkey" PRIMARY KEY, btree (user_id)
    "upu_dflt_id_idx" btree (user_dflt_usr_id)

uportal-war/src/main/java/org/jasig/portal/layout/dao/jpa/StylesheetUserPreferencesImpl.java
```diff
@@ -83,6 +84,7 @@
 class StylesheetUserPreferencesImpl implements IStylesheetUserPreferences {
     @Id
     @GeneratedValue(generator = "UP_SS_USER_PREF_GEN")
+    @Index(name = "IDX_SS_USER_PREF_ID")
     @Column(name = "SS_USER_PREF_ID")
     private final long id;
```
bjagg: In Aaron's email he mentioned `up_ss_user_pref_lay_attr` table. This column is already indexed on this table (`up_ss_user_pref`) and the other. Was there another column that should be indexed?

uportal-war/src/main/resources/properties/db/tables.xml
```diff
@@ -53,6 +53,10 @@
         <name>UPU_DFLT_ID_IDX</name>
         <column-ref>USER_DFLT_USR_ID</column-ref>
     </index>
+    <index>
+        <name>UPU_USER_NAME_IDX</name>
+        <column-ref>USER_NAME</column-ref>
+    </index>
     <unique>
```
bjagg: Yes, this index was needed.

uportal-war/src/main/java/org/jasig/portal/layout/dao/jpa/StylesheetUserPreferencesImpl.java
bpowell: Nope, sorry I messed up :)

uportal-war/src/main/resources/properties/db/tables.xml
drewwills: If we need it, we need it.  I'm on board ;-).

But I am uncertain about this:  my reading suggests that creating a unique constraint on a column always -- on all the database platforms we might care about -- automatically creates an index on that field.  [Here's an example](https://stackoverflow.com/questions/487314/primary-key-or-unique-index#answer-2916464)

asgrant: This is where I'm confused on this piece of code. In uP4 it is acting as a unique constraint and in uP5 it is acting as a unique index at least for PostgreSQL 9. Hibernate's <unique> seems to be executing differently for us depending on the flavor of uPortal.

ChristianMurphy:
Hey @bpowell! :wave: 
Rest assured this PR has not been forgotten. :relieved: 
Many of the uPortal committers are focused on preparation for [Open Apereo 2018](https://www.apereo.org/conferences/open-apereo-2018) right now. :soon: 
More folks should be available to help with reviews after June 6.

PR #1159, opened 2018-03-28
docs: add sequence diagram showing soffit interaction

ChristianMurphy:
##### Checklist

##### Description of change

 
A sequence diagram of how Portlets, Soffit Server, and Soffit Client connect.
Generated with the [mermaid diagram library](https://mermaidjs.github.io/)


jgribonvald:
:1st_place_medal:

prigaux:
> JWT signature is checked with public key

?

AFAIK
- the token is encrypted with PBEWithMD5AndDES
- so it is neither a JWE nor a JWT
- for JWT signature HS512 is used, ie HMAC, ie shared secret.

Cf https://github.com/Jasig/uPortal/blob/master/uPortal-soffit/uPortal-soffit-core/src/main/java/org/apereo/portal/soffit/service/AbstractJwtService.java#L113

Or https://github.com/andrewstuart/soffit-go/blob/master/opts.go#L42 (which decrypts the token, but doesn't verify JWT signature)

ChristianMurphy:
A JWT is still a JWT if it is wrapped and unwrapped with another encryption.
In the same way that HTML is still HTML if it is encrypted and transported over HTTPS.
@drewwills may be able to give more context and insight into the current structure.

> which decrypts the token, but doesn't verify JWT signature

signature verification is strongly recommended, but not required.

PR #1161, opened 2018-03-30
doc(F5): Add documentation for Remote IP Valve

bjagg:
##### Checklist

##### Description of change


drewwills:
Folks,

It might be a good idea to add this `<Valve>` element (but commented-out), together with these instructions, to the server.xml file maintained inside uPortal-start.

What happens when these settings are applied to a tomcat that isn't behind a load balancer?  (e.g. a local dev portal)

PR #1162, opened 2018-04-03
UP-5009:  Provide a simple, secure way of sharing identity and role i…

drewwills:
…nformation to REST APIs based on OpenID Connect

https://issues.jasig.org/browse/UP-5009

##### Checklist

##### Description of change

This feature is for securing REST endpoints -- and sharing some basic user info -- in a post-portlet world.  See included documentation.

### TODO

- [x] Add a `groups` claim (whitelist)

### DIFF
--- a/docs/developer/soffits/securing_apis.md
+++ b/docs/developer/soffits/securing_apis.md
@@ -1,0 +1,94 @@
+# Securing APIs
+
+Soffit projects often provide HTTP-based APIs for supplying data to UI components or performing
+operations on persistent data stores.  In fact, this practice is preferred and recommended over
+_Server-Side Rendering_ strategies that generate dynamic HTML with data already baked-in.  Modules
+that provide separate "front-end" (UI) and "back-end" (server-side) logic are easier to understand,
+maintain, and extend.
+
+In most cases these module-provided APIs follow the principals of <i>Representational State
+Transfer</i> (REST) and use JSON as a data exchange format.  Both of these practices are also
+preferred and recommended.
+
+Often these APIs require some information from a trusted source.  For example, they may need to know
+the identity and/or characteristics of the user, may provide access to sensitive data, or both.
+
+uPortal provides a simple, secure way of sharing this information with APIs.  This capability is
+based on the `SoffitApiPreAuthenticatedProcessingFilter` and [Spring Security][].  The Java class
+files for this feature are available in the `uPortal-soffit-renderer` dependency.
+
+:notebook:  It is not necessary to be a Soffit to use these features.
+
+## Example Implementation
+
+This capability is based on the `SoffitApiPreAuthenticatedProcessingFilter` and [Spring Security][].
+The Java class files for this feature are available in the `uPortal-soffit-renderer` dependency.
+
+The `SoffitApiPreAuthenticatedProcessingFilter` relies on a `Bearer` token in the `Authorization`
+HTTP header.  You can obtain an appropriate token from uPortal's `userinfo` endpoint, which is found
+at the following URI: `/uPortal/api/v5-1/userinfo`.  Any portal user (including unauthenticated
+users) may obtain a personal token from this endpoint.  This token should be obtained with
+JavaScript and passed (also with JavaScript) as a `Bearer` token in the `Authorization` header in
+API requests.
+
+Refer to the sections below to secure API endpoints and access information about the portal user.
+
+### `uPortal-soffit-renderer` Dependency (Gradle Example)
+
+```groovy
+compile "org.jasig.portal:uPortal-soffit-renderer:${uPortalVersion}"
+```
+
+### Spring Security Configuration
+
+This example defines the following rules for accessing API enpoints that begin with `/api`:
+
+  - Authenticated users may send `GET` and `POST` requests
+  - No users may send any other type of request
+
+```java
+@Configuration
+@EnableWebSecurity
+public class SecurityConfiguration extends WebSecurityConfigurerAdapter {
+
+    @Override
+    protected void configure(HttpSecurity http) throws Exception {
+        http
+            .addFilter(preAuthenticatedProcessingFilter())
+            .authorizeRequests()
+                .antMatchers(HttpMethod.GET,"/api/**").authenticated()
+                .antMatchers(HttpMethod.POST,"/api/**").authenticated()
+                .antMatchers(HttpMethod.DELETE,"/api/**").denyAll()
+                .antMatchers(HttpMethod.PUT,"/api/**").denyAll()
+                .anyRequest().permitAll();
+    }
+
+    @Bean
+    public AuthenticationManager authenticationManager() {
+        return new SoffitApiAuthenticationManager();
+    }
+
+    @Bean
+    public AbstractPreAuthenticatedProcessingFilter preAuthenticatedProcessingFilter() {
+        final AbstractPreAuthenticatedProcessingFilter rslt = new SoffitApiPreAuthenticatedProcessingFilter();
+        rslt.setAuthenticationManager(authenticationManager());
+        return rslt;
+    }
+
+}
+```
+
+### Accessing User Information within the API
+
+Information about the user can be obtained from the Spring `SecurityContextHolder`:
+
+```java
+final Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+String username = null;
+if (authentication != null && SoffitApiUserDetails.class.isInstance(authentication.getDetails())) {
+    SoffitApiUserDetails userDetails = (SoffitApiUserDetails) authentication.getDetails();
+    username = userDetails.getUsername();
+}
+```
+
+[Spring Security]: https://projects.spring.io/spring-security/
--- a/docs/developer/soffits/securing_apis.md
+++ b/docs/developer/soffits/securing_apis.md
@@ -13,11 +13,16 @@ preferred and recommended.
 Often these APIs require some information from a trusted source.  For example, they may need to know
 the identity and/or characteristics of the user, may provide access to sensitive data, or both.
 
-uPortal provides a simple, secure way of sharing this information with APIs.  This capability is
-based on the `SoffitApiPreAuthenticatedProcessingFilter` and [Spring Security][].  The Java class
-files for this feature are available in the `uPortal-soffit-renderer` dependency.
+uPortal provides a simple, secure way of sharing this information with module-provided APIs (_i.e._
+APIs within uPortal modules).  This capability is based on the
+`SoffitApiPreAuthenticatedProcessingFilter` and [Spring Security][].  The Java class files for this
+feature are available in the `uPortal-soffit-renderer` dependency.
 
-:notebook:  It is not necessary to be a Soffit to use these features.
+:notebook:  A module need not be a Soffit to use these features, but it really _should_ be a module.
+In other words, APIs that are secured using this approach should have providing data and services
+_within the portal_ as their original and essential purpose.  The reason for this distinction is
+that the information this approach provides is proxy of the user's profile _within the portal_, and
+that may not be (often isn't) the same thing as the user's identity within the organization.
 
 ## Example Implementation
 
@@ -84,11 +89,8 @@ Information about the user can be obtained from the Spring `SecurityContextHolde
 
 ```java
 final Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
-String username = null;
-if (authentication != null && SoffitApiUserDetails.class.isInstance(authentication.getDetails())) {
-    SoffitApiUserDetails userDetails = (SoffitApiUserDetails) authentication.getDetails();
-    username = userDetails.getUsername();
-}
+final String username = (String) authentication.getPrincipal();
+final Collection<? extends GrantedAuthority> authorities = authentication.getAuthorities();
 ```
 
 [Spring Security]: https://projects.spring.io/spring-security/
--- a/uPortal-api/uPortal-api-rest/src/main/java/org/apereo/portal/rest/OidcUserInfoController.java
+++ b/uPortal-api/uPortal-api-rest/src/main/java/org/apereo/portal/rest/OidcUserInfoController.java
@@ -1,0 +1,58 @@
+/**
+ * Licensed to Apereo under one or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information regarding copyright ownership. Apereo
+ * licenses this file to you under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the License at the
+ * following location:
+ *
+ * <p>http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * <p>Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apereo.portal.rest;
+
+import javax.servlet.http.HttpServletRequest;
+import org.apereo.portal.security.IPerson;
+import org.apereo.portal.security.IPersonManager;
+import org.apereo.portal.security.oauth.IdTokenFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RestController;
+
+/**
+ * This controller provides an endpoint through which clients can obtain an ID Token in a format
+ * that is aligned with with OpenID Connect (OIDC). <i>Clients</i> in this case are normally content
+ * objects embedded within the portal page itself (e.g. portlets, soffits, or JS bundles).
+ *
+ * <p>It is critical to point out that this endpoint is not a compliant OIDC Identity Provider. It
+ * does not implement OAuth Authentication Flows of any sort. The user (<code>sub</code> in the ID
+ * Token) is either logged into the portal (ahead of time) or isn't.
+ *
+ * <p>The value of this endpoint is not (therefore) that it brings support for OIDC to uPortal, but
+ * that other modules and services designed to work with uPortal can implement security using
+ * standard OIDC approaches.
+ */
+@RestController
+public class OidcUserInfoController {
+
+    public static final String ENDPOINT_URI = "/v5-1/userinfo";
+    public static final String CONTENT_TYPE = "application/json"; // Should be application/jwt?
+
+    @Autowired private IPersonManager personManager;
+
+    @Autowired private IdTokenFactory idTokenFactory;
+
+    @RequestMapping(
+        value = ENDPOINT_URI,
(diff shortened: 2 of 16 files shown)


drewwills:
@jhelmer-unicon, @chasegawa

drewwills:
@asgrant @bpowell

docs/developer/soffits/securing_apis.md
```diff
@@ -0,0 +1,94 @@
...
+                .antMatchers(HttpMethod.POST,"/api/**").authenticated()
+                .antMatchers(HttpMethod.DELETE,"/api/**").denyAll()
+                .antMatchers(HttpMethod.PUT,"/api/**").denyAll()
+                .anyRequest().permitAll();
+    }
+
+    @Bean
+    public AuthenticationManager authenticationManager() {
+        return new SoffitApiAuthenticationManager();
+    }
```
drewwills: I'm pretty confident this configuration step isn't necessary.  The `SoffitApiAuthenticationManager` doesn't need to be a Spring-managed bean.  We can remove this method -- as well as the line `rslt.setAuthenticationManager(authenticationManager());` (below) -- in favor of calling `setAuthenticationManager(new SoffitApiAuthenticationManager());` in the constructor of  `SoffitApiPreAuthenticatedProcessingFilter `.

uPortal-security/uPortal-security-core/src/main/java/org/apereo/portal/security/oauth/IdTokenFactory.java
```diff
@@ -0,0 +1,265 @@
...
+        list.add(new ClaimMapping("zoneinfo", zoneinfoAttr, DataTypeConverter.STRING));
+        list.add(new ClaimMapping("locale", localeAttr, DataTypeConverter.STRING));
+        list.add(new ClaimMapping("phone_number", phoneNumberAttr, DataTypeConverter.STRING));
+        list.add(
+                new ClaimMapping(
+                        "phone_number_verified",
+                        phoneNumberVerifiedAttr,
+                        DataTypeConverter.BOOLEAN));
+        list.add(new ClaimMapping("updated_at", updatedAtAttributeName, DataTypeConverter.NUMBER));
+
```
drewwills: Does anyone know if it's permitted and/or sane to add claims to the OIDC ID Token that link the user to roles or groups?  @jtgasper3?

uPortal-soffit/uPortal-soffit-renderer/src/main/java/org/apereo/portal/soffit/security/SoffitApiPreAuthenticatedProcessingFilter.java
```diff
@@ -0,0 +1,112 @@
...
+            logger.debug("Found the following pre-authenticated user:  {}", claims.toString());
+
+            final UserDetails rslt = new SoffitApiUserDetails(claims.getBody().getSubject());
+            request.setAttribute(USER_DETAILS_REQUEST_ATTRIBUTE, rslt);
+            return rslt;
+        } catch (Exception e) {
+            logger.error(
+                    "Unable to identify pre-authenticated user with Bearer token '{}'",
+                    bearerToken,
+                    e);
```
drewwills: A stack trace is probably too much to log at `ERROR` level for something as trivial as an expired token.

uPortal-soffit/uPortal-soffit-renderer/src/main/java/org/apereo/portal/soffit/security/SoffitApiPreAuthenticatedProcessingFilter.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        return null;
```
drewwills: Is it typical & appropriate to return `null` when you can't produce a valid Principal?

uPortal-soffit/uPortal-soffit-renderer/src/main/java/org/apereo/portal/soffit/security/SoffitApiAuthenticationManager.java
```diff
@@ -0,0 +1,51 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.security.authentication.AuthenticationManager;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.AuthenticationException;
+
+/**
+ * Concrete <code>AuthenticationManager</code> implementation (Spring Security) for use with {@link
+ * SoffitApiPreAuthenticatedProcessingFilter}. Use <code>
+ * SoffitApiPreAuthenticatedProcessingFilter.setAuthenticationManager</code> in a when constructing
```
drewwills: Remove "in a"

jhelmer-unicon:
I'm a bit conflicted on this.   From a pure technical standpoint, I think a better approach would be to make spring security better/easier in uPortal and then ask implementors to manage this w/ a real OIDC server. 
 I have concerns about putting uPortal in a position where they are a pseudo OIDC IDP (or token provider may be a better term since this doesn't do actual auth)   Since that's not the primary focus of the uPortal community, I worry about their ability to keep up with security threats like a team focused on AIM would.   I'm not worried about the actual implementation.  From what I can tell, it appears to be using solid libraries and it seems to be well implemented.   I'm mostly just concerned that 1. security is hard, 2. it changes often and 3. OIDC is not a primary focus of this project and we may not have the resources to keep up with all the changes that occur in that space.

On the other hand, from a practical perspective, I get it.  This makes standing up new servers much quicker and easier and removes a lot of dependencies.   For developers, this is nice and simple.  It may even be fine for many deployments that don't have access to much sensitive information.   I also appreciate that because it mostly follows OIDC standards, dropping in a real OIDC server should not be terribly difficult.

I guess my general feel is that this sounds great for development, but I would be hesitant about using this in a production environment.  There, I would probably take a long hard look at implementing a real OIDC server.

drewwills:
IMO the feedback from @jhelmer-unicon is insightful, well-articulated, and in line with the conversations I imagined we should have.

There's no question there are pros & cons to this feature.

    I have concerns about putting uPortal in a position where they are a pseudo OIDC IDP (or token provider may be a better term since this doesn't do actual auth) Since that's not the primary focus of the uPortal community, I worry about their ability to keep up with security threats like a team focused on AIM would.

I feel ya.  I have this concern as well.

Nevertheless...
  - To the greatest feasible extent, the "heavy lifting" of the solution has been offloaded to a libraries provided by "a team focused on AIM."
  - The [jjwt dependancy](https://github.com/jwtk/jjwt) is from [Okta](https://www.okta.com/).  This library does all the building/signing/validating/parsing of the token.
  - Consumption of the token and protection of endpoints is provided by Spring Security
  - FWIW, I feel on some level we may be _obligated_ to provide something like this feature.  The portal does not (normally) _own_ identity, user attributes, or groups campus-wide, but it definitely adds value to these concepts for things that produce content _within the portal_.  (In other words, there are attributes & groups inside the portal that don't exist elsewhere in the enterprise.  That's completely normal and 100% "according to plan" -- why would LDAP care about your `User Agent` string?  How would it even manage it?).

    On the other hand, from a practical perspective, I get it. This makes standing up new servers much quicker and easier and removes a lot of dependencies. For developers, this is nice and simple.

We worked our tails off simplifying the adoption of uPortal 5.  "Reluctant" doesn't begin to describe how I feel about telling new adopters that they _must_ implement (and integrate with) a service that provides OAuth & OIDC ID Tokens for the portal to work at all.

This is stuff we had "for free" in the world of JSR-268.  We very much want a new, more contemporary approach to content in the portal.  But one of the most widely discussed drawbacks of portlets is how difficult & arcane they are.  We don't need a version of this feature that is difficult & arcane.

    I also appreciate that because it mostly follows OIDC standards, dropping in a real OIDC server should not be terribly difficult.

*YES!*. Those who want to use another service for ID Tokens should be able to... without difficulty or code changes.

uPortal-security/uPortal-security-core/src/main/java/org/apereo/portal/security/oauth/IdTokenFactory.java
jtgasper3: Per the discussion we had, yes it is permitted (something like `“roles”: ["Admin", … ]`), but I'd recommend some value filtering be in-place to limit values to application associated groups/roles, otherwise token bloat can be an issue. MemberOf can have hundreds of values and will often include the full DN in the value, if being pulled directly from LDAP, which probably provides very little value.

uPortal-soffit/uPortal-soffit-renderer/src/main/java/org/apereo/portal/soffit/security/SoffitApiPreAuthenticatedProcessingFilter.java
jtgasper3: Not sure the entire context here, but http://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthError

jtgasper3:
I'll just quickly weigh in on the OIDC authorization server (IdP/token provider) vs an external service. 

I'm reminded of the Grouper project where they have developed a lot of components from scratch, including their own messaging solution vs just asking folks to deploy Rabbit MQ or ActiveMQ (or other technology). After many years they've come around and now make it easy to integrate with a message broker, which a lot of people have now adopted because they are going the way of messaging. But he team is stuck supporting this other method for legacy purposes. I'm reminded of http://www.astrodigital.org/space/stshorse.html.

It's a tough call. While many organizations don't have them now, it's only a matter of time before they are as common place as SAML/CAS-based SSO. Many SSO solutions now support OAUTH/OIDC, and many more will have support soon. So it's tempting to add in the support to the baseline project, and I'm not saying not to do it, but I'd argue it should be easily swapped out for a formal OIDC authorization server service. The uPortal soffit impls should not be "advantaged" by using the uP OIDC auth server over using an external ODIC auth server implementation. It should just be a matter of changing some config, either way. (Just my opinion though).

uPortal-api/uPortal-api-rest/src/main/java/org/apereo/portal/rest/OidcUserInfoController.java
```diff
@@ -0,0 +1,58 @@
...
+ * <p>The value of this endpoint is not (therefore) that it brings support for OIDC to uPortal, but
+ * that other modules and services designed to work with uPortal can implement security using
+ * standard OIDC approaches.
+ */
+@RestController
+public class OidcUserInfoController {
+
+    public static final String ENDPOINT_URI = "/v5-1/userinfo";
+    public static final String CONTENT_TYPE = "application/json"; // Should be application/jwt?
+
```
prigaux: Yes, make it application/jwt since it always returns a signed JWT:

> If the UserInfo Response is signed and/or encrypted, then the Claims are returned in a JWT and the content-type MUST be application/jwt.

Cf http://openid.net/specs/openid-connect-core-1_0.html#UserInfo

drewwills: Ah yes -- thanks for bringing that point up.

I tried `application/jwt` early on and couldn't get it to work in either Postman or the address bar.  I kept getting HTTP 406 Not Acceptable.  (I even tried setting an Accept header for application/jwt... no luck.)

docs/developer/soffits/securing_apis.md
drewwills: **UPDATE:**  I tried to remove/simplify this step and had no luck at all.  This variant works;  others don't seem to.

bjagg:
I'm new to OIDC (OpenID Connect). That said, I agree with Drew that we should have a solution that is baked in. While most consider a portal as a solution that displays aggregated data in a new way, I see uPortal as a service hub with APIs to access user attributes, groups, permissions, content, layout, etc. Adding an OIDC server would not be much of a stretch of purpose.

The baked-in solution might be leveraging some maintained libraries or maybe an integration with  https://github.com/mitreid-connect/OpenID-Connect-Java-Spring-Server ? In any case, sounds like we are all in agreement that uPortal should be configurable to use an institution's own OIDC services as a replacement.

What would that look like? (Goes back to reading blogs)

jgribonvald:
I'm sharing your point of view, uPortal should provide at least an API tool that permit to provide user informations, authorization + permissions to services, group structure, etc... as a central place to serve portal eco-system services without to re-implement backend service on each service (like portlets thought).
After the user authentication part is an other service and depends on the institution eco-system need.

apetro:
What if we were to

1. Add whatever identity features are needed to CAS, possibly adding optional CAS plugins that fulfill the data for some of these needs from uPortal.
2. Then rely upon, assume these features from CAS.
3. Advise non-CAS adopters to implement an ad-hoc portal-specific CAS instance with their uPortal to fulfill these features.  As in, you're a Shibboleth but not CAS adopter? Great, stand up a portal-specific CAS instance, Shibbolize that, and then rely on that CAS instance from uPortal.

And thereby concentrate the identity/access/authorization/authentication/token/security/stuff in `CAS` rather than in `uPortal`. ?

drewwills:
### Some Incremental Thoughts on The Role of the Portal WRT Access Tokens

@bjagg said:

    While most consider a portal as a solution that displays aggregated data in a new way, I see uPortal as a service hub with APIs to access user attributes, groups, permissions, content, layout, etc. Adding an OIDC server would not be much of a stretch of purpose.

@jgribonvald said:

    I'm sharing your point of view, uPortal should provide at least an API tool that permit to provide user informations, authorization + permissions to services, group structure, etc... as a central place to serve portal eco-system services without to re-implement backend service on each service

I think something very important is being said here, and I agree very much with what I'm hearing.  Let me explain...

I understand completely the line of reasoning that roughly goes:  "uPortal is not a campus identity service;  it is better _not_ to develop/maintain custom code for features outside of uPortal's mission or mandate."  And I agree with that statement 100%, taken out of this context.  Here's the rub:  this feature is entirely within uPortal's mission.

An essential aspect of uPortal's mission is to provide a user experience for navigating the university that is _highly tailored to the individual_ (emphasis added).  How do we provide a "highly tailored" experience?

### Here's How

We have a collection of subsystems in the portal that act in coordination and a strategically-chosen sequence to build this tailored experience:

**Authentication > User Attributes > Groups > Layouts > Content**

When it comes to _Authentication_, _User Attributes_, and _Groups_, uPortal goes to great lengths not to "own" the data, but to integrate with other sources of this information on campus.  That's entirely appropriate & desirable;  but uPortal also adds value to these steps by transforming or enhancing the output of these integrations at every stage.

For example, Authentication provides (strictly-guarded) impersonation features as well as support for attribute release;  User Attributes provides aggregation, merging, and the ability to scrape additional attributes from innovative places like the `HttpServletRequest`;  Groups provides PAGS.

These are value-adds to concepts & data that the portal does not (typically) own.  Importantly -- uPortal doesn't provide these additional bits to other systems on campus... it only provides them to downstream components within itself.

When you consider the issues of whether (1) uPortal should be in the business of providing access tokens to APIs at all, and (2) whether those tokens should contain any portal-specific content, I think it's vitally important to ask yourself "what exactly are these APIs?  Are these APIs logically a part of the portal?"

Ten years ago, we (some of us) were working on a portal framework founded on the assumptions that...
  - Every piece of the portal runs in the same host (though sometimes several copies)
  - Every piece of the portal runs in the same tomcat
  - Every piece of the portal runs in the same Java process

The essence of what we're doing here -- with this PR and several others past, present, and future -- is shattering these assumptions.  We no longer want to build a monolithic portal.  We are working to break the portal into several, more manageable services... but that doesn't mean we want to provide the individual parts of the portal with less than we do now.

I think it matters a lot, when you consider what sort of token an API should receive, whether that API is a part of the portal or not.  I would say that APIs that _aren't_ a part of the portal should not use access tokens provided by the portal and (if for some reason they do) should not rely on portal-specific content within them.  I would say the exact opposite for APIs that are firmly a part of the portal.

Are there any (independently-hosted) APIs that are "firmly a part of the portal?"  I don't know how many there are, but we are building them now.  This week (and last week, and the one before) we are working on new content objects for uPortal that are (1) static Web Components that interact with (2) non-portlet REST APIs that have the ability to run on another host.

The fact that these tokens are (in essence) OIDC ID Tokens is an implementation detail.  They could have been implemented with a completely custom approach.  But in this case we opted to follow a pattern that is vetted, documented, understood by developers, and supported by leading libraries to do the heavy-lifting.

The feedback & participation on this PR has been excellent.  👍 Thanks everyone.

jgribonvald:
Sorry @drewwills I didn't took the time to read all but I trust you, after to respond to @apetro for me it's important to provide such feature into the uportal core as everybody won't use CAS or shibboleth and the portal offered always the possibility to register users into his database, so how to do in this case ? the portal should provide an identity.

After I'm agree on the thing that these API won't be needed by everyone and so aren't mandatory. In my mind the API should be here purpose a replacement of the portlet API, and for me the portal offer many more possibilities than CAS or shibboleth. 
CAS and shibboleth are Identities provider, and could be authorization checkers, but that's all ! In our case we added a delegated authentication part (ie CAS is like a SP to an external authentication service)
in other case when you need more, like PAGS groups, group structure, user's services, permissions, etc... I would prefer to call to uPortal-core instead to develop a custom tool into these apps! until now we never need to use the authorization part as all service already do it, but why not one day but I fear this case about complexe case and implementation that we will need to provide on CAS...

@apetro also if I follow your point in this case if i need to get permissions i will ask to my grouper (internet2 grouper), but it begins to make too much services to request CAS + grouper + uPortal, and what more ?

I don't know what is the best, i'm not experimented in large use of micro services, but all I hear about it is that become a big thing to manage and add a complexity. I prefer to keep or go on simple things !

ChristianMurphy:
I'm not against this being added, but similar to @jhelmer-unicon I'm conflicted.
It seems like we are including our own mini OIDC because people may not want to setup an external OIDC, mainly because of the frustration setup and maintenance of yet another service.
That is a valid point, orchestration can be a pain point.

I feel like including a mini-OIDC is a band-aid though.
The root issue of the problem is with orchestration.
Short term this will work fine.
Long term, we want to make orchestration of education services easier.
Easy to the point that there is no reason to include a mini OIDC in uPortal.

drewwills:
Folks,

    It seems like we are including our own mini OIDC because people may not want to setup an external OIDC, mainly because of frustration around the setup and maintenance of yet another service.

I totally get that what this PR does can appear to be exactly what this sentence says.  But I want to emphasize that it doesn't appear that way to me, at all, any more.

The discussions on this PR have been lively & effective.  They've also covered a lot of territory.  These are the bits that seem important to me...

  1. We want to build new sorts of modules for uPortal that don't have to be in the same host, same Tomcat, same Java process as the portal itself (in the manner that JSR-286 portlets do).
  1. We need a way, however, to communicate with these modules some of the portal-specific info that the Portlet API provides to portlets.
  1. We are fans of the good work that the OAuth and OpenID folks are doing in the identity & access space.
  1. We will tap into some of their good ideas (as we implement this communication) because they are vetted, documented, known to developers, and supported by high-quality libraries.

Note that:
  - This token is not suitable or intended for doing OAuth/OpenID Connect with systems outside the portal.
  - The uPortal modules for which this system exists are (most likely) not suitable/intended for doing OAuth/OpenID Connect with an identity provider like CAS or Shib (at least not directly).

PR #1183, opened 2018-05-08
chore(deps): update spring to version 4.3.17

ChristianMurphy:
##### Checklist

## Release Notes - Spring Framework - Version 4.3.17

**Bug**
    * [SPR-16716] - SpringFailOnTimeout loses original exception when triggering timeout in finally block
    * [SPR-16720] - Inconsistent getTypeForFactoryMethod results for parameterized factory method
    * [SPR-16734] - Generic constructor argument (e.g. ObjectProvider) fails to resolve for inner class
    * [SPR-16739] - AnnotationAwareOrderComparator uses Order instead of Priority for DecoratingProxy
    * [SPR-16743] - Exception swallowed in ResponseEntityExceptionHandler
    * [SPR-16756] - ConfigurationClassBeanDefinitionReader registers duplicate BeanDefinition for nested scoped component
    * [SPR-16762] - Misleading error message when evaluating T operator
    * [SPR-16773] - NPE in SimpleClient-HttpURLConnection with errorstream-buffering
    * [SPR-16783] - Singleton from a FactoryBean may be post-processed twice if the first post-processing triggers a second attempt to get the bean

**Improvement**
    * [SPR-16710] - Proper exception for controller method return types that do not work with MvcUriComponentsBuilder (e.g. final classes)
    * [SPR-16714] - Revise cache safety check to avoid performance regression in EAR packaged applications on WildFly
    * [SPR-16717] - Revise JCA MessageEndpoint exception logging and propagation
    * [SPR-16731] - Flag misguided evaluation attempts in OperatorMatches
    * [SPR-16732] - Expose configuration options for "selector" header 
    * [SPR-16752] - Validate contextPath in RedirectView

##  Release Notes - Spring Framework - Version 4.3.16

**Bug**
    * [SPR-16489] - EL1072E when evaluating compiled null-safe expression
    * [SPR-16695] - ClassCastException in TestDispatcherServlet

## Release Notes - Spring Framework - Version 4.3.15

**Bug**
    * [SPR-15780] - Precondition failed for PUT methods on ResponseEntity return types
    * [SPR-16409] - SubProtocolWebSocketHandler should not log ERROR on "No messages received after ..."
    * [SPR-16414] - ServletServerHttpRequest.getURI() may throw a java.net.URISyntaxException
    * [SPR-16415] - AbstractClientSockJsSession.close call does not propagate IOException from disconnect
    * [SPR-16427] - Spurious WARNINGs when XML declared TransactionProxyFactoryBean's target bean depends on an annotation declared bean that depends on another bean
    * [SPR-16435] - Lambda error detection might not work on JDK 9
    * [SPR-16450] - CachingConnectionFactory - Invalid session in session cache
    * [SPR-16453] - MockMvcRequestBuilder does not decode pathInfo
    * [SPR-16458] - MimeType compareTo implementation is not compatible with equals
    * [SPR-16461] - GSON converter only serialises fields of controller method return type, ignoring subclass fields of response object
    * [SPR-16478] - SimpleJdbcCall can't access synonyms in Oracle database
    * [SPR-16479] - Set thread interrupt flag on InterruptedException
    * [SPR-16486] - JsonMappingException when trying to instantiate org.springframework.messaging.Message
    * [SPR-16488] - Deadlock in SubProtocolWebSocketHandler on shutdown with Undertow
    * [SPR-16495] - NPE in Spring-JDBC with Oracle and SimpleJdbcInsert
    * [SPR-16496] - ExceptionHandlerExceptionResolver advice applicability check may fail against interface-based controller proxy
    * [SPR-16498] - FormTag renders empty <div> tag
    * [SPR-16506] - ForwardedHeaderFilter garbles query params during sendRedirect()
    * [SPR-16526] - StringIndexOutOfBoundsException when rewriting links in CSS resources
    * [SPR-16528] - Spurious ERROR-level logging when using SSEEmitter
    * [SPR-16529] - GsonHttpMessageConverter cannot be used in an SseEmitter because it closes the response stream
    * [SPR-16533] - Race condition in ConcurrentMapCache
    * [SPR-16534] - testBindInstantFromJavaUtilDate fails on systems in the Pacific/Auckland time zone
    * [SPR-16543] - WebApplicationContextFacesELResolver#isReadOnly always return false
    * [SPR-16556] - PostgresTableMetaDataProvider.isGetGeneratedKeysSimulated() does not detect Postgres 10
    * [SPR-16559] - Missing PersistenceException cause message in refresh failure warn log
    * [SPR-16564] - Reading annotations in ConfigurationClassParser does not fall back to ASM on Google App Engine
    * [SPR-16570] - Inconsistent synchronization in AbstractBeanFactoryBasedTargetSource and JdbcAccessor
    * [SPR-16571] - WebAsyncManager concurrentResult should be volatile
    * [SPR-16572] - TransactionTemplate inherits equals()/hashCode() from DefaultTransactionDefinition
    * [SPR-16577] - SimpleAliasRegistry registerAlias not atomic
    * [SPR-16581] - URIEditor should not double escape classpath: URIs
    * [SPR-16582] - RestTemplate with HttpComponentsClientHttpRequestFactory and no buffering with an interceptor throws UnsupportedOperationException
    * [SPR-16584] - Inconsistent handling of null values through Java 8 accessors in ConcurrentReferenceHashMap
    * [SPR-16599] - AcceptHeaderLocaleResolver chooses wrong Locale for language match
    * [SPR-16604] - DefaultResponseErrorHandler wastes the body of a response with an unknown status
    * [SPR-16611] - CallMetaDataContext.reconcileParameters doesn't catch output parameters with DatabaseMetaData.procedureColumnResult type (on Postgres)
    * [SPR-16620] - Consistent thread-safe iteration in DefaultSingletonBeanRegistry
    * [SPR-16625] - FactoryBeanRegistrySupport atomicity issues
    * [SPR-16627] - Address race condition within spring that causes about-to-be-created-bean exceptions
    * [SPR-16630] - An error occurs if a blank character exists before and after the delimiter of the MIME type parameter.
    * [SPR-16640] - Multipart Upload with Commons Fileupload on lazy mode downloads data on cleanup
    * [SPR-16648] - Concurrent result may be missed due to a race condition in MockMvc
    * [SPR-16650] - ServletUriComponentsBuilder should replace context path when X-Forwarded-Prefix is present
    * [SPR-16652] - Annotation lookup on parameter in inner class constructor fails when using javac from JDK versions prior to 9
    * [SPR-16660] - UriComponentsBuilder Forwarded header parsing can throw java.lang.NumberFormatException
    * [SPR-16663] - NamedParameterUtils.parseSqlStatement should parse :{x} style parameter correctly
    * [SPR-16669] - Unable to bind a null value for UUID column with PostgreSQL
    * [SPR-16671] - SimpleMailMessage's handling of to/cc/bcc arrays is inconsistent

**Task**
    * [SPR-7885] - Clarify that RestTemplate registers an AllEncompassingFormHttpMessageConverter by default
    * [SPR-14753] - Spring MVC and AOP: `@EnableAspectJAutoProxy` needs to be redeclared for each application context
    * [SPR-15801] - Doc: Spring AOP dependency on aspectjweaver
    * [SPR-15986] - Doc: ReloadableResourceBundleMessageSource's setUseCodeAsDefaultMessage causes the ValidationMessages.properties to not be found and used correctly
    * [SPR-16428] - Doc: hibernate.dialect_resolvers vs HibernateJpaVendorAdapter.setDatabase
    * [SPR-16463] - Doc: `@Transactional.isolation` does not guarantee the specified isolation level
    * [SPR-16487] - Doc: AbstractMessageListenerContainer rejecting messages on shutdown causes message loss in AUTO_ACK mode
    * [SPR-16552] - Incorrect description for class-level @Transactional with AspectJ
    * [SPR-16592] - Doc: `@PropertySource` meta-annotation not detected next to direct annotation
    * [SPR-16602] - Doc: `@Scope` not inherited from base class
    * [SPR-16616] - Clean duplicate slash and backslash in resource URLs
    * [SPR-16658] - Incorrect Javadocs for ReflectionUtils.doWithLocalFields

**Improvement**
    * [SPR-12452] - Spring Websockets Broker relay supporting a cluster of STOMP endpoint addresses
    * [SPR-16439] - Quartz Scheduler - configurable SchedulerFactory
    * [SPR-16455] - Avoid String concatenation for not-null assertion in BeanProperty/DirectFieldBindingResult
    * [SPR-16457] - AcceptHeaderLocaleResolver should match country locales against supported language locales
    * [SPR-16535] - Support for ResolvableType.getType().getTypeName() on Java 8
    * [SPR-16558] - Consistent incrementer arrangement for PostgreSQL, DB2 and SAP HANA
    * [SPR-16587] - MockMvcResultMatchers.jsonPath(String).value() should have a matching method to declare the expected type
    * [SPR-16588] - Support for SimpleEvaluationContext in SpEL
    * [SPR-16596] - Consistent volatile access to running flag in Lifecycle implementations
    * [SPR-16628] - When `@DependsOn` throws a NoSuchBeanDefinitionException it should include the dependent bean for clarity
    * [SPR-16635] - Generate multipart boundary using SecureRandom 
    * [SPR-16667] - Reduce ClassUtils.forName overhead (in particular for annotation introspection purposes)
    * [SPR-16678] - Add cpp to mime.types as text/plain

## Release Notes - Spring Framework - Version 4.3.14

**Bug**
    * [SPR-15066] - Error in RestTemplate when setting the same HTTP header through ClientHttpRequestInterceptor and HttpEntity
    * [SPR-16196] - Combining `@Retryable` and `@Scheduled`/`@JmsListener` doesn't work
    * [SPR-16252] - Exception when receiving Long collection in MessageMapping
    * [SPR-16255] - NPE in FunctionReference due to race condition in SpelExpression.getValue()
    * [SPR-16262] - spring-web CORS requires X-Forwarded-Port
    * [SPR-16265] - Stomp Broker Relay may ignore configured destination prefixes
    * [SPR-16267] - Embedded cglib 3.2.5 not closing input streams that read class files
    * [SPR-16278] - BeanUtils.isSimpleValueType() returns false for enums overriding a method
    * [SPR-16281] - Unnecessary file system access in SimpleMetadataReaderFactory.getMetadataReader
    * [SPR-16288] - Ambiguous mapping error when using generic interface
    * [SPR-16295] - Programmatic creation of caching proxies using CacheProxyFactoryBean does not work
    * [SPR-16304] - Access-Control-Allow-Origin header returns wrong value using SockJS 
    * [SPR-16316] - Large transaction timeout value (Integer.MAX_VALUE for example) results in transaction expiring immediately after starting.
    * [SPR-16338] - `@JmsListener` concurrency property is ignored if DefaultJmsListenerContainerFactory#concurrency is set
    * [SPR-16353] - JMS Producers are cached even when the destination is a temporary queue causing a memory leak
    * [SPR-16364] - Null path after UriComponents.normalize() results in NullPointerException
    * [SPR-16367] - MockClientHttpResponse should not return null body
    * [SPR-16369] - TestExecutionListener class not found logged at INFO
    * [SPR-16371] - RestTemplate.ResponseEntityResponseExtractor doesn't tolerate unknown status codes
    * [SPR-16383] - EclipseLink does not log SQL parameters when using showSql

**Task**
    * [SPR-15624] - Update Stomp Client Recommendation in Docs
    * [SPR-16315] - Incorrect SpEL syntax in reference documentation
    * [SPR-16405] - Upgrade to ASM 6.0 level (for CGLIB 3.2.6)

**Improvement**
    * [SPR-16264] - Reduce access on user in SimpleBrokerMessageHandler.handleMessageInternal
    * [SPR-16275] - config.enableSimpleBroker("/topic", "/queue"); Should be config.enableSimpleBroker("/topic", "queue");
    * [SPR-16284] - Allow to inject enum with package visibility
    * [SPR-16293] - Improve performance of some string operations
    * [SPR-16356] - ComponentScanBeanDefinitionParser::parseTypeFilters should not fail on ClassNotFoundException
    * [SPR-16378] - Use ArrayList instead of LinkedList for known size

## Release Notes - Spring Framework - Version 4.3.13

**Bug**
    * [SPR-16058] - WebAsyncManager concurrency issue with SseEmitter when client disconnect
    * [SPR-16072] - RequestMapping method returning Future with null result causes NullPointerException
    * [SPR-16075] - setArguments(null) on MethodInvoker no longer coerces null to Object[0]
    * [SPR-16082] - Checkbox/RadioButton incorrectly converts collections of enums with a custom converter
    * [SPR-16083] - `@ModelAttribute` binding defined globally for particular attribute rather than per method invocation
    * [SPR-16102] - WebSphereUowTransactionManager swallows original exception when commit fails for another reason
    * [SPR-16103] - Incorrectly identify bridged method on interface
    * [SPR-16108] - DefaultResponseErrorHandler.hasError doesn't tolerate unknown status codes
    * [SPR-16117] - PathMatchingResourcePatternResolver returns duplicate resources when using classpath* prefix
    * [SPR-16122] - SpEL method invocation with varargs on proxy
    * [SPR-16132] - AbstractRequestExpectationManager fails with "Expectations already declared" when ResponseCreator.createResponse throws an exception
    * [SPR-16138] - MockHttpServletRequest with Host: set builds wrong getRequestURL()
    * [SPR-16146] - ClassPathResource.createRelative is using wrong ClassPathResource constructor for the returned resource
    * [SPR-16149] - Early ApplicationContext close call may lead to ApplicationEventMulticaster/LifecycleProcessor access exception
    * [SPR-16154] - When using NamedParameterJdbcTemplate, NVARCHAR or NCLOB(4000 characters or less) columns are not properly populated since StatementCreatorUtils does setString for these types instead of setNString.
    * [SPR-16160] - MockHttpServletResponse.getDateHeader fails with NPE for non-existing header
    * [SPR-16177] - NumberFormatException caused by property paths from JSR-303 based validation with no index into a collection
    * [SPR-16191] - Wrong byte code for compiled SpEL when JDK proxy method invocation is used

**Task**
    * [SPR-16074] - Doc: ControllerAdvice `@ExceptionHandler` matching for root vs cause in multi-advice scenario
    * [SPR-16076] - Improve documentation of lite configuration mode
    * [SPR-16078] - Clarify Bean destroyMethod documentation
    * [SPR-16092] - Doc: `@Async` in an Interceptor postHandle() doesn't work
    * [SPR-16107] - Compatibility with OpenPDF as alternative to iText 2.1.7
    * [SPR-16111] - Incorrect SpEL example in reference documentation
    * [SPR-16142] - Field name is used to match on bean name when `@Autowired` matches several beans by type
    * [SPR-16173] - Explicitly document UriTemplate as thread-safe
    * [SPR-16185] - End of first-class JDK 6 support
    * [SPR-16213] - Document common use cases for `@Order` vs `@Priority` vs `@DependsOn`
    * [SPR-16219] - Deprecate Castor support

**Improvement**
    * [SPR-16089] - Prevent WebSocket buffer overflow through application-level flow control
    * [SPR-16090] - SchedulingConfigurer and JmsListenerConfigurer should respect `@Order`
    * [SPR-16095] - Avoid temporary String creation in StringUtils.starts/endsWithIgnoreCase
    * [SPR-16162] - Make JpaVendorAdapters JTA-aware (in particular for Hibernate 5.1/5.2)
    * [SPR-16165] - Reduce access on headers for STOMP messaging
    * [SPR-16170] - spring-jdbc : Improve memory allocations when substituting named parameters.

## Release Notes - Spring Framework - Version 4.3.12

**Bug**
    * [SPR-15676] - Request params Optional<List<String> and List<String> are inconsistent
    * [SPR-15918] - java.util.Optional MultipartFile[] `@RequestParam` argument is null in multipart/form-data POST
    * [SPR-15919] - only one MultipartFile object populated when using an java.util.Optional MutipartFile array or list `@RequestParam`
    * [SPR-15952] - HttpEntityMethodProcessor discards headers
    * [SPR-15962] - TaskExecutorRegistration.getTaskExecutor() overrides executor properties of a provided ThreadPoolTaskExecutor
    * [SPR-15965] - JmsMessagingTemplate is not correctly configured
    * [SPR-15976] - ChannelRegistration.setInterceptors is misnamed
    * [SPR-15978] - RestTemplate doesn't consistently tolerate unknown HTTP status codes
    * [SPR-15989] - PathMatchingResourcePatternResolver provides duplicate resources when using classpath* prefix combined with ant-style
    * [SPR-16032] - Spring EL does not allow '\0' characters
    * [SPR-16043] - sort BeanDefinitionRegistryPostProcessors added by other BeanDefinitionRegistryPostProcessors
    * [SPR-16123] - SpelExpression throws NullPointerException instead of EvaluationException for primitives

**Task**
    * [SPR-15973] - Clarify `@Bean` return type recommendation in case of multiple interfaces
    * [SPR-16016] - Deprecate HibernateJpaSessionFactoryBean (against Hibernate 5.2)

**Improvement**
    * [SPR-15854] - Backport s/s/m/j/o/s/m/w/MockPart.java from 5.x to 4.3.x
    * [SPR-16054] - Add convenient method to construct ParameterizedTypeReference from Type

**Refactoring**
    * [SPR-15656] - Backport selected refinements from the nullability efforts in 5.0
    * [SPR-16017] - Consistent configurer access in WebMvcConfigurationSupport


ChristianMurphy:
Smoke testing passes, views still appear to work as expected.

screencapture-localhost-8080-uportal-f-admin-normal-render-up-2018-05-11-07_11_53
screencapture-localhost-8080-uportal-f-welcome-normal-render-up-2018-05-11-07_12_17

PR #1223, opened 2018-06-21
chore: update commons collections to version 4.1

ChristianMurphy:
##### Checklist

##### Description of change

https://commons.apache.org/proper/commons-collections/release_4_0.html
https://commons.apache.org/proper/commons-collections/release_4_1.html


ChristianMurphy:
Smoke tested locally, no issues

uPortal-rendering/src/main/java/org/apereo/portal/portlet/registry/PortletEntityRegistryImpl.java
```diff
@@ -27,7 +27,7 @@
 import javax.servlet.http.HttpSession;
 import net.sf.ehcache.Ehcache;
 import net.sf.ehcache.Element;
-import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.collections4.CollectionUtils;
```
drewwills: @ChristianMurphy -- That's a bit funny...

So this library is used in one line of code in one file???  If that's the case, I'd say we refactor the line and drop the dependency.

But I suspect you will find that the earlier version of commons-collections still ends up in the deployment b/c it's a transitive dependency of one or more other dependencies.  The fact that Apache re-packaged the sources -- i.e. `org.apache.commons.collections4...` -- means the 2 versions can run side-by-side.

PR #1243, opened 2018-07-03
Using xsl:attribute to build the <body>'s class attribute

jonathanmtran:
##### Checklist

##### Description of change

We have a need to style the portal a little differently when viewing portlets in maximized mode. Building the `<body>`'s `class` attribute was inspired by WordPress [body_class()](https://developer.wordpress.org/reference/functions/body_class/).

### DIFF
--- a/uPortal-webapp/src/main/resources/layout/theme/respondr/respondr.xsl
+++ b/uPortal-webapp/src/main/resources/layout/theme/respondr/respondr.xsl
@@ -652,7 +652,11 @@
             -->
             <chunk-point/> <!-- Performance Optimization, see ChunkPointPlaceholderEventSource -->
         </head>
-        <body class="up dashboard portal fl-theme-mist">
+        <body>
+          <xsl:attribute name="class">
+            <xsl:text>up dashboard portal fl-theme-mist</xsl:text>
+            <xsl:if test="$PORTAL_VIEW='focused'"> focused</xsl:if>
+          </xsl:attribute>
           <xsl:call-template name="skipnav" />
           <div class="row-offcanvas">
             <div id="up-notification"></div>
--- a/uPortal-webapp/src/main/resources/layout/theme/respondr/respondr.xsl
+++ b/uPortal-webapp/src/main/resources/layout/theme/respondr/respondr.xsl
@@ -655,7 +655,7 @@
         <body>
           <xsl:attribute name="class">
             <xsl:text>up dashboard portal fl-theme-mist</xsl:text>
-            <xsl:if test="$PORTAL_VIEW='focused'"> focused</xsl:if>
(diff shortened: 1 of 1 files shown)


ChristianMurphy:
CI failures look like potential false positives related to network disruption at Travis CI.
Restarting to see if the issue resolves.

uPortal-webapp/src/main/resources/layout/theme/respondr/respondr.xsl
```diff
@@ -652,7 +652,11 @@
             -->
             <chunk-point/> <!-- Performance Optimization, see ChunkPointPlaceholderEventSource -->
         </head>
-        <body class="up dashboard portal fl-theme-mist">
+        <body>
+          <xsl:attribute name="class">
+            <xsl:text>up dashboard portal fl-theme-mist</xsl:text>
```
drewwills: I'm not sure that `fl-theme-mist` is needed any longer.  I'd be curious to know if anyone can perceive a difference without it.

uPortal-webapp/src/main/resources/layout/theme/respondr/respondr.xsl
```diff
@@ -652,7 +652,11 @@
...
+            <xsl:if test="$PORTAL_VIEW='focused'"> focused</xsl:if>
+          </xsl:attribute>
```
drewwills: I like the solution, generally speaking.  Should the '` focused`' characters also be wrapped in a `<xsl:text>`?  What's the purpose of that?

jonathanmtran: [respondr.xsl#L156](https://github.com/Jasig/uPortal//blob/HEAD/uPortal-webapp/src/main/resources/layout/theme/respondr/respondr.xsl#L156) doesn't have have `<xsl:text>`. The reason for using `<xsl:text>` is to have the classes there regardless of the mode ... hopefully that answers the question

drewwills:
@jonathanmtran -- I did think of one suggestion...
If we're going to add a new CSS class, could we apply a `up-` prefix?  (So `up-focused` works here afaik.)
We use CSS classes for a wide variety of reasons and in collaboration with a number of frameworks.  The prefixes help shed some light on the purpose of the class, for those who see it later and want to know.

jonathanmtran:
`up-focused` makes sense. I'll update the branch shortly

ChristianMurphy:
:thinking: Strange, OpenJDK 8 continues to fail on CI, I'll pull this down locally and see if I can reproduce.

ChristianMurphy:
Not able to reproduce locally, on third CI restart passed :ballot_box_with_check:

PR #698, opened 2016-07-25
[WiP] UP-4711: smartLdapGroupStore adapted/improvement

jgribonvald:
JIRA : https://issues.jasig.org/browse/UP-4711

This work is the result of a merged work done on up-v4.0 in context of ESUP consortium.

I have made some little change (for optimization/conf), but it should be tested if this doesn't break something in use of an Active Directory or overlay's groups.

This PR is only to show (for now) the work and to wait more tests on other context. 
These change works for me on an openLdap with groups published on LDAP by ldappc from Internet Grouper. (We don't use Grouper's api to request groups and groups memberships as we have more than 50 000 groups)

Also test cases are needed, I will be able to works on only after september (it's in my timelines plans, but not a priority :cry: )

So testing on other contexts are needed and will be welcome !

If you need help in configuration, ask ! I can try to provide somes examples if it's needed.

PR #1283, opened 2018-07-27
UP-4711: smartLdapGroupStore improvements

jgribonvald:
##### Checklist

-   [ ] tests are included

##### Description of change

This permit to set a new king of ldap groups loading.
Docs + test will follow in some month but I can provide a description in conversation.
This code works for us and somes other since several years in production.

### DIFF
--- a/uPortal-groups/uPortal-groups-smartldap/src/main/java/org/apereo/portal/groups/smartldap/SmartLdapGroupStore.java
+++ b/uPortal-groups/uPortal-groups-smartldap/src/main/java/org/apereo/portal/groups/smartldap/SmartLdapGroupStore.java
@@ -30,8 +30,10 @@ import org.apache.commons.lang3.concurrent.LazyInitializer;
 import org.apereo.portal.EntityIdentifier;
 import org.apereo.portal.groups.ComponentGroupServiceDescriptor;
 import org.apereo.portal.groups.EntityGroupImpl;
+import org.apereo.portal.groups.EntityImpl;
 import org.apereo.portal.groups.EntityTestingGroupImpl;
 import org.apereo.portal.groups.GroupsException;
+import org.apereo.portal.groups.IEntity;
 import org.apereo.portal.groups.IEntityGroup;
 import org.apereo.portal.groups.IEntityGroupStore;
 import org.apereo.portal.groups.IEntityGroupStoreFactory;
@@ -63,6 +65,18 @@ public final class SmartLdapGroupStore implements IEntityGroupStore {
 
     public void setBaseGroupDn(String baseGroupDn) {
         this.baseGroupDn = baseGroupDn;
+    }
+
+    private String childGroupKeyRegex = null;
+
+    public void setChildGroupKeyRegex(String childGroupKeyRegex) {
+        this.childGroupKeyRegex = childGroupKeyRegex;
+    }
+
+    private String groupTreeSeparator = ":";
+
+    public void setGroupTreeSeparator(String groupTreeSeparator) {
+        this.groupTreeSeparator = groupTreeSeparator;
     }
 
     private String filter = "(objectCategory=group)"; // default
@@ -124,6 +138,19 @@ public final class SmartLdapGroupStore implements IEntityGroupStore {
     private IPersonAttributeDao personAttributeDao;
 
     private final Logger log = LoggerFactory.getLogger(getClass());
+
+    /**
+     * Set it to true to show peoples that are members of groups. Warning must be tested before
+     * production if set to true !
+     */
+    private boolean displayPersonMembers = false;
+
+    public void setDisplayPersonMembers(boolean displayPersonMembers) {
+        this.displayPersonMembers = displayPersonMembers;
+    }
+
+    private Map<String, IEntity> persons =
+            Collections.synchronizedMap(new HashMap<String, IEntity>());
 
     /*
      * Indexed Collections.
@@ -274,7 +301,19 @@ public final class SmartLdapGroupStore implements IEntityGroupStore {
         log.debug("Invoking findEntitiesForGroup() for group:  {}", group.getLocalKey());
 
         // We only deal w/ group-group relationships here...
-        return findMemberGroups(group);
+        // return findMemberGroups(group);
+
+        List<IEntity> rslt = new ArrayList<IEntity>();
+        if (displayPersonMembers) {
+            List<String> list = groupsTree.getPersonChildren().get(group.getLocalKey());
+            if (list != null) {
+                // should only reach this code if its a SmartLdap managed group...
+                for (String s : list) {
+                    rslt.add(persons.get(s));
+                }
+            }
+        }
+        return rslt.iterator();
     }
 
     @Override
@@ -342,6 +381,13 @@ public final class SmartLdapGroupStore implements IEntityGroupStore {
         return rslt.iterator();
     }
 
+    /**
+     * Return an UnsupportedOperationException !
+     *
+     * @param entityType
+     * @return
+     * @throws GroupsException
+     */
     @Override
     public IEntityGroup newInstance(Class entityType) throws GroupsException {
         log.warn("Unsupported method accessed:  SmartLdapGroupStore.newInstance");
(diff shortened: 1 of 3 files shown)


jgribonvald:
Some docs for the use case:

**Description of group structure**
```
ou=people,dc=org,dc=fr
     uid=XXXXX,ou=peoples,dc=org,dc=fr 
           uid=XXXXX
           cn,...
           isMemberOf=leaf1:subleaf1:group1
           isMemberOf=leaf1:subleaf1:group2
           ismemberOf=.....
           isMemberOf=leaf1:subleaf5:groupx

ou=groups,,dc=org,dc=fr
     cn=leaf1:subleaf1:group1,ou=groups,,dc=org,dc=fr
           cn=leaf1:subleaf1:group1
           member=uid=XXXXX,ou=peoples,dc=org,dc=fr
           member= cn=leaf1:subleaf5:groupx,ou=groups,,dc=org,dc=fr
           member=....(dn only of peoples or groups)
           hasMember=leaf1:subleaf5:groupx
           hasMember= XXXXX
           hasMember=...(id key, so uid for people object or cn for group object)      
           
```
like you can see each groups is a flat tree representation, and we recompose the tree into uPortal by decomposition with the ":" char, it could be any other char (but must be forbidden in groups names)

** Example of configuration for a such use case **
```xml
 <bean id="smartLdapGroupStore" class="org.apereo.portal.groups.smartldap.SmartLdapGroupStore" lazy-init="false" depends-on="defaultLdapContext">

        <property name="ldapContext" ref="defaultLdapContext"/>

        <property name="memberOfAttributeName" value="isMemberOf"/>

        <property name="baseGroupDn" value="ou=groups"/>

        <property name="childGroupKeyRegex" value="cn=(.*),ou=groups,${ldap.defaultLdapContext.baseDn}"/>

        <property name="filter" value="(objectClass=groupOfNames)"/>

        <property name="groupsTreeRefreshIntervalSeconds" value="${smartldapgroupstore.groupsTreeRefreshIntervalSeconds:900}"/>

        <property name="resolveMemberGroups" value="false"/>
        <property name="resolveDn" value=""/><!--Used with resolveMemberGroups -->

        <property name="attributesMapper">
            <bean class="org.apereo.portal.groups.smartldap.SimpleAttributesMapper">
                <property name="keyAttributeName">
                    <value>cn</value>
                </property>
                <property name="groupNameAttributeName">
                    <value>cn</value>
                </property>
                <property name="membershipAttributeName">
                    <value>member</value>
                </property>
            </bean>
        </property>
    </bean>
```
We can also resolve peoples members into groups, but it consume a lot of resources on big number of groups and users (400k users into 45k groups, so it's disable for us)

jgribonvald:
These change should not break anything and could work at least with openldap memberOf overlay.

For test case I would like to provide an embedded LDAP with unboundid-ldap-sdk https://ldap.com/unboundid-ldap-sdk-for-java/ like I already done with a custom portlet to retrieve organizational informations and cache it (when you manage more than 250 organizations), 
here is an example of implementation: https://github.com/GIP-RECIA/esco-ChangeEtablissement/commit/92e76122d3f32395a48094c05714ea9ee619cb28

The only thing is that it doesn't simulate an Active Directory, but seemingly tools could be found. And these change should be tested on before merging (I can't make test on).

Any help in testing on native use and against Active Directory would be appreciated.

jgribonvald:
For previous réference see https://github.com/Jasig/uPortal/pull/698

uPortal-groups/uPortal-groups-smartldap/src/main/java/org/apereo/portal/groups/smartldap/SmartLdapGroupStore.java
```diff
@@ -630,6 +690,61 @@ private GroupsTree buildGroupsTree() {
...
+        // Process each group in order to create the groups that will represent the Grouper's
+        // folders
+        // and maintain the consistency regarding the membership.
+        for (String groupKey : newGroups.keySet()) {
+
+            boolean stop = false;
+            boolean isAFolder = false;
+            String currentPath = groupKey;
+            while (!stop) {
+                final String containingFolder = getContainingFolder(currentPath);
```
ChristianMurphy: If I'm reading this correctly, current path will be a `:` separated list of paths, and this loops until is has processed all folders.
Would something like `for (String containingFolder : groupKey.split(":")) {` work?

jgribonvald: This should work. This code come from a Java 1.5 version and I migrate only needed parts...

jgribonvald: Finally this won't works as the id of the group is the entire path and the containing folder is the entire path, so we need to remove only the last part each time we want the containing folder.

As example if you have this group:
Organizations:org_a:students:class_A
the first containing folder is:
Organizations:org_a:students
the next containing folder is:
Organizations:org_a
and the last containing folder is:
Organizations

So you need to make it in this way, with your solution you will need to rebuild the id path on each values.

ChristianMurphy: I understand. :thinking:
My main concern is with `while (!stop) {`, using a flag as a `while` conditional tends to be a code smell. :nose: 
What would be the implications of something like:

```java
var groupKeys = groupKey.split(":");
while (groupKeys.size() > 0) {
  var containingFolder = String.join(":", groupKeys);

  // logic goes here

  groupKeys.removeLast();
}
```

jgribonvald: I'm agree the entire code of the loop could be rewrited ;)  I Will watch it when I will be back from holidays (I can't before I try to solve this issue https://github.com/uPortal-contrib/uPortal-web-components/issues/43)

jgribonvald: what do you think about that:
```java

for (String groupKey : newGroups.keySet()) {
            String currentPath = groupKey;

            // The group that represents the grouper folder is created.
            final IEntityGroup folderGroup =
                new EntityTestingGroupImpl(currentPath, IPerson.class);
            folderGroup.setCreatorID("System");
            folderGroup.setName(currentPath);
            folderGroup.setDescription("Grouper Folder");
            newFolders.put(currentPath, folderGroup);

            String containingFolder = this.getContainingFolder(currentPath);
            while (!containingFolder.isEmpty()) {
                // Adds the containing folder as a parent of this current path
                List<String> parentsList = newParents.get(currentPath);
               ...
                childrenList.add(currentPath);

                // The remaining prefix of the group has already been processed.
                if (newFolders.containsKey(containingFolder)) {
                   break; 
                }
                currentPath = containingFolder;

                containingFolder = this.getContainingFolder(currentPath);
            }
        }
```

ChristianMurphy: That could work :+1:

jgribonvald:
Requested change done and tested !

uPortal-webapp/src/main/resources/org/apereo/portal/groups/smartldap/init.crn
```diff
@@ -42,28 +43,75 @@
...
+		            record = new LdapRecord(record.getGroup(), newChildren);
+	            }
+	            GROUPS.add(record);
+
+	            //   Parses the group names in order to handle the Grouper folders.
+
+                String current = record.getGroup().getName();
+                int sepPos;
+                boolean stop = false;
+	            boolean isGroup = true;
```
ChristianMurphy: Indented different than line 88?

jgribonvald: the indent is not well managed with crn files ...

jgribonvald:
review on file indent !

PR #1284, opened 2018-07-27
feat(xsl): don't show mezzanine content on focused mode

jgribonvald:
##### Checklist

-   [ ] tests are included

##### Description of change

Purposed feature to render only the content of the mezzanine region not in focused mode, it permit to place in this region a content that won't be always rendered when we are focusing on a portlet. it could be a good thing to have a such part (in header) not shown in focused mode.

### DIFF
--- a/uPortal-webapp/src/main/resources/layout/structure/columns/columns.xsl
+++ b/uPortal-webapp/src/main/resources/layout/structure/columns/columns.xsl
@@ -166,7 +166,7 @@
                     </xsl:when>
                     <xsl:otherwise>
                         <!-- Include all regions EXCEPT 'region-customize' when in FOCUSED mode -->
-                        <xsl:for-each select="child::folder[@type!='customize' and @type!='regular' and @type!='sidebar' and channel]"><!-- Ignores empty folders -->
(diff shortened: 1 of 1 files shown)


jgribonvald:
The same thing could be done on header-bottom region, but i think the mezzanine is a the better one region do hide in this case.

jgribonvald:
If you prefer a better approach I'm open, I thought about adding a property (shownInFocused) in DLM on the channel or folder region entry ? but it's a bit fastidious to do, and I'm not sure this will be a good way

uPortal-webapp/src/main/resources/layout/structure/columns/columns.xsl
```diff
@@ -166,7 +166,7 @@
                     </xsl:when>
                     <xsl:otherwise>
                         <!-- Include all regions EXCEPT 'region-customize' when in FOCUSED mode -->
```
drewwills: Can you update this comment to reflect your changes?

drewwills:
I'm okay with this change.  :+1:

jgribonvald:
I Think also I should provide a
documentation about that. Is there a documentation that provide a list of
all region with an overview of their position like it was done before ? (uP
4.?)

drewwills:
Documentation of that topic would be most welcome.

drew

drewwills:
Updated in https://github.com/Jasig/uPortal/pull/1293

PR #1314, opened 2018-08-21
fix(SmartLdapGroupStore): fix configurable property not used

jgribonvald:
##### Checklist

-   [ ] tests are included

##### Description of change

Fix init.crn that doesn't use the property that permit to customize (other than ":") the char to decompose group tree in group id/cn.

### DIFF
--- a/uPortal-webapp/src/main/resources/org/apereo/portal/groups/smartldap/init.crn
+++ b/uPortal-webapp/src/main/resources/org/apereo/portal/groups/smartldap/init.crn
@@ -69,7 +69,7 @@
 
                 def newChildren=[];
                 if (req.hasAttribute('childGroupKeyRegex')) {
-                    def pattern = req.hasAttribute('pattern') ? req.getAttribute('pattern') : java.util.regex.Pattern.compile(ScriptAttributes.REQUEST.getAttribute("childGroupKeyRegex"));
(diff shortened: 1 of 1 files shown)


uPortal-webapp/src/main/resources/org/apereo/portal/groups/smartldap/init.crn
```diff
@@ -87,7 +87,8 @@
                     int sepPos;
                     boolean stop = false;
                     boolean isGroup = true;
-                    while ((sepPos=current.lastIndexOf(':')) > 0 &amp;&amp; ! stop) {
+                    String sep = req.hasAttribute('groupTreeSeparator') ? req.getAttribute("groupTreeSeparator") : ':';
```
drewwills: Where does `groupTreeSeparator` come from?  Was that something already in the configs, but ignored?

jgribonvald: This is coming from my previous Pull Request that changed the SmartLdapGroupStore to be able to work with groups of kind `root:leaf1:subleaf1:group` with tree decomposition.
The property is configurable (but not totally from original ESUP code, and I've found this minor "bug" on reading an other time the init.crn, I didn't encountered the bug as I'm using the default char `:` as group tree separator, so if you use an other separator this won't work expect after my patch !

Else the value is setted from the SmartLdapGroupStore bean and is passed from this code : https://github.com/Jasig/uPortal/blob/master/uPortal-groups/uPortal-groups-smartldap/src/main/java/org/apereo/portal/groups/smartldap/SmartLdapGroupStore.java#L672

PR #1318, opened 2018-08-23
Cleaning on personManager bean

jgribonvald:
##### Checklist

-   [*] the [individual contributor license agreement][] is signed
-   [ ] tests are included

##### Description of change

Moving / cleaning / improving the default PersonManager bean (see comments on commits)


uPortal-security/uPortal-security-core/src/main/java/org/apereo/portal/security/provider/PersonManagerExtenderImpl.java
```diff
@@ -0,0 +1,52 @@
+package org.apereo.portal.security.provider;
+
+import io.jsonwebtoken.lang.Assert;
```
drewwills: This seems like a strange class to use for `Assert.notNull()`.  (Isn't there a Spring alternative?)

drewwills:
@jgribonvald -- Do you need either of the following?...

  - The ability to add attributes to a new `IPerson` object _before_ person-directory processes user attributes;  or
  - The ability for different instances of the same guest user (remember -- multiple guest users are supported) to have different attribute values

I hope the answers are "no" and "no."

If that's the case, I'm confident we can hit your use cases with simply guest users & user attributes.

jgribonvald:
@drewwills : so 
- no need to be before, or it because i don't have acknowledge of a process that use IPerson object before it was loaded from person-directory. Also I've made the change in this way to avoid to modify anything in the person directory, as the serverName is more a request attribute process than person-directory is...
- not sure to understand the finality, what I need is to use the serverName to associate theme and channels (so on multi guest mode I will associate a different guest user per serverName).

jgribonvald:
@drewwills maybe I've got it

I don't necessary need to import the serverName into user informations as I can map a different guest user depending on serverName => that's ok

drewwills:
@jgribonvald,

    I don't necessary need to import the serverName into user informations as I can map a different guest user depending on serverName => that's ok

Yes!  This is the easiest part -- the method that decides which guest user you get (and which you need to implement) takes an `HttpServletRequest`.  It should be very straightforward to fulfill your use case here.

    But I need the serverName for authenticated users to map the good skin, or maybe there is a theme transformer that can apply on session attributes (that should contains the serverName) ?

This use case will also require very little work, but it appears that it _will_ require a change to uPortal.

This bean...

  - https://github.com/Jasig/uPortal/blob/master/uPortal-webapp/src/main/java/org/apereo/portal/context/persondir/PersonDirectoryConfiguration.java#L143

Has this method...

  - https://developer.jasig.org/projects/person-directory/1.5.0-SNAPSHOT/apidocs/org/jasig/services/persondir/support/web/RequestAttributeSourceFilter.html#setServerNameAttribute(java.lang.String)

but (1) we're not calling it by default, and (2) we're not offering a configurable way to invoke it.

Either (1) or (2) needs to change.  The best outcome, I think, would be for someone to set up a clean/elegant way to configure all the options of that SevletFilter inside uPortal-start.  (Your item is simple -- just a property in `uPortal.properties`.)

jgribonvald:
@drewwills : Ok So I will watch to migrate my change to get the serverName into the personDirectory. I've seen this possibility but I didn't like it as it's the personDirectory and for me it used only to load informations from a datasource not to add other properties comming outsite of the datasource... But if you prefer this way I can watch to do it !

drewwills:
@jgribonvald,

  I've seen this possibility but I didn't like it as it's the personDirectory and for me it's used only to load informations from a datasource not to add other properties comming outsite of the datasource... But if you prefer this way I can watch to do it !

I feel obligated to advocate for doing it with person-directory.

uPortal is large and complex.  It's can more difficult to maintain than I wish.  To become more agile, we need to look for things we can let go of.  We need one really good way of doing something, rather than several ways to do something.

User attributes is a "core subsystem" of uPortal -- we will never shed it.  We should focus on leveraging, improving, and better-documenting the "core" parts of uPortal.

jgribonvald:
After looking on it seems that I don't need to push a code to be able to add the serverName. I need only in my customization to add one IPersonAttributeDao bean that provide my property and that use the requestAttributeSourceFilter bean to get the serverName.
Or maybe you wan't that I make a Pull Request to activate it simply by a property ?

drewwills:
I think perhaps the best move here is to update the out-of-the-box config of It `requestAdditionalDescriptors` to include the `serverName` attribute by default.  (It can't hurt anything.)

jgribonvald:
Finaly it works pretty well without add, I don't understand why i don't get it before (maybe it worked after an update of the uPortal version deployed)...

So my PR is more a cleaning of useless code !

drewwills:
LGTM :+1:

PR #689, opened 2016-07-06
UP-4727 :  Managing uPortal deployment on several servername with CAS

jgribonvald:
- update CAS client to be able to manage several client server Name (since version 3.3.0 only)
- setting CAS login and logout url using a pattern replacement to use the good url
- provinding tools to manage several server names without security hole - we set all possibles server names and we can use only them else only the first one is used
- change login jsp invoker to adapt the login url change
- provide a new property environment.build.uportal.otherServers to set other server names added to the main setted by environment.build.uportal.server

WARNING: This PR is based on #688

PR #1320, opened 2018-08-29
UP-4727 : Managing uPortal running on several servername with CAS

jgribonvald:
##### Checklist

-   [*] the [individual contributor license agreement][] is signed
-   [*] commit message follows [commit guidelines][]
-   [*] tests are included

##### Description of change

Upgrade from the previous Pull Request: https://github.com/Jasig/uPortal/pull/689

This PR is needed when you are deploying a multi-tenant uPortal instance working on several server name. In that case the connexion URL should be adapted to the serverName accessed. As example when the user access to https://servername.edu/uPortal the cas auth url is like https://cas.servername.edu/cas/login?service=https://servername.edu/uPortal/login.
But on an other servername like https://othername.edu/uPortal the service should change to https://cas.servername.edu/cas/login?service=https://othername.edu/uPortal/login.

So this PR:
- Provide a configurable way to set the current uPortal serverName to cas service URL:
- Providing also a decorator to be able to set WAYF parameters depending on condition like current serverName. I'm using a such decorator on some serverName where we need to bypass the WAYF. or to add a generated token (this token generator is not provided as internal, but could be added by bean configuration)

### DIFF
--- a/uPortal-security/uPortal-security-mvc/src/test/java/org/apereo/portal/url/CasLoginRefUrlEncoderTest.java
+++ b/uPortal-security/uPortal-security-mvc/src/test/java/org/apereo/portal/url/CasLoginRefUrlEncoderTest.java
@@ -16,12 +16,15 @@ package org.apereo.portal.url;
 
 import static org.junit.Assert.assertEquals;
 
+import com.google.common.collect.Sets;
 import java.io.UnsupportedEncodingException;
+import java.util.ArrayList;
+import java.util.List;
 import org.junit.Before;
 import org.junit.Test;
 import org.springframework.mock.web.MockHttpServletRequest;
 
-/** @Author James Wennmacher, jwennmacher@unicon.net */
+/** Some Testing on CasLoginRefUrlEncoder class. */
 public class CasLoginRefUrlEncoderTest {
 
     CasLoginRefUrlEncoder encoder;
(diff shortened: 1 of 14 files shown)


uPortal-security/uPortal-security-mvc/src/test/java/org/apereo/portal/url/CasLoginRefUrlEncoderTest.java
```diff
@@ -16,12 +16,15 @@
...
+import com.google.common.collect.Sets;
 import java.io.UnsupportedEncodingException;
+import java.util.ArrayList;
+import java.util.List;
 import org.junit.Before;
 import org.junit.Test;
 import org.springframework.mock.web.MockHttpServletRequest;
 
-/** @Author James Wennmacher, jwennmacher@unicon.net */
+/** */
```
ChristianMurphy: :art: empty doc comment?

uPortal-url/src/main/java/org/apereo/portal/url/IAuthUrlCustomizer.java
```diff
@@ -0,0 +1,22 @@
+package org.apereo.portal.url;
+
+import javax.servlet.http.HttpServletRequest;
+
+/** Customize CAS or any other authentication URL. */
+public interface IAuthUrlCustomizer {
+    /**
+     * Does this customizer supports the current HTTP request.
+     *
+     * @param request
```
ChristianMurphy: `request` but no `url` param?

uPortal-url/src/main/java/org/apereo/portal/url/IAuthUrlCustomizer.java
```diff
@@ -0,0 +1,22 @@
...
+     * @return
+     */
+    boolean supports(HttpServletRequest request, String url);
+
+    /**
+     * Customize the supplied external login URL.
+     *
+     * @param url
```
ChristianMurphy: `url` but no `request` param?

uPortal-security/uPortal-security-mvc/src/test/java/org/apereo/portal/url/CasLoginRefUrlEncoderTest.java
jgribonvald: @Author is not requested it seems, so what do you want that I add

ChristianMurphy: Or remove the comment entirely

jgribonvald: I pushed a change and added a comment, but several other test class have a such empty comment, that's why I did it...

uPortal-url/src/main/java/org/apereo/portal/url/IAuthUrlCustomizer.java
jgribonvald: The 2 methods have the sames params, but each one had a different param :laughing: 
=> Solved

for notes I only moved the working code from my old previous PR, applied it on my portal instance (without a lot of change) and finally come back to push it, so sorry i didn't take the time to look more on the code.

jgribonvald:
I have to watch on how to get processed the login url into oidc issuer. Any idea ? I would need the curent request or maybe registering into auth session object the URL.

ChristianMurphy:
I'm not sure I quite understand the use case you are going for.
My best guess that you want the url access the front end code, for that including the Url as a user attribute and custom claim would work.

jgribonvald:
The problem with my commit is that issuer have the default url from uPortal.properties, the URL isn't processed by my Customizer and so the url isn't good into oidc attributes like the issuer (which is set from uPortal.properties)

jgribonvald:
I've updated the commit informations about the use case.

PR #1352, opened 2018-09-25
Removing printable feature

jgribonvald:
##### Checklist

-   [ ] tests are included

##### Description of change

Removing the printable feature:
- removing the jquery plugin that doesn't work anymore and is not efficient
- removing the option from the xsl and portlet params definition
- removing the i18n references
I didn't find any other references, but when doing a grep with "print" isn't easy to identify something related.

the only thing that can't be removed is the fact that adding any param from portlet definition as any param can be added and inserted (the script doesn't filter params that aren't recognized/used).


ChristianMurphy:
@drewwills @apetro Would it be safe to merge this into uPortal 5.3 or should this wait til uPortal 6?

drewwills:
I think it's safe -- and a good idea -- to merge this change into 5.3.

Issue #1321, opened 2018-08-30
Portlet Administrator UI modify properties with wrong charset


jgribonvald:
**Describe the bug**

The UI doesn't save portlet information with the good charset see animated gif from @ChristianMurphy  bellow (reproduction test).
The problems is only in the UI as XML files doesn't have this problem. Also this problem didn't exist before, at least in uP4.3.

**To Reproduce**
![example-with-accent-marks](https://user-images.githubusercontent.com/1597283/44867690-da222280-ac88-11e8-95a3-ffff08ac6bf8.gif)

**Expected behavior**
Saving with UI cause bad charset encoding on UTF-8 chars like [é à ...]

**Platform:**
 - uPortal Version: uPortal 5.2.2
 - Browser: Chrome, Firefox

ChristianMurphy:
Resolved in #1321

PR #1363, opened 2018-10-06
Issue #1321: Fix character encoding on form submit

dodie:
##### Description of change
Fix for issue  #1321. The Character encoding filter has to be applied early in the filter chain to take effect. If another filter touches the request data before it could set the encoding properly, it will just silently do nothing. 

image


dodie:
Hi @ChristianMurphy ,

Sure, I've just filled it.

Thanks and cheers,
David

drewwills:
Welcome @dodie -- thanks for posting this contribution.

PR #1375, opened 2018-10-12
fix:  correct a mistake in JDBC code when new users are provisioned

drewwills:
##### Checklist

##### Description of change

This PR fixes a small mistake in JDBC code that affects the ability of the portal to provision new users when they authenticate for the first time.


drewwills:
Not sure what the issue was or how this fixes it, but the code looks okay.

The query upon which that `ResultSet` was based had been adjusted, and there was no longer a column # 4.

PR #1378, opened 2018-10-19
1325 google custom search

bjagg:
##### Checklist

##### Description of change

Update Google Custom Search class to support changes in remote service.

### DIFF
--- a/uPortal-portlets/src/main/java/org/apereo/portal/portlets/search/google/GoogleCustomSearchService.java
+++ b/uPortal-portlets/src/main/java/org/apereo/portal/portlets/search/google/GoogleCustomSearchService.java
@@ -24,6 +24,8 @@ import org.apereo.portal.portlets.search.IPortalSearchService;
 import org.apereo.portal.search.SearchRequest;
 import org.apereo.portal.search.SearchResult;
 import org.apereo.portal.search.SearchResults;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.web.client.RestOperations;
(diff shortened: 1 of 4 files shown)


bjagg:
Was able to test directly with Google Service. Addressed an issue with null parameter causing an exception and another issue where no return values threw an exception.

uPortal-portlets/src/main/java/org/apereo/portal/portlets/search/google/GoogleCustomSearchService.java
```diff
@@ -64,22 +67,38 @@
...
+    public GoogleCustomSearchService() {
+        logger.debug("GoogleCustomSearchService bean instantiated.");
+    }
+
+    public void setResultSize(int resultSize) {
+        // search fails if this parameter is outside the range 1-10, inclusive
+        if (resultSize < 1) {
+            this.resultSize = 1;
+        } else if (resultSize > 10) {
+            this.resultSize = 10;
```
ChristianMurphy: If there are > 10 results, can the results outside the first 10 still be accessed?

bjagg: There is the concept of paging in uPortal, and the current page requested is passed in at every call for search results. It might be another issue with the search module.

ChristianMurphy: Okay, so should this be managed by `RESULT_SIZE_PARAM` on line 61 then?

bjagg: `resultSize` is the portal configuration value to use as that parameter value. See line 116.

drewwills:
The solution and the code look fine to me, but I'm seeing...

```
  GradleException: There were failing tests.
```

Issue #1409, opened 2018-11-15
Property 'cas.enable.redirect.guest.to.login' not working because of _CURRENT_SERVER_NAME_ token


drewwills:
## Description

The new support for multiple CAS servers breaks the `cas.enable.redirect.guest.to.login` property.

Some configs (currently)...

```
portal.protocol.server.context=${portal.protocol}://_CURRENT_SERVER_NAME_${portal.context}
[...]
portal.login.url=${portal.protocol.server.context}/Login
[...]
org.apereo.portal.channels.CLogin.CasLoginUrl=${cas.protocol.server.context}/login?service=${portal.login.url}
```

## To Reproduce

Set `cas.enable.redirect.guest.to.login=true`.

drewwills:
@mrapczynski, @ChristianMurphy, @jgribonvald

jgribonvald:
@drewwills : what do you obtain as url ?
Maybe I missed somewhere to apply the decorator of this url. But in all case you have the possibility to set the property `portal.protocol.server.context` to override the default's one

ChristianMurphy:
> what do you obtain as url ?

`_CURRENT_SERVER_NAME_`

jgribonvald:
ok I missed to apply the decoration somewhere, so this isn't not too much complicated to apply, but have to find where ...

jgribonvald:
Ok I see the problem in the bean init, so 2 choice:

jgribonvald:
this should be solved by #1410

PR #1410, opened 2018-11-16
fix: missing conf on serveral server name management when activating …

jgribonvald:
…cas guest redirect

##### Checklist

-   [ ] tests are included

##### Description of change

This change must solve the issue #1409 (tested locally)


jgribonvald:
After watching where the property `cas.enable.redirect.guest.to.login` is used I've found a "nonsense" configuration:

jgribonvald:
After watching where the property `cas.enable.redirect.guest.to.login` is used I've found a "nonsense" configuration:
the file `security.properties` set the property `cas.enable.redirect.guest.to.login=false` whereas the bean is defined with this property with a default value
```xml
<bean name="casAuthenticationFilter" class="org.apereo.portal.spring.web.OptionalFilterDecorator">
        <property name="wrappedFilter" ref="casInnerAuthenticationFilter" />
        <property name="enable" value="${cas.enable.redirect.guest.to.login:false}" />
    </bean>
```

So I could purpose a PR with removing the property setted into `security.properties`, what do you think ?

drewwills:
So I could purpose a PR with removing the property setted into security.properties, what do you think ?

We need to ask ourselves:  "what's the best way to set up properties that have defaults, so that someone who may want to change the value of the property is most likely to discover it?"

I believe the presence of the property in `security.properties` is only for the sake of discoverability.  Often we have included these sorts of properties (i.e. with defaults defined in `@Value` annotations) in the properties files anyway, with comments, and commented-out.

Perhaps that's the best approach.  I don't think someone looking to configure uPortal will comb through .java files to figure out the options, but s/he may look at .properties files.  (And of course we should add it to the documentation.)

jgribonvald:
> ```
> Perhaps that's the best approach. I don't think someone looking to configure uPortal will comb through .java files to figure out the options, but s/he may look at .properties files. (And of course we should add it to the documentation.)

From my point of view removing the property from security.properties is good if we document on how to activate the mode !

jgribonvald:
A little reminder to reviewers.
I know you had thanksgiving so you had some vacations :wink:

drewwills:
I'm looking at this item now.

When I build this branch and use it in uPortal-start, the `cas.enable.redirect.guest.to.login` property correctly sends me to the login form instead of the guest layout.

But when I try to authenticate with CAS, I get the following...

cas-403

drewwills:
There's another potential problem (related to this one) brewing here:  https://github.com/Jasig/uPortal/blob/ae8be7877ad9e0f82fd23335c510ae1cff9f0308/uPortal-security/uPortal-security-core/src/main/java/org/apereo/portal/security/oauth/IdTokenFactory.java#L59

mrapczynski:
@drewwills Assuming the issue here is what happens when the value not supplied? I dealt with something similar in another project where I am using Spring. Those `@Value` annotations support a default/fallback. Would it be sensible to add a default value, yet still leaving open the option for the implementer to customize?

> There's another potential problem (related to this one) brewing here:
> 
> [uPortal/uPortal-security/uPortal-security-core/src/main/java/org/apereo/portal/security/oauth/IdTokenFactory.java](https://github.com/Jasig/uPortal/blob/ae8be7877ad9e0f82fd23335c510ae1cff9f0308/uPortal-security/uPortal-security-core/src/main/java/org/apereo/portal/security/oauth/IdTokenFactory.java#L59)
> 
> Line 59 in [ae8be78](/Jasig/uPortal/commit/ae8be7877ad9e0f82fd23335c510ae1cff9f0308)
>  @Value("${portal.protocol.server.context}")

drewwills:
@mrapczynski -- Specifying defaults on `@Value` annotations is a very cool feature that definitely helps us out of a lot of tight spots.

There's probably more we could do with that feature in this area, but I think the heart of the matter is that a custom (roll-your-own) replacement token -- `_CURRENT_SERVER_NAME_` -- is being used in is a property that feeds several Spring EL different expressions.  Spring isn't aware of it, so it can't help us.

drewwills:
Folks -- please take a look at #1470.

jgribonvald:
> I'm looking at this item now.
> 
> When I build this branch and use it in uPortal-start, the `cas.enable.redirect.guest.to.login` property correctly sends me to the login form instead of the guest layout.
> 
> But when I try to authenticate with CAS, I get the following...
> 
> cas-403

You get this issue due to encodeServiceUrl=true ! If you watch better you will see the error is on the serviceUrl, in one case it's encoded in the other not, this depends on the client cas and cas server version sometimes you should activate it sometimes not. If you watch on the other CAS defintion bean you will see that this option is also set to false ;)

Sorry I didn't change this value because on my server CAS version there isn't problem !

jgribonvald:
> There's another potential problem (related to this one) brewing here:
> 
> [uPortal/uPortal-security/uPortal-security-core/src/main/java/org/apereo/portal/security/oauth/IdTokenFactory.java](https://github.com/Jasig/uPortal/blob/ae8be7877ad9e0f82fd23335c510ae1cff9f0308/uPortal-security/uPortal-security-core/src/main/java/org/apereo/portal/security/oauth/IdTokenFactory.java#L59)
> 
> Line 59 in [ae8be78](/Jasig/uPortal/commit/ae8be7877ad9e0f82fd23335c510ae1cff9f0308)
>  @Value("${portal.protocol.server.context}")

About this potential issue I'm open to talk and I don't know how to manage it as I don't know the usage ! I don't know the goal a this property in the OIDC as any request it already done on the good URL. =>  I don't know where and on what it could be used.

ChristianMurphy:
@drewwills @jgribonvald any decision on whether to move forward with this or #1470?

ChristianMurphy:
Thanks @jgribonvald and @bjagg! :bowing_man:

Issue #1448, opened 2018-11-28
Abusive z-index cleaning requested


jgribonvald:
**Is your feature request related to a problem? Please describe.**
When integrating web-components that overlap some part of the UI I'm facing to some undesirable z-index, and I thing somes are abusive. And so I need sometimes to up my z-index value from code, or to override it to the uPortal defaultSkin. (somes are up to 1000, 900, 800)

**Describe the solution you'd like**
Some cleaning would be appreciated, and when a such css property is added a comment justifying it should be mandatory (something that say: we need it to overlap this css - css property + css file and line)

jgribonvald:
here are all z-index used (except from lib): 
```
uPortal-webapp/build/webjars/bootstrap-3.3.7/META-INF/resources/webjars/bootstrap/3.3.7/less/variables.less:@zindex-dropdown:          1000;
uPortal-webapp/build/generated-sources/skin/main/webapp/webjars/bootstrap/less/variables.less:@zindex-dropdown:          1000;

uPortal-webapp/src/main/webapp$ grep -R --line-number "z-index" * | grep -v "build" | grep -v ".gradle" | grep -v "node_modules"
media/skins/common/javascript/uportal/flyout-nav.js:94:     * Provide fix for z-index layering issues in older IE browsers
media/skins/respondr/common/less/entity-selector.less:295:    z-index: +1;
media/skins/respondr/common/less/regions.less:70:        z-index: 1000;
media/skins/respondr/common/less/regions.less:240:        z-index: 1;
media/skins/respondr/common/less/regions.less:243:            z-index: 999;
media/skins/respondr/common/less/regions.less:262:            z-index: 800;
media/skins/respondr/common/less/regions.less:302:            z-index: auto;
media/skins/respondr/common/less/tags.less:77:                   z-index:1;
media/skins/respondr/common/less/tags.less:80:                      z-index:4 !important;
media/skins/respondr/common/less/content.less:28:  /* Do not change this z-index value without side effects on bootstrap modal for rate this portlet see comments and further references in defaultSkin/skin.less */
media/skins/respondr/common/less/content.less:29:  z-index: auto;
media/skins/respondr/common/less/jsr-168.less:271:        z-index: @zindex-dropdown;
media/skins/respondr/common/less/portlet.less:68:        z-index: 1000;
media/skins/respondr/common/less/header.less:34:        z-index: 1;
media/skins/respondr/common/less/header.less:74:        z-index: 10;
[...]

PR #1449, opened 2018-11-28
Fix: on z-index

jgribonvald:
##### Checklist

-   [ ] tests are included

##### Description of change

These change follow issue #1448 

The first commit it sure but the second should be tested on different browsers (I need help for that).

### DIFF
--- a/uPortal-webapp/src/main/webapp/WEB-INF/jsp/BackgroundPreference/viewBackgroundPreference.jsp
+++ b/uPortal-webapp/src/main/webapp/WEB-INF/jsp/BackgroundPreference/viewBackgroundPreference.jsp
@@ -52,7 +52,8 @@
         position:absolute;
         overflow-x: hidden;
         overflow-y: scroll;
-        z-index: 10000;
+        /* Needed as dropdown-menu - based on bootstrap variable @zindex-dropdown */
+        z-index: 1000;
         text-align: center;
         max-height: 515px;
     }
--- a/uPortal-webapp/src/main/webapp/media/skins/respondr/common/less/header.less
+++ b/uPortal-webapp/src/main/webapp/media/skins/respondr/common/less/header.less
@@ -163,7 +163,7 @@ h1.portal-logo a:hover {
 
 #up-notification {
   position: fixed;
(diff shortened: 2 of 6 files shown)


uPortal-webapp/src/main/webapp/WEB-INF/jsp/BackgroundPreference/viewBackgroundPreference.jsp
```diff
@@ -52,7 +52,8 @@
         position:absolute;
         overflow-x: hidden;
         overflow-y: scroll;
-        z-index: 10000;
+        /* Needed as dropdown-menu - based on bootstrap variable @zindex-dropdown */
+        z-index: 1000;
```
ChristianMurphy: Could this directly use `@zindex-dropdown`?

jgribonvald: It's into a jsp, so it won't be possible as this var can be applied on css build
My idea was more to apply a dropdown css class (reusing uPortal skin one), but there is too much to check ...

ChristianMurphy: Outside of https://github.com/Jasig/uPortal/blob/b3315872cb9c5b5b578934c0b1e612b1e5875de0/uPortal-webapp/src/main/webapp/WEB-INF/jsp/BackgroundPreference/viewBackgroundPreference.jsp#L32-L37
It looks like all of this could be moved to less.
But that would be for another PR.

jgribonvald: Ok I see, yes this could be moved into less files.
But I don't understand the need to have portlet execution id into css tags, is there an interest ? (it's for js only when querying a dom element)

jgribonvald:
the first commit is safe and can be integrated in my mind, but the second should be tested (by critical adopters would help). Maybe the best would be to purpose two pull request (one per commit) ?

PR #1462, opened 2018-12-07
feat: extract favicon path as property to be able to customize it wit…

jgribonvald:
…hout modifying it into the xsl

##### Checklist

-   [ ] tests are included

##### Description of change

This permit to extract as a property the favicon path and avoid to modify it into the xsl (and so avoid having into uPortal-start a copy of the xsl only for this change)

### DIFF
--- a/uPortal-webapp/src/main/java/org/apereo/portal/context/rendering/RenderingPipelineConfiguration.java
+++ b/uPortal-webapp/src/main/java/org/apereo/portal/context/rendering/RenderingPipelineConfiguration.java
@@ -146,6 +146,9 @@ public class RenderingPipelineConfiguration {
     @Value("${org.apereo.portal.layout.useFlyoutMenus:false}")
     private String useFlyoutMenus;
 
+    @Value("${org.apereo.portal.layout.faviconPath:#{null}}")
+    private String faviconPath;
+
     @Resource(name = "org.apereo.portal.rendering.THEME_TRANSFORM")
     private Cache themeTransformCache;
 
@@ -411,6 +414,9 @@ public class RenderingPipelineConfiguration {
         parameters.put("useTabGroups", useTabGroups);
         parameters.put("UP_VERSION", uPortalVersion);
         parameters.put("USE_FLYOUT_MENUS", useFlyoutMenus);
+        if (faviconPath != null) {
+            parameters.put("PORTAL_SHORTCUT_ICON", faviconPath);
+        }
 
         final Map<String, String> parameterExpressions = new HashMap<>();
         parameterExpressions.put("CURRENT_REQUEST", "request.nativeRequest");
--- a/uPortal-webapp/src/main/resources/layout/theme/respondr/respondr.xsl
+++ b/uPortal-webapp/src/main/resources/layout/theme/respondr/respondr.xsl
@@ -149,7 +149,8 @@
 <xsl:variable name="ABSOLUTE_MEDIA_PATH" select="concat($CONTEXT_PATH,'/',$MEDIA_PATH)"/>
 <xsl:variable name="SKIN_RESOURCES_PATH" select="concat('/',$MEDIA_PATH,'/',$SKIN,'/skin.xml')"/>
 <xsl:variable name="SKIN_PATH" select="concat($ABSOLUTE_MEDIA_PATH,'/',$SKIN)"/>
-<xsl:variable name="PORTAL_SHORTCUT_ICON" select="concat($CONTEXT_PATH,'/favicon.ico')" />
+<xsl:variable name="SHORTCUT_ICON" select="concat($CONTEXT_PATH,'/favicon.ico')" />
+<xsl:param name="PORTAL_SHORTCUT_ICON"><xsl:value-of select="$SHORTCUT_ICON"/></xsl:param>
 <xsl:variable name="FOCUSED_CLASS">
     <xsl:choose>
         <xsl:when test="//content/focused">focused <xsl:value-of select="//content/focused/channel/@fname"/></xsl:when>


jgribonvald:
It permit to set a custom path of the favicon. By default uPortal provide his favicon, but in a customization case you could need to change it to a custom one (in my case it comes from the servername). In this case, and before the change, you need to modify the path into the xsl manually, and so you will need to add the xsl to your uPortal-start and anymore you won't be able to follow xsl change (if there is).

uPortal-webapp/src/main/resources/properties/portal.properties
```diff
@@ -457,6 +457,13 @@ org.apereo.portal.layout.useTabGroups=false
 ##
 org.apereo.portal.layout.useFlyoutMenus=false
 
+
+##
+## Usefull to apply a customized favicon.
+## as example on https://my.univ.edu/favicon.ico set the property to /favicon.ico
+##
+org.apereo.portal.layout.faviconPath=${portal.context}/favicon.ico
```
drewwills: Here I think this property should be commented-out by default (or omitted entirely, if we think it will be a seldom-used option).

The `uPortal.properties`, `global.properties`, _etc._ files are supposed to be optional.  The way it's set up now, I think the Spring context won't load if this property is not defined.

uPortal-webapp/src/main/java/org/apereo/portal/context/rendering/RenderingPipelineConfiguration.java
```diff
@@ -146,6 +146,9 @@
     @Value("${org.apereo.portal.layout.useFlyoutMenus:false}")
     private String useFlyoutMenus;
 
+    @Value("${org.apereo.portal.layout.faviconPath}")
```
drewwills: I suggest a default here, like... `@Value("${org.apereo.portal.layout.faviconPath:/favicon.ico}")`.

(More on this suggestion below.)

uPortal-webapp/src/main/resources/properties/portal.properties
jgribonvald: Yes if the property isn't set it won't load.

uPortal-webapp/src/main/java/org/apereo/portal/context/rendering/RenderingPipelineConfiguration.java
jgribonvald: I didn't do that to avoid to apply the Spel, as the property (if we want to keep the actual path) depends as we should add the app context like : ${portal.context}/favicon.ico

uPortal-webapp/src/main/resources/layout/theme/respondr/respondr.xsl
```diff
@@ -149,7 +149,8 @@
 <xsl:variable name="ABSOLUTE_MEDIA_PATH" select="concat($CONTEXT_PATH,'/',$MEDIA_PATH)"/>
 <xsl:variable name="SKIN_RESOURCES_PATH" select="concat('/',$MEDIA_PATH,'/',$SKIN,'/skin.xml')"/>
 <xsl:variable name="SKIN_PATH" select="concat($ABSOLUTE_MEDIA_PATH,'/',$SKIN)"/>
-<xsl:variable name="PORTAL_SHORTCUT_ICON" select="concat($CONTEXT_PATH,'/favicon.ico')" />
+<xsl:variable name="SHORTCUT_ICON" select="concat($CONTEXT_PATH,'/favicon.ico')" />
+<xsl:param name="PORTAL_SHORTCUT_ICON"><xsl:value-of select="$SHORTCUT_ICON"/></xsl:param>
```
drewwills: To avoid needing sometthing like `@Value("${org.apereo.portal.layout.faviconPath:${portal.context}/favicon.ico}")` (which won't work, I think), how about...

```
<xsl:variable name="SHORTCUT_ICON">/favicon.ico</xsl:variable>
<xsl:param name="PORTAL_SHORTCUT_ICON" select="concat($CONTEXT_PATH, $SHORTCUT_ICON)"/>
```

uPortal-webapp/src/main/resources/properties/portal.properties
jgribonvald: as it was set in `portal.properties` you don't need to set it into `uPortal.properties` and `global.properties`... so the property is initialized in the default conf.

On an other side I modified some parts -> the property is needed only if you want to modify the favorite.ico path, in this case you will have only to set `org.apereo.portal.layout.faviconPath` into `uPortal.properties` to your value.

uPortal-webapp/src/main/java/org/apereo/portal/context/rendering/RenderingPipelineConfiguration.java
jgribonvald: After change the default value is null and so the parameter is passed to the xsl only if set from `uPortal.properties` (or `global.properties`)

uPortal-webapp/src/main/resources/layout/theme/respondr/respondr.xsl
jgribonvald: There isn't more needs to modify anything.

uPortal-webapp/src/main/resources/layout/theme/respondr/respondr.xsl
```diff
@@ -149,7 +149,8 @@
 <xsl:variable name="ABSOLUTE_MEDIA_PATH" select="concat($CONTEXT_PATH,'/',$MEDIA_PATH)"/>
 <xsl:variable name="SKIN_RESOURCES_PATH" select="concat('/',$MEDIA_PATH,'/',$SKIN,'/skin.xml')"/>
 <xsl:variable name="SKIN_PATH" select="concat($ABSOLUTE_MEDIA_PATH,'/',$SKIN)"/>
-<xsl:variable name="PORTAL_SHORTCUT_ICON" select="concat($CONTEXT_PATH,'/favicon.ico')" />
+<xsl:variable name="SHORTCUT_ICON" select="concat($CONTEXT_PATH,'/favicon.ico')" />
```
drewwills: I like your change to `RenderingPipelineConfiguration`.

It doesn't look like the extra `xsl:variable` (`SHORTCUT_ICON`) is needed any longer.

jgribonvald: The SHORTCUT_ICON is needed for the default value when the property isn't set

drewwills:
@jgribonvald -- are you okay with #1471?

PR #1468, opened 2018-12-12
Issue #1467:  deprecate the ThemeNameEqualsIgnoreCaseTester and provi…

### LINKED ISSUE #1467: Guest user occasionally receives a portal page without a skin when using the dynamic-respondr-skin portlet
## Describe the bug

Occasionally (but typically uncommonly) unauthenticated users will start receiving a portal page that does not include the `dynamic-respondr-skin` portlet and therefore has no skin.  The affected page is completely unusable.

This issue has been encountered by several, and has been discussed more than once on the `uportal-user` list:  https://groups.google.com/a/apereo.org/forum/#!topic/uportal-user/JBy0Mp7YvG0

The issue stems from the `ThemeNameEqualsIgnoreCaseTester`.  This PAGS tester requires an active `HttpServletRequest` to function, otherwise it always returns `false` (i.e. the user is not a member of the group).  The problem is that this tester can be invoked within the Event Aggregation processes, which happen on a worker (non-request) thread.

In a typical data set (since the portlet was introduced), users receive the `dynamic-respondr-skin` portlet by virtue of their membership in the `Respondr Theme Users` group -- a PAGS group based on the `ThemeNameEqualsIgnoreCaseTester`.

Membership evaluations are cached in the portal.  When a test for membership in the `Respondr Theme Users` group for the `guest` user returns `false`, that result will be cached for a period.  During that period, real guest users will not receive the `dynamic-respondr-skin` portlet.

## How To Fix

Clearly the `ThemeNameEqualsIgnoreCaseTester` has issues, and (furthermore) we don't need it the way we did when it was introduced.  (The Universality theme has been removed.)  We should deprecate it, remove all usages of it from data sets in uPortal-start, and publish the solution in the released notes.

drewwills:
…de a more specific warning (it should be avoided)

resolves #1467 

##### Checklist

##### Description of change

The ThemeNameEqualsIgnoreCaseTester causes problems;  see #1467.

### DIFF
--- a/uPortal-events/src/main/java/org/apereo/portal/events/aggr/PortalEventProcessingManagerImpl.java
+++ b/uPortal-events/src/main/java/org/apereo/portal/events/aggr/PortalEventProcessingManagerImpl.java
@@ -63,7 +63,7 @@ public class PortalEventProcessingManagerImpl
             new ThreadLocal<Map<Class<?>, Collection<Serializable>>>() {
                 @Override
                 protected Map<Class<?>, Collection<Serializable>> initialValue() {
-                    return new HashMap<Class<?>, Collection<Serializable>>();
+                    return new HashMap<>();
                 }
             };
     private volatile boolean shutdown = false;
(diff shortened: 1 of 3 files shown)


drewwills:
See also https://github.com/Jasig/uPortal-start/pull/256.

PR #1470, opened 2018-12-14
[WIP] Issue #1409:  [WIP] support portal.allServerNames property in a safe way

### LINKED ISSUE #1409: Property 'cas.enable.redirect.guest.to.login' not working because of _CURRENT_SERVER_NAME_ token

drewwills:
Resolves #1409 

##### Checklist

##### Description of change

I could really use your help on this one @jgribonvald.  ;-)

This PR is another take on issue #1409 and pull request #1410.

The key difference in this PR -- and the essence of the issue, I think -- is that I feel we should avoid having the token `_CURRENT_SERVER_NAME_` (or any such token) inside a Spring-managed property.

The token replacement tech is 100% custom.  Spring doesn't know about it, so it faithfully inserts the original property value (incl. the token) everywhere the property is referenced.

All of the following have multiple uses in the portal:
  - `portal.protocol.server.context`
  - `portal.login.url`
  - `org.apereo.portal.channels.CLogin.CasLoginUrl`

Putting `_CURRENT_SERVER_NAME_` into any of them (inside security.properties) will result in the unprocessed token getting used somewhere in the portal.

This PR updates the `UrlMultiServerNameCustomizer`, attempting to parse the URL being processed into its bits (`host`, `port`, `path`, _etc._) and replace some of those bits in the intended way .

It's not entirely working ATM.  All the tests (3) are broken (`CasLoginRefUrlEncoderTest`).  One seems to be only an encoding issue;  the other two don't get the `hostname` they expect.

### DIFF
--- a/uPortal-security/uPortal-security-mvc/src/test/java/org/apereo/portal/url/CasLoginRefUrlEncoderTest.java
+++ b/uPortal-security/uPortal-security-mvc/src/test/java/org/apereo/portal/url/CasLoginRefUrlEncoderTest.java
@@ -28,13 +28,12 @@ import org.springframework.mock.web.MockHttpServletRequest;
 public class CasLoginRefUrlEncoderTest {
 
     CasLoginRefUrlEncoder encoder;
-    MockHttpServletRequest request;
 
     @Before
     public void setUp() throws UnsupportedEncodingException {
         encoder = new CasLoginRefUrlEncoder();
         encoder.setCasLoginUrl(
-                "https://cas:80/cas/login?service=_CURRENT_SERVER_NAME_/uPortal/Login");
+                "https://cas:80/cas/login?service=https://testhost:8080/uPortal/Login");
         UrlMultiServerNameCustomizer urlCustomizer = new UrlMultiServerNameCustomizer();
         urlCustomizer.setAllServerNames(Sets.newHashSet("myhost:8080", "theirhost:8443"));
         List<IAuthUrlCustomizer> customizers = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


jgribonvald:
I understand the problem but I'm against providing a default secure value as anybody won't be able to see/obtain a problem when a change will be done, or we should find a way to test it correctly with a CAS server on all side/request it could be needed. 

I provided the feature on this way to provide a real view of the use and to be able to check if it work on all change case. 
Also **I can assume some effort to keep it working**, just provide me an issue, the use case, and condition to reproduce the problem - the best would be to provide a unit test to reproduce the problem. And on case that I didn't manage something like OIDC we should talk of what is desired and I can tell how to solve it !

drewwills:
Folks,

In my case, I think I could be a lot more effective contributing to a good solution if I understood the new feature better.

Help me remember/understand more thoroughly:  what is the problem the `_CURRENT_SERVER_NAME_` change is trying to solve? 

  - We need one portal deployment to support authentication with multiple CAS deployments (I hope it's this one)
  - We need one CAS deployment to support authentication with multiple portal deployments (probably not this one... this one is easy)
  - We need one portal deployment to operate as several different hostnames (I hope it's _not_ this one;  this one sounds hard)

The  `portal.server` and `portal.protocol.server.context` properties are used in *a lot* of places, and some adopters use them in additional ways that aren't currently configured in uPortal-start.  I would like to keep these properties simple... 100% managed by Spring, and predictable.

From looking around yesterday, it appears that the important properties for this feature are `org.apereo.portal.channels.CLogin.CasLoginUrl` and `logout.redirect`.  They both include a query string parameter containing a URL that points back to the portal.  They aren't used very often.  I think we could decorate these properties with beans that contain logic to rewrite them in custom ways.

jgribonvald:
Drew the `_CURRENT_SERVER_NAME_ ` permit to manage several public uPortal hostname on the same uPortal instance (don't need more than that to work). And it's working really well on my case. If you need we can talk more about it on hangout.
This permit to set the current url from current user request into CAS serviceValidate URL into the service url param. It's important to get back the user to the good hostname whereas the session cookie won't be on the good hostname...

So this doesn't permit to manage several CAS server in CAS authentications, for that you should have  a WAYF like page to ask the user "where to authenticate". For more details we can talk about what I did (also I should work on CAS server for a big upgrade when back from christmas hollydays)

jgribonvald:
Also this is the proff that I should provide a documentation on how it's working and on how to use it. And surely present the use case !

ChristianMurphy:
@drewwills @jgribonvald any decision on whether to move forward with this or #1410?

jgribonvald:
#2525 fix the last remaining problem from OIDC. Like that the OIDC server can be multi-tenant URL (several server name on tenant)

bjagg:
@jgribonvald should we drop this PR now?

jgribonvald:
@bjagg yes you can as all was merged

PR #1759, opened 2019-06-26
feat(jwt): Add option to turn off JWT encryption

### LINKED ISSUE #1743: Add option to turn off JWT encryption
**Is your feature request related to a problem? Please describe.**
Development can be facilitated if JWT are not encrypted.

**Describe the solution you'd like**
Add a configuration option to turn off encryption of JWT (tokens) with a default of encrypt for backwards compatibility. This will include adding default configuration values, optionally signing JWT based on configuration, and documentation.

bjagg:
Resolves #1743

##### Checklist

-   [x] tests are included

##### Description of change


bjagg:
So Soffits encrypt their JWTs by default while OIDC JWTs never supported encryption. This PR will attempt to check if a JWT is encrypted in both places; however, it only encrypts in Soffit code. Should this flag also signal encryption in OIDC?

nickwoodle:
@bjagg Just curious why you are using regex to determine if the JWT is encrypted? Couldn't create a Boolean variable (on/off switch) and then check if true, then decrypt, else don't decrypt?

amgoodfellow:
Following up a bit on what @nickwoodle was talking about, I'm also kind of curious about what exactly we gain by using regex here.

I'm not sure if I'm reading everything correctly, but it looks to me that the `JWT_PATTERN` is trying to determine whether or not the provided string is a valid JWT. Then, if it's *not* a valid JWT it assumes that it's an encrypted JWT? Again, it could be I'm just reading this wrong, but that doesn't seem like a good assumption to make.

I'll second @nickwoodle 's idea about the logic looking something like:

```java
if (encryptJwtYN) {
   jwt = textEncryptor.decrypt(jwt)
}
  return Jwts.parser().setSigningKey(signatureKey).parseClaimsJws(jwt);
}
```

That way, if the `encryptJwtYN` flag is set to `true`, it will assume that your jwt strings are encrypted and throw an `EncryptionOperationNotPossibleException`. This is consistent with the `AbstractJwtService`'s practice of throwing `SecurityExceptions` if the expiration date has already passed. 

And I think there's an added benefit to this simplification - it removes the need for the new `JwtEncryptor.java`. Looking at the functionality of that class, the init method checks the encryption password, and there's two methods that call Jasypt's `BasicTextEncryptor`. I'd argue this might be able to stay in the AbstractJwtService's `parseEncryptedToken` method (although it might need to be renamed for clarity).

uPortal-soffit/uPortal-soffit-core/src/test/java/org/apereo/portal/soffit/service/JwtEncryptorTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import java.util.regex.Matcher;
+import org.junit.Test;
+
+public class JwtEncryptorTest {
+
+    private static final String VALID_JWT = "sdfSDFSD4345.sdfsdf2342534fsadf/++.sdf3w4tvav==";
```
amgoodfellow: Is this a valid JWT?
jwt.io doesn't seem to think so. It might be a good idea to test using the default one generated from that site.

bjagg:
> @bjagg Just curious why you are using regex to determine if the JWT is encrypted? Couldn't you create a Boolean variable (on/off switch) and then check if true, then decrypt, else don't decrypt?

The issue is that OIDC code never supported encryption. So there is currently a chance that even when the encrypt flag is set, a token may not be encrypted.

It's a concern tbh

uPortal-soffit/uPortal-soffit-core/src/test/java/org/apereo/portal/soffit/service/JwtEncryptorTest.java
bjagg: It is not actually a valid ticket that can be parsed. I used this as the BASE64 legal characters are `[a-z][A-Z][0-9]+/=`

Good idea!

bjagg:
> And I think there's an added benefit to this simplification - it removes the need for the new JwtEncryptor.java

I like the externalization of this into a separate class since encryption support is still incomplete in JJWT. It looks like an upcoming feature. Would like to change it in this one class rather than the Soffit and OIDC classes.

nickwoodle:
> So there is currently a chance that even when the encrypt flag is set, a token may not be encrypted.

Maybe I dont know enough about the OIDC code. However, I would think if the flag is being used to identify when the JWT is encrypted and decrypted then it would work. I personally feel like this would be better then regex, just my opinion. With regex there is always the possibility that something being missed and it ends up causing problems.

uPortal-soffit/uPortal-soffit-core/src/main/java/org/apereo/portal/soffit/service/JwtEncryptor.java
```diff
@@ -0,0 +1,97 @@
...
+        if (encryptJwtYN) {
+            return textEncryptor.encrypt(jwt);
+        } else {
+            return jwt;
+        }
+    }
+
+    public String decryptIfTokenEncrypted(String jwt) {
+        if (encryptJwtYN) {
+            Matcher m = JWT_PATTERN.matcher(jwt);
```
ChristianMurphy: Does the JWT library used in uPortal include a util function that checks if a string is a JWT?

bjagg: I did not find a format validator; however, I did look at their code to see how the validate the format. This whole issue was due to this JJWT code throwing an exception because there were not 2 `.`s in the string.

ChristianMurphy: Would there be any negatives to going with something like
```
try {
  decodeJWT
} catch {
  decryptJWT
  decodeJWT
}
```

to avoid maintaining a separate regex?

uPortal-soffit/uPortal-soffit-core/src/main/java/org/apereo/portal/soffit/service/JwtEncryptor.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.jasypt.util.text.BasicTextEncryptor;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+/**
+ * Wrapper for code to encrypt and decrypt JWT tokens. The spec is evolving, so this implementation
+ * is expected to change.
+ *
+ * @since 5.6.1
```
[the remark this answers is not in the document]
bjagg: If we define the uPortal "API" to include configuration options, then I agree. And I do think we consider configuration "API". Yes, this next release will be 5.7.0.

uPortal-soffit/uPortal-soffit-core/src/main/java/org/apereo/portal/soffit/service/JwtEncryptor.java
bjagg: Yes, 1) depending on an exception when a known condition is expected is a bad practice, 2) the regex follows a well-defined spec. Also this special handling will go away in the next major release. It's here to deal with Soffit and OIDC using the same header for their potentially different tokens.

ChristianMurphy: > It's here to deal with Soffit and OIDC using the same header for their potentially different tokens.

If OIDC and Soffit are using different header handling mechanisms, should this class get split, one for soffit, one for OIDC?

> special handling will go away in the next major release

My caution is without a comment or having a separate class, it would be easy to forget and miss that this should be removed in the next release.

ChristianMurphy: > depending on an exception when a known condition is expected is a bad practice

It probably emits a known exception.
Would

```java
try {
  decodeJWT(); // might be encrypted, if so it will throw
} catch (JwtDecodeError e) { // catch only decode exception
  decryptJWT();
  decodeJWT();
} catch (Error e) {
  throw e; // rethrow other errors
}
```

bjagg:
I think there might be a misconception and something I may be conflating.

JWTs in AUTHORIZATION headers can come from various sources and are used for other things besides soffits. uPortal can process tokens from external sources with the same signing key whereas those sources may not be encrypting the tokens. So there is a need to check even if the encryption flag is set to true.

But I do see the point in the Soffit code to always assume the flag is correct for tokens. So should I create a method that skips the regex matching that can be used by `AbstractJwtService.java`?

ChristianMurphy:
> uPortal can process tokens from external sources with the same signing key whereas those sources may not be encrypting the tokens. So there is a need to check even if the encryption flag is set to true.

Can we accurately differentiate between soffit and non-soffit sources?
If so couldn't the soffit sources always be run through decrypt + decode while the non-soffit sources only are run through decode?

> So should I create a method that skips the regex matching that can be used by AbstractJwtService.java?

This approach sounds promising.

amgoodfellow:
@bjagg 
I know that this is already merged in, but I thought I ought to voice my last remaining concern on behalf of OU.

I appreciate the separation suggested by @ChristianMurphy for separate methods, but I'm still a bit skeptical about the usage of regular expressions to perform validation. They seem to be a bit lower level than we need, you know? We're not trying to build a JWT parsing library ourselves, so we ought to rely on more thoroughly vetted libraries for these sorts of things. 

Although they do follow a well-defined spec, regular expressions this large aren't exactly 
intuitive. And it's easy to miss things in a long spec sheet like the JWT RFC. For example, 
the definition of `Base64URL encoding` according to RFC7519 (referenced in the code), disallows all 
trailing `=` characters, which this regex accepts. My point isn't really to nitpick implementation details, and in general, I really like regular expressions. I just think in this case, we should probably try validation with an established library and catch errors if JWTs turn out to be invalid.

bpowell:
@bjagg long time no chat! I highly suggest not using regex to validate something like this. 
```
Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems.
```
The idea that @ChristianMurphy suggested to try to decode and if that throws an exception, try to decrypt the token is probably the best option here. It saves you from any regex issues in the future while still being safe.

PR #1761, opened 2019-06-26
feat(local login): Add a warning to the logs when an attempt to use local login occurs

### LINKED ISSUE #1760: Add a warning to the logs when an attempt to use local login occurs
Local login for uPortal is turned on by default and allows for someone to bypass CAS and log in using the example logins. This would not prevent someone from doing that, rather it adds a warning to the logs so that it could be used to identify an attempt.

nickwoodle:
Resolves #1760

##### Checklist

-   [x] the [individual contributor license agreement](https://github.com/Jasig/uPortal/blob/master/CONTRIBUTING.md#individual-contributor-license-agreement) is signed
-   [x] commit message follows [commit guidelines](https://github.com/Jasig/uPortal/blob/master/CONTRIBUTING.md#commit)

##### Description of change
Adds a simple warning log when someone tries to use the local login.

### DIFF
--- a/uPortal-security/uPortal-security-authn/src/main/java/org/apereo/portal/security/provider/SimpleSecurityContext.java
+++ b/uPortal-security/uPortal-security-authn/src/main/java/org/apereo/portal/security/provider/SimpleSecurityContext.java
@@ -54,10 +54,10 @@ public class SimpleSecurityContext extends ChainingSecurityContext implements IS
         if (this.myPrincipal.UID != null && this.myOpaqueCredentials.credentialstring != null) {
 
             // Logs if an attempt is made to log into a local account
-            StringBuilder msg = new StringBuilder();
-            msg.append("An attempt to log into the local login has occurred. ");
-            msg.append("user=" + this.myPrincipal.UID);
-            if (log.isWarnEnabled()) log.warn(msg.toString()); 
+            if (log.isWarnEnabled())
+                log.warn(
+                        "An attempt to log into the local login has occurred. user="
+                                + this.myPrincipal.UID);
 
             try {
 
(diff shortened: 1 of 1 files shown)


uPortal-security/uPortal-security-authn/src/main/java/org/apereo/portal/security/provider/SimpleSecurityContext.java
```diff
@@ -53,6 +53,12 @@ public synchronized void authenticate() throws PortalSecurityException {
         this.isauth = false;
         if (this.myPrincipal.UID != null && this.myOpaqueCredentials.credentialstring != null) {
 
+            // Logs if an attempt is made to log into a local account
+            StringBuilder msg = new StringBuilder();
+            msg.append("An attempt to log into the local login has occurred. ");
+            msg.append("user=" + this.myPrincipal.UID);
+            if (log.isWarnEnabled()) log.warn(msg.toString()); 
+
```
bjagg: I would wrap the StringBuilder in the if statement.

uPortal-security/uPortal-security-authn/src/main/java/org/apereo/portal/security/provider/SimpleSecurityContext.java
```diff
@@ -53,6 +53,12 @@ public synchronized void authenticate() throws PortalSecurityException {
         this.isauth = false;
         if (this.myPrincipal.UID != null && this.myOpaqueCredentials.credentialstring != null) {
 
+            // Logs if an attempt is made to log into a local account
+            StringBuilder msg = new StringBuilder();
+            msg.append("An attempt to log into the local login has occurred. ");
+            msg.append("user=" + this.myPrincipal.UID);
```
bjagg: Advantage of String buffer is to avoid string concatenation. Either move the "user=" text to the end of the previous string so that this becomes `msg.append(this.myPrincipal.UID);` or drop the whole StringBuffer var and just use `final String msg = "..." + this.myPrincipal.UID;`. The later only makes since since the message is only concatenating strings once.

bjagg:
Build likely failed because of formatting. Run `./gradlew goJF` to format source.

uPortal-security/uPortal-security-authn/src/main/java/org/apereo/portal/security/provider/SimpleSecurityContext.java
nickwoodle: Thanks for the suggestion. I moved the message into the if statement and cleaned up the code. It now looks like the rest of the logging in this file. I originally built it so that if anyone wanted to add additional things to the log message (i.e. IP or something of that nature) they could.

PR #1763, opened 2019-06-27
feat(props): Add support to overwrite properties based on a UP_CLUSTE…

### LINKED ISSUE #1747: Enhance properties loader class to honor env prefixes
**Is your feature request related to a problem? Please describe.**
Some Operations teams prefer to have a single set of files to track over sets of files per environment.

**Describe the solution you'd like**
Allow for an environment system value that define the current "cluster" (i.e. test, test2, prod) to activate the current values for some keys, such as CAS host.

For example, given:
1. `UP_CLUSTER=test2` as a system environment variable
2. `test.cas.server`, `test2.cas.server` and `prod.cas.server` defined in `uPortal.properties`

In Spring contexts that load `uPortal.properties`, the value for `cas.server` should equal `test2.cas.server`.

bjagg:
…R env property

Resolves #1747

##### Checklist

-   [X] tests are included

##### Description of change

Client wished to place all properties in a single set of properties files that would work in test, and prod. To support this, a system environment variable `UP_CLUSTER` can be set. If properties are found that start with that value plus `.`, then those key/value pairs will be added back, overwriting existing values, without the prefix.

For example if `UP_CLUSTER=prod` is set in the System environment, an initial properties set of:

```
url=www.github.com
dev.url=github.com
prod.url=www.mycompany.com
```
will then become the following, post-processed:

```
url=www.mycompany.com
dev.url=github.com
prod.url=www.mycompany.com
```
Also, `-Dcluster=prod` added to `CATALINA_OPTS`, `JAVA_OPTS`, etc. will also work.

N.B. `UP_CLUSTER` takes precedent over the `-Dcluster` JVM argument.

### DIFF
--- a/uPortal-spring/build.gradle
+++ b/uPortal-spring/build.gradle
@@ -20,4 +20,6 @@ dependencies {
 
     testCompile "${portletApiDependency}"
     testCompile "${servletApiDependency}"
-}
+
+    compileOnly "org.projectlombok:lombok:${lombokVersion}"
+    annotationProcessor "org.projectlombok:lombok:${lombokVersion}"}
--- a/uPortal-spring/src/main/java/org/springframework/context/support/PortalPropertySourcesPlaceholderConfigurer.java
+++ b/uPortal-spring/src/main/java/org/springframework/context/support/PortalPropertySourcesPlaceholderConfigurer.java
@@ -17,11 +17,12 @@ package org.springframework.context.support;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Properties;
+import java.util.stream.Collectors;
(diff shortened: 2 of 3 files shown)


ChristianMurphy:
Do we want to do this per property?
Or would it be better to generalize this to have different properties files for different envs, so multiple properties can be customized?

bjagg:
> Do we want to do this per property?
> Or would it be better to generalize this to have different properties files for different envs, so multiple properties can be customized?

Before this PR, different properties files was the only option. Thus, Ops folks would have a separate uPortal.properties, global.properties, notifications.properties, etc. for each environment. This is still a good practice.

This PR is a convenience enhancement that will allow all those versions of a same file to be brought back into a single file while still supporting multiple environments.

Issue #1691, opened 2019-05-11
Aggregation doesn't aggregate past "now"


jonathanmtran:
**Describe the bug**
Aggregation doesn't appear to aggregate past "now". The numbers in the Portal Activity portlet for "today" and "yesterday" remain at 0 and there are no search terms.

**To Reproduce**
1. Log in, browse, etc to generate events
2. Wait at least 5 minutes
3. Review Portal Activity portlet

**Expected behavior**
Updated numbers and search terms

bjagg:
https://github.com/Jasig/uPortal/blob/master/uPortal-webapp/src/main/webapp/WEB-INF/jsp/Activity/activity.jsp

bjagg:
https://github.com/Jasig/uPortal/blob/fb42cf9525325f6fea234bf9593910e0ef45d51f/uPortal-portlets/src/main/java/org/apereo/portal/portlets/activity/ActivityController.java

ChristianMurphy:
resolved in #1788

PR #1788, opened 2019-07-17
Fix/1691 aggregation break past now

cbeach47:
Backing ticket #1691 

Event aggregators were not being wired in properly to uPortal, so this drops the specificity of the `IPortalEventAggregator` to allow all of the event aggregrators to be pulled in.

I also switched the standard event aggregators not needing configuration to use `@Component` to clean up the `eventHandlerContext.xml` file.


ChristianMurphy:
@cbeach47 code looks good :+1: 
Could you run `./gradlew goJF` to clear up the lint warnings on CI?

cbeach47:
Thanks @ChristianMurphy - `goJF` changes committed.

jonathanmtran:
Tested this fix against a vanilla instance of uPortal. As far as I can tell, the fix addresses the issue with event aggregation.

Issue #1801, opened 2019-07-29
Using sitemap in guest layout forces user to log in


allanjackson:
On uPortal 5.7.1, if you're on a guest layout and you click a portlet name in the sitemap, it'll force you to log in.

I have a code fix I'm preparing to contribute.

Issue #1802, opened 2019-07-29
Invalid link to "transient" tab in sitemap


allanjackson:
On uPortal 5.7.1, sometimes the sitemap will display a link to the transient tab, "ft1" in the sitemap. I've had it show up in a couple different situations: after logging out (and loading our custom signout page) and after clearing the portal's cache.

The link shows the portlet you previously had focused as the text, but the url is servername/uPortal/f/ft1/normal/render.uP, which won't work.

I have a code fix I'm preparing to contribute.

ChristianMurphy:
resolved in #1803

PR #1803, opened 2019-07-29
Use tab's externalId for sitemap links. Fixes GH-1801 and GH-1802.

allanjackson:
-   [ x ] the [individual contributor license agreement][] is signed
-   [ x ] commit message follows [commit guidelines][]

Use tab's externalId for sitemap links. Fixes GH-1801 and GH-1802.

This fixes 1801 because using the externalId doesn't redirect the user to log in to the portal. This fixes 1802 because the transient tab doesn't have an externalId, so it will never show up in the sitemap at all.


ChristianMurphy:
@allanjackson is this related to https://groups.google.com/a/apereo.org/d/msg/uportal-user/44Uw1YP8_Mg/hLaTlEVZFAAJ ?
Does updating CAS config resolve the issue?

allanjackson:
No, it's not related. That other discussion is about how CAS encodes special characters in the URL it redirects to. This one is about how the portlet & tab links are generated in the sitemap. I just tested it to be completely sure, and this issue still exists even with the other "encodeServiceUrl" change in place.

PR #1822, opened 2019-08-07
fix: updated encodeServiceUrl of the ticketValidationFilter bean to true

jonathanmtran:
##### Checklist

##### Description of change

See https://groups.google.com/a/apereo.org/d/msg/uportal-user/44Uw1YP8_Mg/hLaTlEVZFAAJ


allanjackson:
Looks great, thanks!

jgribonvald:
hi
did you validate this change with the embedded CAS ?

allanjackson:
It does look like this breaks with the embedded cas server. So maybe the setting you need depends on both the version of the cas-client and the cas server?

jonathanmtran:
Ouch. 

How does the following sound?
- make `encodeServiceUrl` a property, default it to `true` as it is the default for cas-client
- set the property in `uportal.properties` in uPortal-start to `false` with some verbiage and documentation

jgribonvald:
At least a documentation is needed, after having an embedded CAS server into uPortal isn't a requirement, but it helps for tests !

Issue #1858, opened 2019-08-20
"Direct URL" feature should not display in a form field


allanjackson:
When showing the "direct url" for a portlet, this information should just be displayed as text rather than put into a form input element.

As context, we had a user who was trying to add a random (non-portal related) url to the Favorites portlet. They were clicking the "direct url" link under Favorites, erasing the link, and pasting in the url they were trying to add to their Favorites. Then they were upset because the form didn't have a save button. This whole misunderstanding would be avoided if the display wasn't using a form element.

PR #1859, opened 2019-08-20
GH-1858 - Don't show "direct url" in a form input field.

allanjackson:

Show the portlet's direct url as read-only text rather than putting it in a form field that shouldn't be edited.

https://github.com/Jasig/uPortal/issues/1858


bjagg:
Was the original intent of using the form text field to make it easier for users to copy the text?

allanjackson:
I added the p tag to the tabindex, so tabbing through the document should find it now.

allanjackson:
@cbeach47 Did my tabindex change address your accessibility concerns?

bjagg (approved):

PR #1973, opened 2020-02-24
fix(pags): InvertedRegexTester against multivalued attributes

jgribonvald:
##### Checklist

-   [x] tests are included

##### Description of change

InvertedRegexTester doesn't provide a logical test result against a multivalued attribute.

The Inverted response should be done against the final result and not for each value test (this case work only on a monovalued attribute). The problem is on the foreach attribute values, when it find a not match it returns true without testing all other values, and so it return always true expect if a single matching result appear.

### DIFF
--- a/uPortal-groups/uPortal-groups-pags/src/main/java/org/apereo/portal/groups/pags/testers/AbstractStringTester.java
+++ b/uPortal-groups/uPortal-groups-pags/src/main/java/org/apereo/portal/groups/pags/testers/AbstractStringTester.java
@@ -26,7 +26,7 @@ public abstract class AbstractStringTester extends BaseAttributeTester {
     }
 
     @Override
-    public final boolean test(IPerson person) {
+    public boolean test(IPerson person) {
         boolean result = false;
         Object[] atts = person.getAttributeValues(getAttributeName());
         if (atts != null) {
--- a/uPortal-groups/uPortal-groups-pags/src/main/java/org/apereo/portal/groups/pags/testers/InvertedRegexTester.java
+++ b/uPortal-groups/uPortal-groups-pags/src/main/java/org/apereo/portal/groups/pags/testers/InvertedRegexTester.java
@@ -14,8 +14,8 @@
  */
 package org.apereo.portal.groups.pags.testers;
 
-import java.util.regex.Pattern;
 import org.apereo.portal.groups.pags.dao.IPersonAttributesGroupTestDefinition;
+import org.apereo.portal.security.IPerson;
 
 /**
  * A tester for matching the possibly multiple values of an attribute against a regular expression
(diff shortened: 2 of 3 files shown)


uPortal-groups/uPortal-groups-pags/src/main/java/org/apereo/portal/groups/pags/testers/AbstractStringTester.java
```diff
@@ -26,7 +26,7 @@ public AbstractStringTester(IPersonAttributesGroupTestDefinition definition) {
     }
 
     @Override
-    public final boolean test(IPerson person) {
+    public boolean test(IPerson person) {
```
ChristianMurphy: out of curiosity, why is this no longer `final`?

jgribonvald: because we can't override it or have a same method name from a parent class.

jgribonvald: Also final method is used only when you have a public method on a library and that you don't want that anybody extend and change the comportement.

I can't find why we have a such case here !

jgribonvald:
And so any comment/review ? Can I merge it ?

uPortal-groups/uPortal-groups-pags/src/main/java/org/apereo/portal/groups/pags/testers/InvertedRegexTester.java
```diff
@@ -49,27 +49,15 @@
...
-     * @param patternString regex pattern string
-     */
-    protected void setPattern(String patternString) {
-        pattern = Pattern.compile(patternString);
     }
 
     @Override
-    public boolean test(String att) {
-        return !pattern.matcher(att).matches();
+    public boolean test(IPerson person) {
```
ChristianMurphy: I'm curious why this tests a person rather than an attribute

jgribonvald: It's a change that Drew made when migrate PAGS from xml to JPA, it's better as you can access to all IPerson attributes in an abstract way (It implements the [IPersonTester interface](https://github.com/Jasig/uPortal/blob/master/uPortal-groups/uPortal-groups-pags/src/main/java/org/apereo/portal/groups/pags/IPersonTester.java)).

The oldest method boolean test(String att) isn't anymore used expect internally in some cases.

ChristianMurphy: thanks for the clarity

Issue #2006, opened 2020-03-17
Refactor search to support previous functionality


bjagg:
**Describe the bug**
Search supported case-insensitive, partial searches. Now it case is used in ranking and requires wildcards.

ChristianMurphy:
https://github.com/kentcdodds/match-sorter is what esco-content-menu used to solve a similar problem, it may help here as well

PR #2009, opened 2020-03-19
fix(search): fix partial searching and run indexer at startup

bjagg:
##### Checklist

##### Description of change

Lucene indexing was not supporting partial term searching without wildcard, which is how search had been working. Also, Lucene indexer needed to run before results were returned (exception broken results).

1. Wildcard is added automatically with this PR
2. Indexer is run at startup
3. Indexer is run every minute by default now

Resolves Issue #2006

### DIFF
--- a/uPortal-api/uPortal-api-rest/src/main/java/org/apereo/portal/rest/search/PortletsSearchStrategy.java
+++ b/uPortal-api/uPortal-api-rest/src/main/java/org/apereo/portal/rest/search/PortletsSearchStrategy.java
@@ -86,12 +86,12 @@ public class PortletsSearchStrategy implements ISearchStrategy {
 
     @Override
     public List<?> search(String query, HttpServletRequest request) {
-
+        logger.debug("Entering search() with query={}", query);
         final List<Object> rslt = new ArrayList<>();
 
         try (IndexReader indexReader = DirectoryReader.open(directory)) {
-
-            final Query q = queryParser.parse(query);
+            final String queryString = query.endsWith(" ") ? query : query + "*";
+            final Query q = queryParser.parse(queryString);
             final IndexSearcher searcher = new IndexSearcher(indexReader);
             final TopDocs topDocs = searcher.search(q, 50);
             Arrays.stream(topDocs.scoreDocs)
--- a/uPortal-webapp/src/main/resources/properties/contexts/schedulerContext.xml
+++ b/uPortal-webapp/src/main/resources/properties/contexts/schedulerContext.xml
@@ -63,7 +63,7 @@
         <task:scheduled ref="portalEventDaoQueuingEventHandler" method="flush" fixed-delay="997"/> <!-- ~1 second period -->
         <task:scheduled ref="cacheManagerExpiredElementEvictor" method="evictExpiredElements" fixed-delay="61900"/> <!-- ~1 minute period -->
         <task:scheduled ref="cacheHealthReporterService" method="generateReports" fixed-delay="60013"/> <!-- ~1 minute period -->
-        <task:scheduled ref="portalSearchIndexer" method="updateIndex" fixed-delay="${org.apereo.portal.index.portalSearchIndexer.updateIndexPeriod:180001}"/> <!-- ~3 minute period -->
+        <task:scheduled ref="portalSearchIndexer" method="updateIndex" fixed-delay="${org.apereo.portal.index.portalSearchIndexer.updateIndexPeriod:58001}"/> <!-- ~1 minute period -->
 
         <!-- clustered tasks -->
         <task:scheduled ref="portletCookieService" method="purgeExpiredCookies" fixed-delay="${org.apereo.portal.portlet.container.services.PortletCookieServiceImpl.purgeExpiredCookiesPeriod}"/>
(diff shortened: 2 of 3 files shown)


bjagg:
Looking at sorting results tabs by # of returned items. Is annoying when the empty People tab shows up first with desired results in the Portlets tab.

uPortal-webapp/src/main/resources/properties/contexts/schedulerContext.xml
```diff
@@ -63,7 +63,7 @@
         <task:scheduled ref="portalEventDaoQueuingEventHandler" method="flush" fixed-delay="997"/> <!-- ~1 second period -->
         <task:scheduled ref="cacheManagerExpiredElementEvictor" method="evictExpiredElements" fixed-delay="61900"/> <!-- ~1 minute period -->
         <task:scheduled ref="cacheHealthReporterService" method="generateReports" fixed-delay="60013"/> <!-- ~1 minute period -->
-        <task:scheduled ref="portalSearchIndexer" method="updateIndex" fixed-delay="${org.apereo.portal.index.portalSearchIndexer.updateIndexPeriod:180001}"/> <!-- ~3 minute period -->
+        <task:scheduled ref="portalSearchIndexer" method="updateIndex" fixed-delay="${org.apereo.portal.index.portalSearchIndexer.updateIndexPeriod:58001}"/> <!-- ~1 minute period -->
```
ChristianMurphy: Would this have a significant affect on performance?
Or is indexing trivial in this case?

bjagg: Good question!
Performance seems reasonable. This is replacing the calls doing similar work for each search request. It's also configurable for those institutions that see issues.

uPortal-api/uPortal-api-rest/src/main/java/org/apereo/portal/rest/search/PortletsSearchStrategy.java
```diff
@@ -86,12 +86,12 @@ public String getResultTypeName() {
...
     @Override
     public List<?> search(String query, HttpServletRequest request) {
-
+        logger.debug("Entering search() with query={}", query);
         final List<Object> rslt = new ArrayList<>();
 
         try (IndexReader indexReader = DirectoryReader.open(directory)) {
-
-            final Query q = queryParser.parse(query);
+            final String queryString = query.endsWith(" ") ? query : query + "*";
```
ChristianMurphy: If I'm reading this correctly it will add a wildcard if a user is typing and has reached the end of a word for example "notification ", but would not add a wildcard for midword typing such as "notificatio"?
Is that expected? Would we want a wildcard in that case?

jgribonvald: why not adding a wildcard all the time ? 
I think that users won't understand (me first) that they should add a space at the end of a word to have a largest search. Also how it will works if the user search on "notifica*" already with a wildcard ?

uPortal-webapp/src/main/resources/properties/contexts/schedulerContext.xml
jgribonvald: And we won't see affects if not enabled :laughing: (sorry hard weeks due to confinement and so school apps were never used as much but all is running well after 2 days of works :eyeglasses:  !)

An idea, for all these task we should add a doc that explain how to watch on them if there is a cost on server ressources

bjagg: Agree that we need to document parameters to consider when addressing performance.

uPortal-api/uPortal-api-rest/src/main/java/org/apereo/portal/rest/search/PortletsSearchStrategy.java
bjagg: I will have to test more, but I believe double asterisks will not be an issue. We don't want an asterisk alone because returning everything indexed is not a very good search result.

ChristianMurphy: > I will have to test more, but I believe double asterisks will not be an issue.

Are you saying there is already an asterisk on text that doesn't end with a space?
If so, would it make sense to update that code to always add an asterisk?

> We don't want an asterisk alone because returning everything indexed is not a very good search result.

Right, if we're looking to catch an empty string, would a condition like `query.trim().length > 0` make more sense?

bjagg: What the code is doing is adding an asterisk mid-word.

Example 1: "cache " -> ends with space -> "cache " -> 'cache' as token
Example 2: "notif" -> does not end with space -> "notif*"  -> 'notif*' is the token

I wan to avoid "cache " -> "cache *" -> 'cache' and '*' tokens

Searches of empty string are not allowed / submitted. Lucene is pretty good about the basics. It just leaves wildcards as an explicit. My research to enable partial search tokens without tokens has suggested simply adding a trailing asterisk.

ChristianMurphy: Ah, that makes more sense, thanks for the clarification

jgribonvald: Ok, It's good for me too !
But I'm guessing how users will understand that we need to add a space at the end of the word to make search on the exact word. Is there something to help to understand that ? like a placeholder ? a tooltip ? any other thing to help ?

PR #2128, opened 2020-08-11
fix: catch expected runtime exception when race condition of cookie...

bjagg:
… updates

##### Checklist

##### Description of change

During portlet cookie processing, there is a known potential for a race-condition where more than one portlet thread is updating the user's portlet cookie. The code previously caught the optimistic locking exception and continued. The exception was changed in Spring/Hibernate but the uPortal code was not updated. This caused failed pages (uncaught runtime exception) on occasion.

### DIFF
--- a/uPortal-web/build.gradle
+++ b/uPortal-web/build.gradle
@@ -12,6 +12,7 @@ dependencies {
     compile project(':uPortal-rendering')
 
     compile "com.google.visualization:visualization-datasource:${googleVisualizationVersion}"
+    compile "joda-time:joda-time:${jodaTimeVersion}"
 
     testCompile "org.apache.portals.pluto:pluto-container-api:${plutoVersion}"
     testCompile "${servletApiDependency}"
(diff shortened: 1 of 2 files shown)


bjagg:
Planning to cut a release once this is approved & merged.

uPortal-web/build.gradle
```diff
@@ -12,6 +12,7 @@ dependencies {
     compile project(':uPortal-rendering')
 
     compile "com.google.visualization:visualization-datasource:${googleVisualizationVersion}"
+    compile "joda-time:joda-time:${jodaTimeVersion}"
```
ChristianMurphy, line 15: I'm missing something here...
Why is this dependency needed now? It doesn't appear to be used in this PR.

bjagg: My IDE was showing errors about this being a missing dependency. It is a transitive dependency, but it should be included as it is being used by code in this module.

ChristianMurphy: :confused: Doesn't Gradle include all transitive dependencies unless they are explicitly excluded?
see https://docs.gradle.org/current/userguide/dependency_constraints.html#sec:direct-vs-transitive-deps and https://docs.gradle.org/current/userguide/dependency_downgrade_and_exclude.html

bjagg: If code relies on a dependency the module should include that rather than relying on a transitive dependency of another dependency to bring that in. Right?

ChristianMurphy: Ah, that was the missing puzzle piece, it is directly used.
https://github.com/Jasig/uPortal/blob/647b5e072aab35cacdcaa6aab0bc2c0dd429a613/uPortal-web/src/main/java/org/apereo/portal/portlet/dao/jpa/PortalCookieImpl.java#L42

PR #2173, opened 2020-09-27
Add workflow for Update Gradle Wrapper Action.

cristiangreco:
Hey there 👋, first of all thanks for your work on uPortal!

I've got a suggested change: would you be willing to use [this GitHub Action](https://github.com/gradle-update/update-gradle-wrapper-action) to automatically keep Gradle Wrapper updated to latest release?

**What does "Update Gradle Wrapper Action" do?** It can be configured to run at scheduled intervals (e.g. daily or weekly) and will check whether the Wrapper script in the repo is up-to-date to the latest Gradle release: in case a new Gradle version is available, it will create a PR to update the Wrapper. And that's it!

**Why is that a good thing?** Well, first of all it alleviates the chore of manually updating the Wrapper, as you got a task that keeps track of new Gradle releases for you! More importantly, it boosts security around the Wrapper update and usage processes: this actions verifies that the `gradle-wrapper.jar` file has not been tampered with (uses checksum comparison), and it sets the `distributionSha256Sum` property so that the new Gradle binary itself will be verified locally upon download.

**Where can I find more about?** The [README](https://github.com/gradle-update/update-gradle-wrapper-action/blob/master/README.md) contains quite detailed information!

In this PR I propose adding a new workflow which runs the action every day at midnight (but feel free to adjust the frequency as you prefer). Given uPortal is already up-to-date to Gradle 6.6.1 the action won't basically do anything until a new version is out (great work keeping it updated so regularly!).

The action is under active development, you can have a look at the [list of inputs](https://github.com/gradle-update/update-gradle-wrapper-action#action-inputs) currently supported. There's new features coming up soon and if you'd like to request any particular change just let me know!

I'd love to see the action used by uPortal and I genuinely hope you can find this useful. Would love your feedback! ❤️

### DIFF
--- a/.github/workflows/update-gradle-wrapper.yml
+++ b/.github/workflows/update-gradle-wrapper.yml
@@ -1,0 +1,17 @@
+name: Update Gradle Wrapper
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  update-gradle-wrapper:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Update Gradle Wrapper
+        uses: gradle-update/update-gradle-wrapper-action@v1
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}


ChristianMurphy:
Thanks for reaching out @cristiangreco!
Currently uPortal uses [renovate](https://github.com/renovatebot/renovate) to automatically update the gradle wrapper ([example](https://github.com/Jasig/uPortal/pull/2146)).
It also handles updating gradle, maven, and npm dependencies, [among others](https://docs.renovatebot.com/modules/manager/).

I'd prefer to stick with a single platform, which meets several of the project's dependency update needs and has been reliable, over adopting new platform(s) which are still stabilizing.

cristiangreco:
@ChristianMurphy thanks for your answer. I didn't realise you are already using renovate for updating Gradle itself, makes sense to stick to a single tool! 👍

Issue #2189, opened 2020-10-16
SELECT not using bind variables


bjagg:
**Describe the bug**
DB performance suffers from not using prepared statements.

https://github.com/Jasig/uPortal/blob/44b303f8406b042edde388f2bcd9ffed445be912/uPortal-groups/uPortal-groups-core/src/main/java/org/apereo/portal/groups/RDBMEntityGroupStore.java#L482

This should be rewritten to use bindings.

PR #2198, opened 2020-10-27
feat(uPortal-groups): Use prepared statement for entities for groups …

tbain:
…query

##### Checklist

-   [ ] tests are included

##### Description of change

Changed the Entities Group SQL statement to use a prepared statement rather than concatenated strings

### DIFF
--- a/uPortal-groups/uPortal-groups-core/src/main/java/org/apereo/portal/groups/RDBMEntityGroupStore.java
+++ b/uPortal-groups/uPortal-groups-core/src/main/java/org/apereo/portal/groups/RDBMEntityGroupStore.java
@@ -112,6 +112,19 @@ public class RDBMEntityGroupStore implements IEntityGroupStore, IGroupConstants 
                     + GROUP_NAME_COLUMN
                     + " = ?";
 
+    private static final String SEARCH_ENTITIES_FOR_GROUP =
+            "SELECT "
+                    + MEMBER_MEMBER_KEY_COLUMN
+                    + " FROM "
+                    + MEMBER_TABLE
+                    + " WHERE "
+                    + MEMBER_GROUP_ID_COLUMN
+                    + "=? AND "
+                    + MEMBER_IS_GROUP_COLUMN
+                    + " = '"
+                    + MEMBER_IS_ENTITY
+                    + "'";
+
     private static final Log LOG = LogFactory.getLog(RDBMEntityGroupStore.class);
 
     private static String groupNodeSeparator;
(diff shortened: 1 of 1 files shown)


uPortal-groups/uPortal-groups-core/src/main/java/org/apereo/portal/groups/RDBMEntityGroupStore.java
```diff
@@ -471,30 +484,19 @@ public Iterator findParentGroups(IGroupMember gm) throws GroupsException {
...
         Collection entities = new ArrayList();
         Connection conn = null;
+        PreparedStatement ps = null;
         String groupID = group.getLocalKey();
         Class cls = group.getLeafType();
 
         try {
             conn = RDBMServices.getConnection();
-            Statement stmnt = conn.createStatement();
+            ps = conn.prepareStatement(SEARCH_ENTITIES_FOR_GROUP);
```
ChristianMurphy, line 493: I'm not as familiar with connection behavior here.
Does this mean the query will be re-prepared each time it is called?
Or is this clever and it avoid preparing it again if it already exists?

ChristianMurphy (commented):
Looks pretty good, a quest for my own edification.

uPortal-groups/uPortal-groups-core/src/main/java/org/apereo/portal/groups/RDBMEntityGroupStore.java
tbain: I think you're correct that it will re-prepare the statement each time. This is how the same sort of code is used in several other instances in this file though, so this is the established pattern. It might be worth a ticket to perhaps revisit those and make it so the statement is only prepared once, if that would still work?

ChristianMurphy: > This is how the same sort of code is used in several other instances in this file though, so this is the established pattern

Right

> I think you're correct that it will re-prepare the statement each time
> It might be worth a ticket to perhaps revisit those and make it so the statement is only prepared once, if that would still work?

Absolutely, it doesn't need to be a part of this PR, I was mainly curious. :heavy_check_mark: 
If there's time a follow up task/PR to do some optimizations would be welcome. :slightly_smiling_face:

ChristianMurphy (approved):

cbeach47:
Issue link - #2189

PR #2266, opened 2021-01-26
fix: handle null return from iterator

bjagg:
##### Checklist

-   [ ] tests are included

##### Description of change

### DIFF
--- a/uPortal-groups/uPortal-groups-core/src/main/java/org/apereo/portal/groups/ReferenceIndividualGroupService.java
+++ b/uPortal-groups/uPortal-groups-core/src/main/java/org/apereo/portal/groups/ReferenceIndividualGroupService.java
@@ -14,6 +14,13 @@
  */
 package org.apereo.portal.groups;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
 import org.apereo.portal.EntityIdentifier;
 import org.apereo.portal.concurrency.CachingException;
 import org.apereo.portal.concurrency.IEntityLock;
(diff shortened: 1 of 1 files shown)


uPortal-groups/uPortal-groups-core/src/main/java/org/apereo/portal/groups/ReferenceIndividualGroupService.java
```diff
@@ -175,17 +175,22 @@ public Iterator findParentGroups(IGroupMember gm) throws GroupsException {
         log.debug("Finding containing groups for member {}", gm.getKey());
         Collection groups = new ArrayList(10);
         IEntityGroup group = null;
-        for (Iterator it = getGroupStore().findParentGroups(gm); it.hasNext(); ) {
+        Iterator it = getGroupStore().findParentGroups(gm);
+        while (it.hasNext()) {
```
ChristianMurphy, line 179: Out of general curiosity, why was `for (item : iterable)` used elsewhere, but `while` used here?

ChristianMurphy (approved):
Thanks @bjagg!

uPortal-groups/uPortal-groups-core/src/main/java/org/apereo/portal/groups/ReferenceIndividualGroupService.java
bjagg: It is to signify when the collection is the right class verses when the collection is being cast. If I had more time, I would consider making it consistent across the class.

ChristianMurphy: thanks for the clarification, totally get the need to timebox.

PR #2322, opened 2021-04-12
Fix: uniqueness on favorited portlets GH:2275

RamiroPuentesSCCCD:
##### Checklist

-   [x] tests are included

##### Description of change

### DIFF
--- a/uPortal-api/uPortal-api-internal/src/main/java/org/apereo/portal/portlets/favorites/FavoritesUtils.java
+++ b/uPortal-api/uPortal-api-internal/src/main/java/org/apereo/portal/portlets/favorites/FavoritesUtils.java
@@ -22,7 +22,11 @@ import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+import java.util.function.Predicate;
 import org.apache.commons.lang3.Validate;
 import org.apereo.portal.layout.IUserLayout;
 import org.apereo.portal.layout.node.IUserLayoutChannelDescription;
@@ -284,5 +288,17 @@ public class FavoritesUtils {
         // portlets
         return (!getFavoritePortletLayoutNodes(layout).isEmpty()
                 || !getFavoriteCollections(layout).isEmpty());
+    }
+
+    /**
+     * A helper method introduced to filter out duplicates by key
+     *
+     * @param keyExtractor
+     * @param <T>
+     * @return
+     */
+    public static <T> Predicate<T> distinctByKey(Function<? super T, Object> keyExtractor) {
+        Map<Object, Boolean> map = new ConcurrentHashMap<>();
+        return t -> map.putIfAbsent(keyExtractor.apply(t), Boolean.TRUE) == null;
     }
 }
--- a/uPortal-api/uPortal-api-internal/src/main/java/org/apereo/portal/portlets/favorites/FavoritesUtils.java
+++ b/uPortal-api/uPortal-api-internal/src/main/java/org/apereo/portal/portlets/favorites/FavoritesUtils.java
@@ -18,12 +18,7 @@ import static org.apereo.portal.layout.node.IUserLayoutFolderDescription.FAVORIT
 import static org.apereo.portal.layout.node.IUserLayoutFolderDescription.FAVORITE_COLLECTION_TYPE;
 import static org.apereo.portal.layout.node.IUserLayoutNodeDescription.LayoutNodeType.FOLDER;
 
-import java.util.Enumeration;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 import java.util.function.Predicate;
@@ -301,4 +296,18 @@ public class FavoritesUtils {
         Map<Object, Boolean> map = new ConcurrentHashMap<>();
         return t -> map.putIfAbsent(keyExtractor.apply(t), Boolean.TRUE) == null;
     }
+
+    public static List<IUserLayoutNodeDescription> filterFavoritesToUnique(
+            List<IUserLayoutNodeDescription> originalFavoritesList) {
+        List<IUserLayoutNodeDescription> uniqueFavorites = new ArrayList<>();
+        Predicate<IUserLayoutNodeDescription> predicate;
+        predicate = FavoritesUtils.distinctByKey(p -> p.getName());
+        for (IUserLayoutNodeDescription favorite : originalFavoritesList) {
+            if (predicate.test(favorite)) {
+                uniqueFavorites.add(favorite);
+            }
+        }
+        return uniqueFavorites;
+    }
+
 }
--- a/uPortal-api/uPortal-api-internal/src/main/java/org/apereo/portal/portlets/favorites/FavoritesUtils.java
+++ b/uPortal-api/uPortal-api-internal/src/main/java/org/apereo/portal/portlets/favorites/FavoritesUtils.java
@@ -309,5 +309,4 @@ public class FavoritesUtils {
         }
         return uniqueFavorites;
     }
-
 }
--- a/uPortal-api/uPortal-api-internal/src/main/java/org/apereo/portal/portlets/favorites/FavoritesUtils.java
+++ b/uPortal-api/uPortal-api-internal/src/main/java/org/apereo/portal/portlets/favorites/FavoritesUtils.java
@@ -18,10 +18,17 @@ import static org.apereo.portal.layout.node.IUserLayoutFolderDescription.FAVORIT
 import static org.apereo.portal.layout.node.IUserLayoutFolderDescription.FAVORITE_COLLECTION_TYPE;
 import static org.apereo.portal.layout.node.IUserLayoutNodeDescription.LayoutNodeType.FOLDER;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Enumeration;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 import org.apache.commons.lang3.Validate;
 import org.apereo.portal.layout.IUserLayout;
 import org.apereo.portal.layout.node.IUserLayoutChannelDescription;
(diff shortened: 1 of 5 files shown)


jgribonvald (commented):
I guess if it would be preferable to filter on frname instead of name. @bjagg what do you think ?

RamiroPuentesSCCCD:
Are some of these checks failing because I'm developing on Windows 10?

jonathanmtran:
The check is failing because of the following:

>  org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':verifyGoogleJavaFormat'.

Try `./gradlew goJF` to format the code

ChristianMurphy:
> Are some of these checks failing because I'm developing on Windows 10?

I don't spot any Windows specific issue flagged (CRLF related at least).
It does seem to be google code style related though (https://github.com/uPortal-Project/uPortal/pull/2322/checks?check_run_id=2495940121#step:5:39).
Try running the formatter locally

```
./gradlew goJF
```

RamiroPuentesSCCCD:
The current favorites list object is of type IUserLayoutNodeDescription which doesn't have a method to retrieve FName, I could be wrong. Should I cast this favorites list to type IUserLayoutChannelDescription then filter on Fname then cast back to a list of IUserLayoutNodeDescription? Thank you for the feedback and I'd appreciate any suggestions.

RamiroPuentesSCCCD:
Just following up, is it the consensus to refactor to use FName?

uPortal-api/uPortal-api-internal/src/main/java/org/apereo/portal/portlets/favorites/FavoritesUtils.java
```diff
@@ -285,4 +284,29 @@ public boolean hasAnyFavorites(IUserLayout layout) {
...
+        List<IUserLayoutNodeDescription> uniqueFavorites = new ArrayList<>();
+        Predicate<IUserLayoutNodeDescription> predicate;
+        predicate = FavoritesUtils.distinctByKey(p -> p.getName());
+        for (IUserLayoutNodeDescription favorite : originalFavoritesList) {
+            if (predicate.test(favorite)) {
+                uniqueFavorites.add(favorite);
+            }
+        }
+        return uniqueFavorites;
+    }
```
bjagg, line 311: You may want to look at this example to make `filterFavoritesToUnique()` a little more tidy:
https://howtodoinjava.com/java8/java-stream-distinct-examples/

```
List<IUserLayoutNodeDescription> uniqueFavorites = originalFavoritesList.stream()
  .filter( FavoritesUtils.distinctByKey(p -> p.getName() )
  .collect( Collectors.toList() );
```

uPortal-api/uPortal-api-internal/src/test/java/org/apereo/portal/portlets/favorites/FavoritesUtilsTest.java
```diff
@@ -0,0 +1,91 @@
...
+        String ObjectIdOfSecondItem = uniqueFavorites.get(1).getId();
+
+        // Confirms the first duplicate item was removed from the
+        // list
+        Assert.assertEquals("n28", ObjectIdOfSecondItem);
+
+        Assert.assertEquals(0, emptyArray.size());
+        Assert.assertEquals(0, emptyFavoritesListReturnedFromFavoritesFilter.size());
+    }
+}
```
bjagg, line 91: Excellent!!

bjagg (changes requested):

cbeach47:
@RamiroPuentesSCCCD - regarding filtering by `name` vs `fname` - since the `fname` is the key for portlets, and uPortal only "favorites" portlets, so `fname` should always be available, I recommend going with @jgribonvald 's note on filtering by that attribute.

RamiroPuentesSCCCD:
This is great feedback, thank you. I'll get started on these changes as soon as I can.

jgribonvald:
> @RamiroPuentesSCCCD - regarding filtering by `name` vs `fname` - since the `fname` is the key for portlets, and uPortal only "favorites" portlets, so `fname` should always be available, I recommend going with @jgribonvald 's note on filtering by that attribute.

I've found it, the fname filtering could be done by p.getId() instead of p.getName().

uPortal-api/uPortal-api-internal/src/test/java/org/apereo/portal/portlets/favorites/FavoritesUtilsTest.java
jgribonvald: I'm wondering if this will works in the case we add channels in these nodes ? and having same channels in different nodes, that would be authorized, I think that your filter will remove "not duplicated" channels on different folders.

Could you add a such case in your tests ?

thanks !

RamiroPuentesSCCCD: Always glad to add additional unit tests however I'm not sure I understand the scenario. By different folders do you mean folders other than the "favorites" folderType?

cbeach47: @jgribonvald - I chatted offline with @bjagg a bit about this - if I understand your comment correctly, the favorites portlet should only show 1 entry per favorited portlet, no matter if it was favorited from different layouts / channels (thus @RamiroPuentesSCCCD should be able to base uniqueness solely on the portlet's `fname`).

Did I understand your comment correctly?

RamiroPuentesSCCCD:
@jgribonvald - I tried filtering by ID however it appears the ID is created to be unique. In the event a user has already added Test Portlet 1 and Test Portlet 2 to their favorites it will appear with the assigned ID in the portlet "n51" & "n52". However when those same two portlets are added as channels their ID will appear with additional characters prefixing the portlet ID. To confirm this, I added both Test Portlets as channels with ID's "n501" and "n502" then I called the getId() method which returned "u21|1n501" & "u21|1n502" for each Portlet. So when filtering a list with duplicates using getId() the list will appear to be unique as the ID's are different. The Interface IUserLayoutChannelDescription has a method to return the FName - getFunctionalName(). So I passed a list of IUserLayoutChannelDescription to the filterToUnique method then used the Fname to filter the list as well as refactoring the accompanying unit tests.

uPortal-api/uPortal-api-internal/src/test/java/org/apereo/portal/portlets/favorites/FavoritesUtilsTest.java
jgribonvald: @cbeach47: I think that some adopters are using the folder feature to organize favorites, in this case we can put as favorite a same portlet in several folders and so the filtering will permit to get only once the favorited portlet. Or I'm wrong ?
After In my mind using folders with favorites isn't a good idea, so this contribution is good for me.

jgribonvald:
@RamiroPuentesSCCCD thanks a lot it's better so and good for me !

jgribonvald (changes requested):
Also please edit the PR name to match something more readable. It will be used as reference and description in Changelog when a new version will be cut.

uPortal-api/uPortal-api-internal/src/test/java/org/apereo/portal/portlets/favorites/FavoritesUtilsTest.java
cbeach47: That's my understanding @jgribonvald - favorites portlet should ignore folders and just focus on the portlet (keyed by the `fname`).

cbeach47 (commented):
Minor comment inline besides what @bjagg and @jgribonvald  have noted

uPortal-api/uPortal-api-rest/src/main/java/org/apereo/portal/layout/dlm/remoting/UpdatePreferencesServlet.java
```diff
@@ -692,52 +689,56 @@ public ModelAndView removeFavorite(
...
+                                            + "from favorites for user "
+                                            + (upm.getPerson() == null
+                                                    ? "unknown"
+                                                    : upm.getPerson().getID()));
+                            response.setStatus(HttpServletResponse.SC_FORBIDDEN);
+                            // load fail message
+                            resp =
+                                    getMessage(
+                                            "error.remove.favorite",
+                                            "Can''t remove favorite",
```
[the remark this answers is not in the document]
RamiroPuentesSCCCD: I just confirmed the first ' is an escape to allow for the second to appear in the string.

uPortal-api/uPortal-api-internal/src/main/java/org/apereo/portal/portlets/favorites/FavoritesUtils.java
RamiroPuentesSCCCD: Thanks for this feedback, I refactored the filterFavoritesToUnique method to follow the example you provided.

uPortal-api/uPortal-api-internal/src/main/java/org/apereo/portal/portlets/favorites/FavoritesUtils.java
```diff
@@ -18,11 +18,10 @@
 import static org.apereo.portal.layout.node.IUserLayoutFolderDescription.FAVORITE_COLLECTION_TYPE;
 import static org.apereo.portal.layout.node.IUserLayoutNodeDescription.LayoutNodeType.FOLDER;
 
-import java.util.Enumeration;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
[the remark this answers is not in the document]
RamiroPuentesSCCCD: No problem, I have updated my IDE and confirmed the wildcards were replaced with specific imports.

uPortal-api/uPortal-api-rest/src/main/java/org/apereo/portal/layout/dlm/remoting/UpdatePreferencesServlet.java
```diff
@@ -692,52 +689,56 @@ public ModelAndView removeFavorite(
...
+                            (UserLayoutChannelDescription) deleteNode;
+                    try {
+                        if (!ulm.deleteNode(channelDescription.getChannelSubscribeId())) {
+                            logger.warn(
+                                    "Error deleting the node"
+                                            + channelId
+                                            + "from favorites for user "
+                                            + (upm.getPerson() == null
+                                                    ? "unknown"
+                                                    : upm.getPerson().getID()));
```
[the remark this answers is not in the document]
RamiroPuentesSCCCD: I have updated the logger to use string interpolation instead of concatenation.

uPortal-api/uPortal-api-rest/src/main/java/org/apereo/portal/layout/dlm/remoting/UpdatePreferencesServlet.java
```diff
@@ -695,21 +696,22 @@ public ModelAndView removeFavorite(
...
                             (UserLayoutChannelDescription) deleteNode;
                     try {
                         if (!ulm.deleteNode(channelDescription.getChannelSubscribeId())) {
+
                             logger.warn(
-                                    "Error deleting the node"
-                                            + channelId
-                                            + "from favorites for user "
-                                            + (upm.getPerson() == null
+                                    MessageFormat.format(
```
jgribonvald, line 708: You don't need to use MessageFormat, SLF4J permit to replace variable in the order of argument passed. As example see on line https://github.com/uPortal-Project/uPortal/blob/d9fdb4523a5ff7bd8a9c791c6377966dba267233/uPortal-api/uPortal-api-rest/src/main/java/org/apereo/portal/layout/dlm/remoting/UpdatePreferencesServlet.java#L233

But anyway thanks !

uPortal-api/uPortal-api-rest/src/main/java/org/apereo/portal/layout/dlm/remoting/UpdatePreferencesServlet.java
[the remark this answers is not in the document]
cbeach47: ok cool

uPortal-api/uPortal-api-rest/src/main/java/org/apereo/portal/layout/dlm/remoting/UpdatePreferencesServlet.java
RamiroPuentesSCCCD: Thanks for the tip, I removed MessageFormat.

cbeach47 (approved):
LGTM :+1:

jgribonvald (approved):
LGTM and thanks for your contribution @RamiroPuentesSCCCD

bjagg (approved):

PR #2526, opened 2022-07-27
deps: bump person-directory to 1.8.16

jgribonvald:
##### Checklist

-   [ ] tests are included

##### Description of change

Update person-directory dependency with new IPersonAttributeDao methods signature (would have been a breaking change in dependency).

Warning: only applies compatibility change, the new lib introduced a Filter parameter, so in some case a local implementation of the filter could be a better way than filtering after methods call with a old java code.


ChristianMurphy (approved):
Does the filter always need to be passed?
I'm a bit surprised it doesn't/wouldn't default to `alwaysChoose`.

Other than that, LGTM, thanks!

jgribonvald:
Any other review before to merge ?

Thanks

jgribonvald:
Finally don't merge it, I don't know why my test didn't show me a problem when using it in pre-production, but it seems that the new version need some cache configuration as it provides several warn due to that.
So don't know, maybe the best is to wait to upgrade to latest version when migrating to next major version ?

PR #2527, opened 2022-07-27
deps: bump jackson version to 2.13.3 - comment possibility on transient

jgribonvald:
##### Checklist

-   [ ] tests are included

##### Description of change

Update jackson dependency from 2.13.0 to 2.13.3 and provide in comment a possibility to fix transient dependency adding com.fasterxml.jackson.dataformat:jackson-dataformat-cbor:2.12.6 version from aws s3 lib


ChristianMurphy (approved):
Updating Jackson sounds good.
Which dependency is/wad holding Jackson back, requiring thr additional `constraint`?
Could we update that dependency and remove the `constraint`?

jgribonvald:
> Updating Jackson sounds good. Which dependency is/wad holding Jackson back, requiring thr additional `constraint`? Could we update that dependency and remove the `constraint`?

The constraint is behind the dependency linked:
```
compile "com.amazonaws:aws-java-sdk-s3:${awsVersion}"
    constraints {
        implementation("com.fasterxml.jackson.dataformat:jackson-dataformat-cbor:${jacksonVersion}") {
            because 'previous versions have security problems'
        }
    }
```
I didn't find any update of the dependency that solve the problem, see here: https://mvnrepository.com/artifact/com.amazonaws/aws-java-sdk-core/1.12.268

and it's coming from here
Sélection_323

ChristianMurphy:
Updating AWS SDK to version `1.12.187` or higher should resolve the issue https://github.com/aws/aws-sdk-java/issues/2728#issuecomment-1082182287

jgribonvald:
> Updating AWS SDK to version `1.12.187` or higher should resolve the issue [aws/aws-sdk-java#2728 (comment)](https://github.com/aws/aws-sdk-java/issues/2728#issuecomment-1082182287)

So you prefer to stay on a jackson dependency that doesn't have the same version that other included jackson dependencies ?

ChristianMurphy:
> So you prefer to stay on a jackson dependency that doesn't have the same version that other included jackson dependencies ?

I could go either way.
I'd lean a bit toward keeping the version AWS SDK uses by default in place.
IIRC we don't have a lot of coverage on the AWS integration, I'm cautious that version drift away from what AWS SDK expects could cause issues which are not currently caught by continuous integration/github actions.

jgribonvald:
@ChristianMurphy I've kept my code in comments, in case it's required one day...

jgribonvald:
Any other review ? I plan to merge that.
Thanks

uPortal-portlets/build.gradle
```diff
@@ -6,6 +6,11 @@ dependencies {
     compile project(':uPortal-web')
 
     compile "com.amazonaws:aws-java-sdk-s3:${awsVersion}"
+//    constraints {
+//        implementation("com.fasterxml.jackson.dataformat:jackson-dataformat-cbor:${jacksonVersion}") {
+//            because 'previous versions have security problems'
+//        }
+//    }
```
ChristianMurphy, line 13: do we still need this commented out code?
could it be removed?

ChristianMurphy (approved):

uPortal-portlets/build.gradle
jgribonvald: The comment provide an example and a description of what to do, so in my mind it's required to keep it, as only some peoples deploy this lib (nobody in France at least).

jgribonvald: The AWS SDK lib seems to be needed even if we don't use it to be able to compile, but it provides a transient lib that isn't required but I tested removing it without having a context using AWS SDK. So I don't know if it works in AWS SDK context when removing this transient dependency.
After removing this transient dependency is a "security" patch (CVE at low-level) when we don't need it.
Also keeping this comment could provide a warning point to someone who know how to test that.

After the question is how to manage de potential security problem at this point ?

ChristianMurphy: Looking though the known vulns https://security.snyk.io/vuln/maven/?search=Jackson
None apply to the current version of Jackson used by AWS SDK https://github.com/aws/aws-sdk-java/blob/04eaedd554094f734967ecee8b54cd938ccbd55f/pom.xml#L362

And we are running latest AWS SDK which has these patches https://github.com/uPortal-Project/uPortal/pull/2542

Unless you're aware of an unreported vulnerability, I don't think there is a potential vuln here.
Which means a warning seems unwarranted.

PR #2558, opened 2022-09-18
feature/EventAnalytics rest controller and service to publish webpage…

mgillian:
… events to uPortal event service

##### Checklist

-   [ ] the [individual contributor license agreement][] is signed
-   [ ] tests are included

##### Description of change

Introduces a controller and service that will handle receiving web page events and processing them through the uPortal events processor.  Initial version handles web page link click events sent via a Javascript post..


uPortal-events/src/main/java/org/apereo/portal/events/rest/AnalyticsPortalEventsController.java
```diff
@@ -0,0 +1,50 @@
...
+
+	@Autowired
+	private AnalyticsPortalEventsService service;
+
+	public AnalyticsPortalEventsController(AnalyticsPortalEventsService service) {
+		this.service = service;
+	}
+
+	@RequestMapping(method = RequestMethod.GET)
+	public @ResponseBody Map<String, Object> getAnalytics() {
```
cbeach47, line 33: Should this be removed?

uPortal-events/src/main/java/org/apereo/portal/events/PortalEventFactoryImpl.java
```diff
@@ -561,6 +568,24 @@ public void publishTenantRemovedTenantEvent(
...
+     */
+
+	@Override
+	public void publishAnalyticsPortalEvents(HttpServletRequest request, Object source, Map<String, Object> analyticsData, IPerson person) {
+        final PortalEvent.PortalEventBuilder eventBuilder =
+                this.createPortalEventBuilder(source, request);
+        Date eventDate = Calendar.getInstance(TimeZone.getTimeZone("UTC")).getTime();
+        String rawEventType = (String) analyticsData.computeIfAbsent("type", key -> "UNKNOWN");
+        // throws IllegalArgumentException
+        String eventType = InputValidator.validateAsWordCharacters(rawEventType, "String");
```
cbeach47, line 582: I would change the second parameter to 'event type' to help end users locate the issue with the failed request.

uPortal-events/src/main/java/org/apereo/portal/events/rest/AnalyticsPortalEventsService.java
```diff
@@ -0,0 +1,28 @@
...
+import org.apereo.portal.events.IPortalAnalyticsEventFactory;
+import org.apereo.portal.security.IPerson;
+import org.apereo.portal.security.IPersonManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+public class AnalyticsPortalEventsService {
```
cbeach47, line 16: Could you add an interface here so adopters can overlay this logic with a customized flow in uPortal-start?

uPortal-webapp/src/main/resources/properties/contexts/servlet/analyticsEventsServletContext.xml
```diff
@@ -0,0 +1,51 @@
...
+
+-->
+<beans xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns="http://www.springframework.org/schema/beans"
+       xmlns:context="http://www.springframework.org/schema/context"
+       xmlns:mvc="http://www.springframework.org/schema/mvc"
+       xsi:schemaLocation="
+           http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-3.2.xsd
+           http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context-3.2.xsd
+           http://www.springframework.org/schema/mvc http://www.springframework.org/schema/mvc/spring-mvc-3.2.xsd">
```
cbeach47, line 29:
```suggestion
           http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd
           http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context.xsd
           http://www.springframework.org/schema/mvc http://www.springframework.org/schema/mvc/spring-mvc.xsd">
```

There was an issue a while back with pinning these schemas to a version. Would it work to just leave at the default?

cbeach47 (changes requested):
Looking good @mgillian !

Comments in line.

For the failing github actions / jobs. You can format the code with `./gradlew goJF`.

jgribonvald:
I think it would be good to be able to secure a bit more this part, because as I can see any anonymous request can be done and will be registered. Some attacks can come from here so !
So maybe forbidden guest access, and/or disabling anonymous request, or disabling all of this part from a property could be an option.
For the security part I'm not sure that something more than blocking guest activity can be done.

----------------------------------------------------------------------------------------------------------------------------------------------

Else on my side we have a Servlet that is used by redirects for all other apps that can't be logged from events: https://github.com/GIP-RECIA/recia-uPortal-adds/blob/e6a6c0bd7353d5ec1e1283d4ccd4b146c8bdaffd/src/main/java/fr/recia/portal/events/handlers/EscoStatsLogger.java

And we are processing all logs to obtains stats, we can get session times, number of time a user connected, and times inside apps, etc... All is logged and processed but outside of the portal for better stats.

bjagg:
Good point, @jgribonvald -- there should be a flag (and defaulted to `true`) to ignore posts to the API where there is no user session. 

Also, would it feasible (and quick) to add a nonce to the page to pass along to the API? This could make the API a little more secure.

bjagg:
Let's step back from the nonce. I think session tracking for non-guests is enough

cbeach47 (approved):
👍

PR #2605, opened 2022-12-19
fix/java version in CI

loulou2u:
##### Checklist

-   [ ] tests are included

##### Description of change


ChristianMurphy:
Thanks for looking into the CI error @loulou2u!

I was under the impression the code should build under Java 11, even if it isn't quite fully compatible yet.
I'd ideally like to keep it a part of CI if it isn't the root cause.

It looks like the issue has been appearing in the Java 8 build for a while https://github.com/uPortal-Project/uPortal/actions/runs/3715033479/jobs/6299708747#step:3:4

Perhaps the fix for Java 8 will resolve the issue in Java 11 as well?

.github/workflows/CI.yml
```diff
@@ -21,7 +21,6 @@ jobs:
           - zulu
         java-version:
           - 8
-          - 11
```
ChristianMurphy, line 24: Do the tests still break on Java 11 with the dependency changes?
If not would it make sense to re-add it to the test matrix?

ChristianMurphy (approved):
Thanks @loulou2u!

.github/workflows/CI.yml
loulou2u: I'm re-adding Java 11 back to this file. I found that there were 2 dependency upgrades that were causing the tests to fail, despite it working after performing a dataInit, portalInit, and tomcatDeploy.

gradle.properties
```diff
@@ -42,7 +42,7 @@ antVersion=1.10.12
 awsVersion=1.12.352
 apereoPortletUtilsVersion=1.1.3
 aspectjVersion=1.9.9.1
-casClientVersion=3.6.4
+casClientVersion=3.6.2
```
bjagg, line 45: Thanks for tracking down these issues!

bjagg (approved):
🎖️

ChristianMurphy (approved):
:shipit:

PR #2612, opened 2023-01-10
Gradle upgrade

mgillian:
##### Checklist

-   [x ] commit message follows [commit guidelines][]
-   [ ] tests are included

##### Description of change

Moving towards Gradle 7.  compile, runtime, and testCompile are all deprecated.  In addition, java-library required to support api, which handles classpath issues associated with subprojects.


ChristianMurphy (commented):
Thanks @mgillian!
Dependency resolution and `java-library` changes look good. 👍 
There are a few other things which may be wanted/needed.

* The gradle wrapper is not updated in this PR, it can be updated with `./gradlew wrapper --gradle-version={version number goes here}`
  * I suspect this update may uncover some plugin incompatibility
* Validation of Gradle wrapper is now recommended in GitHub actions (https://gradle.org/whats-new/gradle-7/#verifying-the-integrity-of-gradle-wrappers)
* the `uploadArchives` is now deprecated and slated for removal, it could be ideal to replace with the maven publish plugin
* The publishing task has changed it's configuration (needing `=` assignment instead of function to pass options)
* Dependency verification is now generally recommended https://docs.gradle.org/current/userguide/dependency_verification.html, a verification file can be generated with `./gradlew --write-verification-metadata sha256`

There was some previous work in https://github.com/uPortal-Project/uPortal/pull/2337/files which may offer some useful reference

We may also be able parts of this into future PRs to make the upgrade more bite-sized.

bjagg:
Hmm, should break these down into issues or even copy them into a single issue to drive multiple PRs

mgillian:
The current gradle wrapper version is 6.9.3  The next version is 7.x, and that will break stuff.  The main reason I only did these changes was because I wasn't ready to make a bigger effort.  The maven plugin needs to be replaced with the maven-publish plugin, but I wasn't ready to write the configurations that are in uportal-start's build.gradle.  That's another task.

All of these suggestions I intend to do, but after speaking with Benito (and running out of time) I limited the scope to just these changes.  I know about the other ones and will do them as available.  I figured I'd just limit the scope of this PR, just to get it done.

ChristianMurphy (approved):

PR #2686, opened 2023-07-28
fix/issue-2684: fix newly added entity group caches

groybal:
##### Checklist

-   [x] tests are included

##### Description of change

This PR fixes an issue that is causing several Playwright scripts to fail, and causing admin users to not belong to the groups they should.  The issue is that some new caches were recently added, but the objects being stored in the cache were Iterators, which are single use objects.  The fix was to update to instead store the list of groups.

Issue:  https://github.com/uPortal-Project/uPortal/issues/2684

### DIFF
--- a/uPortal-groups/uPortal-groups-core/src/main/java/org/apereo/portal/groups/RDBMEntityGroupStore.java
+++ b/uPortal-groups/uPortal-groups-core/src/main/java/org/apereo/portal/groups/RDBMEntityGroupStore.java
@@ -21,6 +21,7 @@ import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import net.sf.ehcache.CacheManager;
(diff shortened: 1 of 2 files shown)


uPortal-groups/uPortal-groups-core/src/test/java/org/apereo/portal/groups/RDBMSEntityGroupStoreTest.java
```diff
@@ -21,181 +29,263 @@
...
+
+    private Cache createNewCache(String cacheName) {
+        return new Cache(cacheName, 100, false, false, 300, 300);
+    }
+
+    private void setUpRdbms() throws Exception {
+        when(conn.prepareStatement(anyString())).thenReturn(ps);
+        when(ps.executeQuery()).thenReturn(rs);
+        when(rs.next()).thenReturn(true, false);
+        // set to null in order to avoid more mocks in
```
cbeach47, line 260: I don't understand this comment. What is being set to `null` ?

cbeach47 (approved):
One minor, non-blocking comment in line.

Tried this out locally - bugs are resolved, thanks @groybal  ! 🎉

uPortal-groups/uPortal-groups-core/src/test/java/org/apereo/portal/groups/RDBMSEntityGroupStoreTest.java
groybal: Ah, I see ...original code was 
```
            // set to null in order to avoid more mocks in
            // RDBMEntityGroupStore.instanceFromResultSet since we don't care about the results here
            when(rs.getString(1)).thenReturn(null);
```

I refactored to use actual cache instead of mock, so I had to set an actual ID instead of null.

uPortal-groups/uPortal-groups-core/src/main/java/org/apereo/portal/groups/RDBMEntityGroupStore.java
```diff
@@ -340,20 +342,22 @@ public IEntityGroup find(String groupID) throws GroupsException {
...
-            java.util.Iterator it = (java.util.Iterator) el.getObjectValue();
-            assert it != null;
-            return it;
-        }
-
         Integer type = EntityTypesLocator.getEntityTypes().getEntityIDFromType(ent.getLeafType());
-        java.util.Iterator it = findParentGroupsForEntity(memberKey, type.intValue());
-        parentGroupEntityCache.put(new Element(memberKey, it));
-        return it;
+        String cacheKey = memberKey + ":" + type.intValue();
```
bjagg, line 349: Good compromise

bjagg (approved):
Excellent! 🥇

PR #2719, opened 2023-11-09
Fix flaky test testExternalUrl using LinkedHashMap

KiruthikaJanakiraman:
##### Checklist

##### Description of change

This PR aims to fix the flaky test: [org.apereo.portal.redirect.PortletRedirectionControllerTest.testExternalUrl](https://github.com/KiruthikaJanakiraman/uPortal/blob/59f8895fe98782e1ff23dd14f9eb2076db09b04e/uPortal-webapp/src/test/java/org/apereo/portal/redirect/PortletRedirectionControllerTest.java#L44-L60) similar to a previously raised PR https://github.com/uPortal-Project/uPortal/pull/2718.

I found and confirmed the flaky behaviour of the tests using an open-source research tool [NonDex](https://github.com/TestingResearchIllinois/NonDex), which shuffles implementations of nondeterminism operations.

**Problem**
The test case [org.apereo.portal.redirect.PortletRedirectionControllerTest.testExternalUrl](https://github.com/uPortal-Project/uPortal/blob/989ba284b63dd8d26565d7b62213d68fd95be472/uPortal-webapp/src/test/java/org/apereo/portal/redirect/PortletRedirectionControllerTest.java#L44-L61) fails due to the below assertion. 

https://github.com/uPortal-Project/uPortal/blob/989ba284b63dd8d26565d7b62213d68fd95be472/uPortal-webapp/src/test/java/org/apereo/portal/redirect/PortletRedirectionControllerTest.java#L57-L60

This is because `additionalParameters` in `url` is a HashMap which does not preserve the order of the items causing the order of `action` and `list` to vary.
https://github.com/uPortal-Project/uPortal/blob/989ba284b63dd8d26565d7b62213d68fd95be472/uPortal-webapp/src/test/java/org/apereo/portal/redirect/PortletRedirectionControllerTest.java#L48

**Solution**
This PR fixes the test by using a custom function `compareQueryParameters` to compare two urls irrespective of the order of the query parameters.

**Steps to reproduce**
The following command can be used to reproduce the assertion errors and verify the fix.

**Integrate NonDex:**

**Add the following snippet to the top of the build.gradle in $PROJ_DIR:**
```
plugins {
    id 'edu.illinois.nondex' version '2.1.1-1'
}
```

**Append to build.gradle in $PROJ_DIR:**
```
apply plugin: 'edu.illinois.nondex'
```

**Execute Test with Gradle:**
```
./gradlew --info uPortal-webapp:test --tests org.apereo.portal.redirect.PortletRedirectionControllerTest.testExternalUrl
```

**Run NonDex:**
```
./gradlew --info uPortal-webapp:nondexTest --tests=org.apereo.portal.redirect.PortletRedirectionControllerTest.testExternalUrl --nondexRuns=50 -x autoLintGradle
```

**Test Environment:**
```
Java version "1.8.0_381"
macOS Venture Version 13.4.1 (22F82)
```

Please let me know if you have any concerns or questions.


ChristianMurphy:
Thanks @KiruthikaJanakiraman!
To clarify, I believe the intent of this test is to be order insensitive.
Would this change it to become order sensitive and potentially cause tests to fail over a property (order) we don't intend to check?

ChristianMurphy:
> The assert statement checks for a specific order of parameters. Doesn't that make the test order sensitive?

You are right, the test in is current form is indeed order sensitive.
This is testing URL query params. Query params are _not_ order sensitive.
I'd personally lean that the test could/should account query params being in any/different order.
But I'd be interested to hear what you (@KiruthikaJanakiraman) and others (@uPortal-Project/uportal-committers) think.

bjagg:
I think the test should be changed to account for any order.

KiruthikaJanakiraman:
> This is testing URL query params. Query params are _not_ order sensitive.
That makes sense. If the test is meant to be order insensitive, I would suggest 2 alternative fixes:
1. Checking for both possible order of `additionalParameters`
2. Writing a custom function to compare urls with different query parameter orders

Could you kindly let me know the preferred fix?

KiruthikaJanakiraman:
I have updated the commit with a custom function that compares two urls two urls irrespective of the order of the query parameters. Could you kindly review the same?

uPortal-webapp/src/test/java/org/apereo/portal/redirect/PortletRedirectionControllerTest.java
```diff
@@ -57,6 +60,19 @@ public void testExternalUrl() {
...
+                uri1.getPath().equals(uri2.getPath())
+                        && compareQueryParameters(uri1.getQuery(), uri2.getQuery()));
+    }
+
+    private static boolean compareQueryParameters(String query1, String query2) {
+        String[] params1 = query1.split("&");
+        String[] params2 = query2.split("&");
+
+        return Arrays.asList(params1).containsAll(Arrays.asList(params2))
+                && Arrays.asList(params2).containsAll(Arrays.asList(params1));
```
ChristianMurphy, line 76: would it make sense to use an order insensitive data structure like `Set`, and use set `.equals` to ensure all of the same values are present?

ChristianMurphy (approved):
Thanks @KiruthikaJanakiraman!

ChristianMurphy:
@KiruthikaJanakiraman could you run `./gradlew goJF` to format the changes?

PR #2738, opened 2023-11-18
Chore 2733 design level final refactor

snehit221:
##### Checklist

-   [x] tests are included

##### Description of change

[individual contributor license agreement]: https://github.com/uPortal-Project/uPortal/blob/master/docs/CONTRIBUTING.md#individual-contributor-license-agreement
[commit guidelines]: https://github.com/uPortal-Project/uPortal/blob/master/docs/CONTRIBUTING.md#commit
[message properties]: https://github.com/uPortal-Project/uPortal/tree/master/uPortal-webapp/src/main/resources/properties/i18n


snehit221:
Hi @ChristianMurphy ,
May I request your review on this PR as well please?
Thanks

uPortal-api/uPortal-api-rest/src/test/java/org/apereo/portal/rest/PortletsRESTControllerTest.java
```diff
@@ -209,41 +217,31 @@ private void givenPortletDefinitionsInRegistry(IPortletDefinition... portletDefi
...
+                    this.authorizationService.canPrincipalSubscribe(
+                            this.authorizationPrincipal, portletDefinitionStringId);
                     break;
                 case RENDER:
-                    given(
-                                    this.authorizationService.canPrincipalRender(
-                                            this.authorizationPrincipal, portletDefinitionStringId))
-                            .willReturn(true);
+                    this.authorizationService.canPrincipalRender(
+                            this.authorizationPrincipal, portletDefinitionStringId);
```
ChristianMurphy, line 244: Shouldn't these tests assert something?
Why has the testing part of the test been removed?

uPortal-rendering/src/main/java/org/apereo/portal/character/stream/events/PortletContentPlaceholderEventImpl.java
```diff
@@ -35,30 +33,6 @@ public CharacterEventTypes getEventType() {
...
-        result =
-                prime * result
-                        + ((this.getPortletWindowId() == null)
-                                ? 0
-                                : this.getPortletWindowId().hashCode());
-        result =
-                prime * result
-                        + ((this.getEventType() == null) ? 0 : this.getEventType().hashCode());
-        return result;
-    }
```
ChristianMurphy, line 60: What has this been replaced by?
Is this inherited? Is there a lombok annotations I'm not able to see in the diff? Something else?

uPortal-utils/uPortal-utils-core/src/main/java/org/apereo/portal/character/stream/events/PortletHeaderPlaceholderEventImpl.java
```diff
@@ -35,30 +33,6 @@ public CharacterEventTypes getEventType() {
...
-        result =
-                prime * result
-                        + ((this.getPortletWindowId() == null)
-                                ? 0
-                                : this.getPortletWindowId().hashCode());
-        result =
-                prime * result
-                        + ((this.getEventType() == null) ? 0 : this.getEventType().hashCode());
-        return result;
-    }
```
ChristianMurphy, line 60: same here, where has this implementation been moved to?

uPortal-utils/uPortal-utils-core/src/main/java/org/apereo/portal/character/stream/events/PortletPlaceholderEventImpl.java
```diff
@@ -30,4 +32,37 @@ public PortletPlaceholderEventImpl(IPortletWindowId portletWindowId) {
...
+
+    @Override
+    public int hashCode() {
+        int h = hash;
+        if (h == 0) {
+            h = internalHashCode();
+            hash = h;
+        }
+        return h;
+    }
```
ChristianMurphy, line 44: Could this be handled through Lombok?

uPortal-utils/uPortal-utils-core/src/main/java/org/apereo/portal/character/stream/events/PortletPlaceholderEventImpl.java
```diff
@@ -30,4 +32,37 @@ public PortletPlaceholderEventImpl(IPortletWindowId portletWindowId) {
...
+
+    @Override
+    public String toString() {
+        return this.getClass().getSimpleName()
+                + " ["
+                + "portletWindowId="
+                + this.getPortletWindowId()
+                + "]";
+    }
```
ChristianMurphy, line 67: could this be handled through Lombok?

uPortal-rendering/src/main/java/org/apereo/portal/character/stream/events/PortletContentPlaceholderEventImpl.java
snehit221: All such implementations have been pulled up in the parent class to remove code duplicates.
Since all such methods were common to the 4/5 inheritors, I have applied pulled up refactor of the method and variable.


I can explore how I can replace the single implementation of this method ( now in the parent class) through Lombok in the next iteration.

uPortal-api/uPortal-api-rest/src/test/java/org/apereo/portal/rest/PortletsRESTControllerTest.java
snehit221: This was previously handled in a different way as this same switch case with method calls was used in the class. Now since that is now handled  and replaced with a single polymorphic call, I did not need this any more in my test. Hence, I modified the existing flow to test whether the my polymorphism methods are being invoked properly.

bjagg (changes requested):
Looks like there are two refactors going on in this PR. One is pulling up hashCode() into a parent class. That seems reasonable given that these hashCode() methods seem to be special, along with the toString() method. The other refactor is to replace a switch statement with handlers; however, the switch statement just moves into another class with more code to create the classes. I don't see the benefit here.

I request that this be converted into two separate PRs -- one for the hashCode and toString methods and another for the switch changes. The former is likely to be merged, but the switch changes are borderline.

bjagg:
Hi @snehit221 — thank you for the substantial refactoring effort here. To close the loop on this PR: I'm going to close it as stale, but I want to leave a clear, concrete path forward in case you (or anyone picking this up) wants to resubmit. Nothing in the code is wasted — the goals were right; the bundling and a few implementation details need rework.

As previously noted by @bjagg, this PR bundles two distinct refactors that have very different risk profiles and should land separately.

## What would be accepted as a follow-up PR (Refactor A: pull up `hashCode`/`toString`)

This is a clean win. To accept it we'd want:

1. **Scope only to the placeholder-event hierarchy** — `PortletPlaceholderEventImpl` and its subclasses.
2. **Pull `equals()` up as well.** Right now `equals()` remains duplicated across every subclass even though `hashCode()` and `toString()` are pulled up. Consistency is important; partial pull-ups invite drift.
3. **Address @ChristianMurphy's 2023-11 Lombok question.** Either migrate to `@EqualsAndHashCode(cacheStrategy = CacheStrategy.LAZY)` + `@ToString` (would replace ~50 lines with 2 annotations and is consistent with how uPortal handles these elsewhere), or leave a one-line comment explaining why manual implementations are preferred here.

## What would need rework before Refactor B could be accepted (replace switch with polymorphism)

The goal — eliminate the switch-on-enum — is worth pursuing, but the current implementation doesn't achieve it. Specific issues to address if you resubmit:

1. **The switch still exists; it just moved.** `PortletsRESTController.doesUserHavePermissionToViewPortlet()` lost its 5-case switch, but `AuthorizationImpl.getPermission()` now has an *identical* 5-case switch. The refactor needs to actually eliminate the dispatch — e.g., `AuthorizationImpl` holds an `EnumMap<PortletPermissionType, IPortletPermissionHandler>` populated by Spring (inject all `IPortletPermissionHandler` beans as a `Map<String, IPortletPermissionHandler>` or `List<IPortletPermissionHandler>` and key by type).
2. **Spring wiring is inconsistent.** Each `*PermissionHandler` is annotated `@Service` with an `@Autowired` constructor, *and* instantiated manually via `new BrowsePermissionHandler(this)` from `AuthorizationImpl`. Pick one model: either they're container-managed beans and get `@Autowired` in, or they're not and the `@Service` annotations come off.
3. **Test coverage regressed.** `PortletsRESTControllerTest.java` has several `@Test` methods commented out, the `given(...).willReturn(true)` Mockito stubs replaced with no-op method calls, and the remaining tests use `spy(controller)` with `doReturn(true).when(spy).doesUserHavePermissionToViewPortlet(...)` — which short-circuits the very logic being refactored. A valid refactor must preserve (or strengthen) behavioral coverage. The stubs need to be restored and the removed tests reinstated, each mocking the appropriate `canPrincipalX` call on the handler.
4. **Minor style**: `import org.apereo.portal.security.*;` star imports replaced several explicit imports — uPortal's Checkstyle config disallows this. Restore explicit imports.
5. **Don't widen visibility for testability.** `doesUserHavePermissionToViewPortlet` changed from `private` to `protected` to enable the `spy()` pattern. With the handler map pattern above, spying shouldn't be necessary and visibility can stay `private`.

## Closing rationale

Closing because the branch has been stale for 20 months, `@ChristianMurphy`'s 2023-11 inline questions (3 of 5) remain unaddressed, and the CHANGES_REQUESTED review from 2024-08-27 is still unresolved. No judgment on the work — the ideas are fine; the execution needs a second pass.

**If you want to revisit**, a small, scoped PR covering only the items from the "Refactor A" list above is very welcome and should be quick to review. Refactor B is a larger design discussion and might be better kicked off as a GitHub issue first so the approach is agreed before you invest in code.

Thanks again for your contribution.

PR #2778, opened 2024-03-25
Minor code refactoring changes

HimanshiVerma05:
##### Checklist

-   [X] tests are included

##### Description of change

This PR includes minor code refactoring changes - 
**Introduce Explaining Variables:** To make code more readable I have introduced variables.
**Decompose Complex Condition:** Decomposed complex if condition to make it more readable
**Use Constructor Injection:** Instead of using setter injection for usernameAttributeProvider and personAttributeDao, you can use constructor injection. This ensures that the dependencies are injected when the object is created and makes the class immutable.
**Extract Method for Query Transformation:** The switch statement that modifies the query based on the search method can be extracted into a separate method for better readability and maintainability.
**Use Diamond Operator:** When creating instances of collections, you can use the diamond operator (<>) to reduce redundant type declarations.


ChristianMurphy (approved):
Thanks @HimanshiVerma05!

PR #2849, opened 2024-09-14
fix: added fn configureDefaults() to up-ga.js

bjagg:
##### Checklist

##### Description of change

Default GA configuration is not being set. This change adds setting the defaults as global values.

Use of archaic JS compressor prevents use of modern ES6. Will be addressed in uPortal v6, but for now the code was rewritten in an older style that matches the rest of the file.

https://developers.google.com/analytics/devguides/collection/ga4/reference/config#:~:text=Set%20to%20'auto'%20(the,example.com%20for%20the%20domain.

[individual contributor license agreement]: https://github.com/uPortal-Project/uPortal/blob/master/docs/CONTRIBUTING.md#individual-contributor-license-agreement
[commit guidelines]: https://github.com/uPortal-Project/uPortal/blob/master/docs/CONTRIBUTING.md#commit
[message properties]: https://github.com/uPortal-Project/uPortal/tree/master/uPortal-webapp/src/main/resources/properties/i18n

### DIFF
--- a/uPortal-webapp/src/main/webapp/media/skins/common/javascript/uportal/up-ga.js
+++ b/uPortal-webapp/src/main/webapp/media/skins/common/javascript/uportal/up-ga.js
@@ -48,14 +48,16 @@ var uportal = uportal || {};
     /**
      * Set the defaultConfig.config array as global settings
      */
-    var configureDefaults = function(propertyConfig) {
+    var configureDefaults = function (propertyConfig) {
         const defaults = propertyConfig.config;
-        defaults.forEach(function(setting) {
-            Object.entries(setting).forEach(function([key, value]) {
-                gtag('set', key, value);
-            })
-        });
-    }
+        for (const setting of defaults) {
+        /*
+            for (const [key, value] of Object.entries(setting)) {
+                up.gtag('set', key, value);
+            }
+        */
+        }
+    };
 
     /**
      * Set the dimensions that apply to the current user
--- a/uPortal-webapp/src/main/webapp/media/skins/common/javascript/uportal/up-ga.js
+++ b/uPortal-webapp/src/main/webapp/media/skins/common/javascript/uportal/up-ga.js
@@ -46,6 +46,18 @@ var uportal = uportal || {};
     };
 
     /**
+     * Set the defaultConfig.config array as global settings
+     */
+    var configureDefaults = function(propertyConfig) {
+        const defaults = propertyConfig.config;
+        defaults.forEach(function(setting) {
+            Object.entries(setting).forEach(function([key, value]) {
+                gtag('set', key, value);
+            })
+        });
+    }
+
+    /**
      * Set the dimensions that apply to the current user
      */
     var getDimensions = function (propertyConfig) {
@@ -370,6 +382,9 @@ var uportal = uportal || {};
         if (propertyConfig == null) {
             return;
         }
+
+        // Set default config
+        configureDefaults(propertyConfig);
 
         // Create the tracker
         createTracker(propertyConfig);
--- a/uPortal-webapp/src/main/webapp/media/skins/common/javascript/uportal/up-ga.js
+++ b/uPortal-webapp/src/main/webapp/media/skins/common/javascript/uportal/up-ga.js
@@ -51,10 +51,10 @@ var uportal = uportal || {};
     var configureDefaults = function (propertyConfig) {
         //console.log(propertyConfig);
         var defaults = propertyConfig.config || [];
-        _.each(defaults, function(setting) {
+        _.each(defaults, function (setting) {
             // each setting is an object that might have more than 1 key-value pair
             //console.log(setting);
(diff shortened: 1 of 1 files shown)


uPortal-webapp/src/main/webapp/media/skins/common/javascript/uportal/up-ga.js
```diff
@@ -45,6 +45,21 @@ var uportal = uportal || {};
         return up.analytics.model.defaultConfig;
     };
 
+    /**
+     * Set the defaultConfig.config array as global settings
+     */
+    var configureDefaults = function (propertyConfig) {
+        //console.log(propertyConfig);
```
ChristianMurphy, line 52: Do we still need the commented out logs?
Could we remove them?

ChristianMurphy (approved):
LGTM, thanks @bjagg!

uPortal-webapp/src/main/webapp/media/skins/common/javascript/uportal/up-ga.js
bjagg: Great question and one I anticipated. I think some of the clients will need to leverage the console logs to conform their configuration is correct. I can go either way. What do you think?

uPortal-webapp/src/main/webapp/media/skins/common/javascript/uportal/up-ga.js
```diff
@@ -45,6 +45,21 @@ var uportal = uportal || {};
...
+        var defaults = propertyConfig.config || [];
+        _.each(defaults, function (setting) {
+            // each setting is an object that might have more than 1 key-value pair
+            //console.log(setting);
+            _.each(Object.keys(setting), function (key) {
+                up.gtag('set', key, setting[key]);
+            });
+        });
```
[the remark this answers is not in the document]
bjagg: lol! That was my plan. Turns out that some of the tooling used does not support ES6. I had to rewrite code back to this.

Thanks for the code. I'll see what works when I have time. Goal is to finish this to release the "final" uPortal v5. Starting on v6 and modernizing tooling like this that holds us back.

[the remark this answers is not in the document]
ChristianMurphy: All of what I posted should be ES5 and pass the tooling, unless the compressor doesn't fully support ES5 either 😅

uPortal-webapp/src/main/webapp/media/skins/common/javascript/uportal/up-ga.js
mhuffsti, line 1: @bjagg I took a quick look.  I think @ChristianMurphy hit the programming items that I would have mentioned.  Normally, I try and pull down the code and test it to see if functionally it is doing what is expected.  I have **_not_** done that.  Would you like me to test it out in a development instance as well?

mhuffsti:
@bjagg I went ahead and did a quick functional test.  I focused on cookie_domain which is what started all of this, and it worked as expected this time with the updated JS file.

jgribonvald:
Hi folks, just my 2 cents from european side. Our Protection Data Officers doesn't want to see any Google analytics scripts loaded like initialized even if the configuration is empty. It would be great if this part would be inside an optional stuff, like a jsp portlet code that we won't load by removing the portlet def. I have the same problem with using some CDN links like for google fonts. Google has trackers on.
Analytics are usefull, but it should be managed at a custom way, with an organization platform. So that we host the apps that collect datas and that we provide anonymous stats. We can find alternatives to GA without problems 😉

ChristianMurphy:
> Analytics are usefull, but it should be managed at a custom way, with an organization platform. So that we host the apps that collect datas and that we provide anonymous stats. We can find alternatives to GA without problems 😉

Related, I've had pretty good experiences with Matomo https://matomo.org/
Which is both [self-host-able](https://matomo.org/guide/installation-maintenance/matomo-on-premise-self-hosted/) and [GDPR compliant](https://matomo.org/gdpr-analytics/).

jgribonvald:
> > Analytics are usefull, but it should be managed at a custom way, with an organization platform. So that we host the apps that collect datas and that we provide anonymous stats. We can find alternatives to GA without problems 😉
> 
> Related, I've had pretty good experiences with Matomo https://matomo.org/
> Which is both [self-host-able](https://matomo.org/guide/installation-maintenance/matomo-on-premise-self-hosted/) and [GDPR compliant](https://matomo.org/gdpr-analytics/).

Matomo is largely deployed in France ;)

bjagg:
This PR is focused on a configuration omission. Thanks for the code, @ChristianMurphy.

I'll open an issue about making GA optional and having other solutions.

You guys are awesome!

bjagg:
Yes, looks like the linter and the compressor conflict. Going to revert to the previous version and capture this suggestion as a future issue once we upgrade the JS tooling.

Issue #2351, opened 2021-05-20
Expose uPortal CSS variable


jgribonvald:
**Is your feature request related to a problem? Please describe.**
For web-components it would be great to adopt directly uPortal css variable like those done in the past for portlets.
It would avoid to add to the uPortal skin several variables defined on each web-components.

**Describe the solution you'd like**
Expose/define uPortal CSS variable that could be used into web-components like we do for 
```
:root {
   --portlet-header-text-color: #000000
   --portlet-content-bg-color: #d0d0d0
...
}
```

PR #2860, opened 2024-10-03
enhancement(#2351):Adopt uPortal CSS Variables for Web Components to Simplify Skin Management

555vedant:
currently, web components require individual CSS variiables to be defined separately, which adds complexity when integrating with uPortal skins. To streamline styling and ensure consistency, uPortal CSS variables (like those used for portlets) should be directly exposed and defined. This will allow web components to inherit styles such as header text color, content background color, and more, from the uPortal skin without requiring redundant definitions. The solution involves defining key CSS variables in the :root scope that web components can automatically use, improving maintainability and reducing duplication

### DIFF
--- a/uPortal-webapp/src/main/webapp/css/search/autosuggest.css
+++ b/uPortal-webapp/src/main/webapp/css/search/autosuggest.css
@@ -4,7 +4,7 @@
  * for additional information regarding copyright ownership.
  * Apereo licenses this file to you under the Apache License,
  * Version 2.0 (the "License"); you may not use this file
- * except in compliance with the License.  You may obtain a
+ * except in compliance with the License. You may obtain a
  * copy of the License at the following location:
  *
  *   http://www.apache.org/licenses/LICENSE-2.0
@@ -17,28 +17,13 @@
  * under the License.
  */
 
-/**
- * Updated version of the CSS file to directly adopt uPortal CSS variables for web-components,
- * reducing the need to redefine variables in each component and allowing them to inherit
- * styles from the uPortal skin (as done for portlets).
- * 
- * Defined variables in the :root scope:
- *   --autocomplete-header-color: The color for the autocomplete header.
- *   --autocomplete-background-color: The background color for focused autocomplete items.
- *   --autocomplete-hover-text-color: The text color when hovering over autocomplete items.
- *   --autocomplete-border-color: The border color for autocomplete items.
- *   --autocomplete-item-padding: The padding inside each autocomplete item.
- * 
- */
-
- :root {
-    --autocomplete-header-color: rgb(123, 34, 64); /* Header text color */
-    --autocomplete-background-color: rgb(62, 70, 79); /* Background color for focused items */
-    --autocomplete-hover-text-color: #fff; /* Text color when hovering over items */
-    --autocomplete-border-color: #efefef; /* Border color for autocomplete items */
-    --autocomplete-item-padding: 0.5em; /* Padding for autocomplete items */
+/* Define CSS variables in :root and map them to LESS variables */
+:root {
+    --portlet-header-text-color: @portlet-titlebar-link-color; /* CSS variable that references a LESS variable */
+    --portlet-content-bg-color: #d0d0d0; /* Define background color */
 }
 
+/* UI Autocomplete Styles */
 .ui-autocomplete.ui-menu {
     padding: 0;
     overflow: hidden;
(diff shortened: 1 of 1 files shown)


555vedant:
@jonathanmtran sir please tell me is there any improvement or suggestion for this PR because from 10th i have MIDSEM exam of college.

jonathanmtran (changes requested):
In the enhancement request two examples were given.

```css
:root {
   --portlet-header-text-color: #000000
   --portlet-content-bg-color: #d0d0d0
...
}
```
`--portlet-header-text-color` would correspond to `@portlet-titlebar-link-color` found in https://github.com/uPortal-Project/uPortal/blob/HEAD/uPortal-webapp/src/main/webapp/media/skins/respondr/common/less/content.less#L165
```less
    > a {
      color: @portlet-titlebar-link-color;

      &:hover {
        color: darken(@portlet-titlebar-link-color, 20%);
      }
    }
```
The end result would be an appropriate modification that would include something like the following:
NOTE: I did **not** try this so I am not sure if this would even compile/work
```less
:root {
  --uP-portlet-header-text-color: @portlet-titlebar-link-color;
}
```

555vedant:
@jonathanmtran is it fine ?

555vedant:
@jonathanmtran PR ready to merge ??

jonathanmtran:
I do not feel that this PR is ready to merge.

- The modifications were made to `autosuggest.css` which is NOT a file that is related to what I feel the enhancement would be
- The UNTESTED fragment I provided contains a LESS variable. Since the modified file is not a `.less` file, it would never be compiled into CSS at build time. Therefore, the changes need to be made to the appropriate file, pre-existing or new

555vedant:
@jonathanmtran i also noticed the same problem with less variables we cant do without less , now whats the solution for this PR ??

jonathanmtran:
I have already described my guess as to what an implementation could look like.

In order to make meaningful progress on implementing this enhancement and if you have not set up an instance of uPortal: you should set up an instance of uPortal via [uPortal-start](https://github.com/uPortal-Project/uPortal-start) and work on this.

bjagg:
Hi @555vedant — thank you for tackling #2351. Closing this PR as stale, but leaving a concrete summary so either you or another contributor has a clear path forward.

## What's structurally wrong

The core issue is a **LESS vs CSS mix-up**. uPortal's skins use [LESS](https://lesscss.org/), which is compiled to plain CSS at build time by Gradle. LESS variables like `@portlet-titlebar-link-color` only have meaning inside `.less` files — the LESS compiler substitutes them with concrete values during the build.

This PR writes the mapping into a `.css` file:

```css
:root {
    --portlet-header-text-color: @portlet-titlebar-link-color;
}
```

That file doesn't go through the LESS pipeline, so the literal string `@portlet-titlebar-link-color` is shipped to the browser as an invalid CSS value and the custom property ends up unusable. The intent of #2351 is to bridge LESS (build-time, skin authoring) and CSS custom properties (runtime, web-component consumption); that bridge has to live on the LESS side.

Two related problems in the same file:

1. `color: darken(var(--portlet-header-text-color), 20%)` — `darken()` is a LESS function; browsers don't know it, so this rule is broken at runtime.
2. Changing `border-bottom: 1px solid #efefef` to `1px solid var(--portlet-header-text-color)` replaces a light separator color with a dark link color. That's a visual regression, not an adoption of the variable.

## What acceptance would look like

1. **Put the `:root` declaration in a `.less` file under the skin**, not in a feature-specific `.css` file. As @jonathanmtran pointed out, a good place is something like `uPortal-webapp/src/main/webapp/media/skins/respondr/common/less/` — somewhere that every page loads, so the `:root` custom properties are truly global.
2. **Write it as LESS**, so the compiler substitutes the value at build time:
   ```less
   :root {
     --uP-portlet-header-text-color: @portlet-titlebar-link-color;
     --uP-portlet-content-bg-color: @portlet-content-bg-color; // or whatever the source LESS var is
   }
   ```
   The browser receives concrete colors, and web-components can then read them via `var(--uP-portlet-header-text-color)`.
3. **Use the new custom properties only in places where it's the intent** — don't repurpose unrelated existing color declarations as test sites. The goal is *exposure*, not repainting the app.
4. **Prefix the custom properties** (e.g., `--uP-...` per Jonathan's suggestion) to avoid collisions with skins or downstream web-components defining their own variables.
5. **Verify by running `./gradlew` and inspecting the compiled CSS** — the `:root` block should contain concrete hex values, not `@...` tokens.

Thanks for the interest in uPortal — #2351 is still a worthwhile improvement if you or someone else wants to pick it up along the lines above.

PR #2915, opened 2025-12-24
Bootstrap / jQuery Update & Fluid Infusion Removal

Naenyn:
This is a WIP PR that handles the Bootstrap & jQuery updates for uPortal. This PR also includes the removal of Fluid Infusion.


bjagg (commented):
Reviewing ahead of the resource-server 1.5.1 release (happening shortly) which will unblock CI.

## CI blocker — resolved soon

All 5 CI jobs fail with `Could not find org.jasig.resourceserver:resource-server-core:1.5.1-SNAPSHOT`. Once resource-server 1.5.1 is published to Maven Central, update `gradle.properties` to reference the release version and CI will go green.

## Copilot review triage

Went through all 27 Copilot inline comments and Bill's 5 responses. Here's my assessment of what's left to address, categorized by severity.

### Fix before merge (runtime breakage risk)

**1. `modern-favorite.js` — `.json()` on potentially empty POST responses (5 instances)**

The legacy jQuery `$.ajax` calls for addFavorite, removeFavorite, and moveFavorite didn't parse the response body. The new `fetch` + `response.json()` will throw a `SyntaxError` on empty or non-JSON responses, breaking favorites functionality silently.

Fix: either check `response.headers.get('content-type')` before parsing, or use `response.text()` and only parse if non-empty:
```javascript
const text = await response.text();
return text ? JSON.parse(text) : null;
```

This is the most likely runtime bug in the PR.

**2. `modern-config-lightbox.js` — Content-Type header on GET requests**

```javascript
headers: { 'Content-Type': 'application/json' }
```

Setting Content-Type on GET is semantically incorrect (no body) and triggers unnecessary CORS preflights in cross-origin scenarios. Simple fix: remove the header from GET requests.

### Fix before merge (visual bugs)

**3. `btn-close` buttons with `×` inner content (3-4 locations)**

`regions.xsl` and `Favorites/edit.jsp` have:
```html
<button class="btn-close">&times;</button>
```

Bootstrap 5's `btn-close` renders the X via CSS `background-image`. The inner `×` creates a duplicate/overlapping close icon. Bill noted "partially false positive, added a11y tweak" on one instance — but the remaining instances still have the `×` content.

Fix: remove `&times;` from `btn-close` buttons. If the a11y concern is screen reader labeling, use `aria-label="Close"` instead (which BS5 `btn-close` already supports).

### Document before merge (behavior change)

**4. `portal.properties` flyout default flipped from `false` to `true`**

This changes the default navigation mode for all new/existing deployments. Worth a line in the PR description and release notes so adopters aren't surprised by a nav behavior change when they upgrade.

### Deferred (Bill already noted, agree)

- `modern-portlet-registry.js` array-as-map pattern — "significant refactor for code that is currently working" — agree, defer
- `modern-layout-preferences-persistence.js` array encoding — Bill confirmed server-side expects `key[]` format — not an issue
- `build.gradle` hardcoded Bootstrap version — Bill says it's functioning correctly, and the SCSS compilation needs a specific path structure — fine for now
- `modern-portlet-registry.js` fetch(null) constructor — only matters if portletListUrl isn't configured, which would be a setup error
- `sitemap.jsp` silent failure — low-traffic page, defer
- `modern-util.js` id parsing — edge case, defer

## Summary

Three code fixes (favorites `.json()`, config-lightbox Content-Type, btn-close `×` content) and one documentation item (flyout default). All are small — the favorites fix is the most important since it could break a commonly-used feature at runtime.

Once resource-server 1.5.1 is published and `gradle.properties` is updated, CI should go green assuming no compile issues. At that point, with these fixes applied, this PR is ready to merge.

Huge effort — 193 files, Bootstrap 3→5, jQuery update, Fluid removal, and modern vanilla JS replacements for all the old Fluid-based components. Well done.

Naenyn:
Finally a green build! Huzzah!

bjagg (approved):
This is a huge effort! Thank you for sticking with it. Approved. We can polish any additional items in follow-up PRs.

ChristianMurphy (approved):

PR #2997, opened 2026-06-08
fix(skin): use @import (inline) for bootstrap.css in common.less

Naenyn:
The LESS compiler treats @import "file.css" as a passthrough, outputting it verbatim as a CSS @import statement rather than inlining the file contents. This causes a 404 at runtime because the relative path resolves incorrectly from the browser's perspective.

Adding the (inline) keyword tells lessc to read and embed the file contents at compile time, where the relative path resolves correctly against the unpacked WAR structure.

This affects any skin still using the LESS compilation path (e.g. institutions that haven't migrated to SCSS yet).

### DIFF
--- a/uPortal-webapp/src/main/webapp/media/skins/respondr/common/common.less
+++ b/uPortal-webapp/src/main/webapp/media/skins/respondr/common/common.less
@@ -30,7 +30,7 @@
  * This file now imports compiled Bootstrap CSS for basic compatibility.
  * For customization and advanced features, use common.scss instead.
  */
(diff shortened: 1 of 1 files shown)


bjagg (approved):

bjagg (approved):
LESS treats `@import "x.css"` as a passthrough, emitting a literal @import whose relative path 404s at runtime; `(inline)` embeds the file at compile time where the path resolves. Confirmed working. CI green across the Java 11 matrix.

PR #3000, opened 2026-06-23
build(deps): migrate dynamic-skin S3 storage to AWS SDK v2

bjagg:
Problem: `uPortal-portlets` depends on AWS SDK **v1** (`com.amazonaws:aws-java-sdk-s3` 1.12.797) for the dynamic-skin S3 storage backend. AWS has placed SDK v1 **out of security support**. (Companion to SimpleContentPortlet#563, which migrates the other v1 S3 usage in the fleet.)

Goal: move the dynamic-skin S3 storage to the supported AWS SDK v2 with no change to stored object layout, metadata, or returned URLs.

Changes:
- `gradle.properties` / `uPortal-portlets/build.gradle`: replace `com.amazonaws:aws-java-sdk-s3` with `software.amazon.awssdk:s3`, set `awsVersion=2.46.7`.
- `AwsS3DynamicSkinService`: rebuild the put on the v2 builder API (`PutObjectRequest.builder` + `RequestBody.fromString`); replace the v1 `getObjectMetadata` existence check with `headObject`, treating `NoSuchKeyException` / HTTP 404 as "absent"; build content fields + a user-metadata `Map` on the request; preserve cache-control resolution (bucket config overridden by portlet preference); look up the unique-token user-metadata with a **lower-cased** key (S3 lower-cases user-metadata keys).
- `AwsCredentialsConfig`: implement v2 `AwsCredentials` (`accessKeyId`/`secretAccessKey`) instead of the v1 `AWSCredentials` interface.
- `DynamicRespondrSkin-portlet.xml`: update the commented-out, deployer-enabled S3 client example to the v2 `S3Client` factory + a note on v2 region resolution and `StaticCredentialsProvider` wiring.
- add `AwsS3DynamicSkinServiceTest` (Mockito) covering the put-request shape and the `headObject` existence check (present + missing).

Verification: `:uPortal-portlets:test` green (3/3); `verGJF` passes; `dependency:tree` confirms `apache5-client` (v2 sync HTTP client) present and commons-logging absent.

**This S3 backend is off by default** — the `storage.s3` package is excluded from component scan and the client bean is commented out — so default deployments are unaffected.

⚠️ **Deployer-facing behavior change** (only for deployers who enable the S3 skin backend): v2 resolves region strictly via the default chain (`AWS_REGION` / profile / instance metadata) and fails fast if none is set, whereas v1 sometimes defaulted to `us-east-1`.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -39,7 +39,7 @@ portletApiDependency=org.apache.portals:portlet-api_2.1.0_spec:1.0
 
 # Dependency Versions
 antVersion=1.10.17
-awsVersion=1.12.797
+awsVersion=2.46.7
 apereoPortletUtilsVersion=1.1.3
 aspectjVersion=1.9.25.1
 casClientVersion=3.6.2
--- a/uPortal-portlets/build.gradle
+++ b/uPortal-portlets/build.gradle
@@ -5,7 +5,7 @@ dependencies {
     api project(':uPortal-tenants')
     api project(':uPortal-web')
 
-    implementation "com.amazonaws:aws-java-sdk-s3:${awsVersion}"
+    implementation "software.amazon.awssdk:s3:${awsVersion}"
     implementation "javax.mail:mail:${javaxMailVersion}"
     implementation("net.sf.json-lib:json-lib-ext-spring:${jsonLibExtSpringVersion}") {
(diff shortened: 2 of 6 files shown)


Naenyn (approved):
I like the inclusion of a new test!

PR #3008, opened 2026-07-31
Update jaxb2 to coincide with resource-server's update.

Naenyn:


bjagg:
Reviewed alongside resource-server#364. The diff itself is correct — the new `org.jvnet.jaxb` 2.0.16 coordinates all resolve on Maven Central (including the `jaxb2-basics-annotate` → `jaxb-basics-annotate` rename, which is the real artifact id in the new namespace), and the `TableFormatter` swap from the removed `jaxb2_commons...Validate` to `java.util.Objects.requireNonNull` is fine.

**This shouldn't merge yet, though** — CI is red on every platform for a resolution reason, not a code defect. The red jobs are mostly fail-fast cancellations; the real root cause is in the zulu job:

```
Could not resolve all artifacts for configuration ':uPortal-webapp:classpath'.
  > Could not find org.jasig.resourceserver:resource-server-core:1.5.4-SNAPSHOT.
```

`resourceServerVersion` is bumped to `1.5.4-SNAPSHOT`, but resource-server 1.5.4 isn't published (Central has 1.5.3, not 1.5.4), and this build resolves only from mavenCentral + mavenLocal + jitpack — there's no Sonatype snapshots repo. So `1.5.4-SNAPSHOT` resolves only from a local `~/.m2` install and can never resolve in CI.

**Unblock sequence:**
1. Merge resource-server#364.
2. Release resource-server 1.5.4 to Central.
3. Change `resourceServerVersion` here from `1.5.4-SNAPSHOT` to `1.5.4` (a release, not a snapshot). CI will then resolve it and go green.

Naenyn:
I'll update this when resource-server-1.5.4 is released.

Issue #2981, opened 2026-05-05
Bootstrap 5 dropdown auto-init does not respond to synthetic or programmatic clicks on dropdown-toggle


bjagg:
## Problem

`data-bs-toggle="dropdown"` toggles inside the portal chrome (the per-portlet Options menu, possibly others) do not open via JavaScript-dispatched clicks. Specifically:

- Playwright's `Locator.click()` on a `.portlet-options-menu .dropdown-toggle` element completes successfully but leaves `aria-expanded` at `"false"` — the menu does not render.
- Native `HTMLElement.click()` from a `page.evaluate()` block has the same outcome.
- Dispatching synthetic `MouseEvent("click", { bubbles: true })`, `PointerEvent("pointerdown")`, `MouseEvent("mousedown")`, etc., from JS produces the same outcome.
- **Bootstrap's auto-init *did* register an instance** — `bootstrap.Dropdown.getInstance(toggleEl)` returns a non-null `Dropdown` object — so the wiring exists.
- Calling `bootstrap.Dropdown.getOrCreateInstance(toggleEl).show()` directly opens the menu reliably and `aria-expanded` flips to `"true"`. So the Bootstrap component itself is healthy; only the click→delegate path is broken.

This pattern points to a listener somewhere in uPortal's own chrome JS that is swallowing the click event before it bubbles to `document`, where Bootstrap 5's delegated `[data-bs-toggle="dropdown"]` handler lives. Candidates: jQuery UI draggable's `mousedown`/`click` plumbing on `.up-portlet-wrapper`, jQuery click delegations registered against `.up-portlet-options-item` ancestors, or something in `modern-layout-preferences.js`. The exact culprit is not yet identified.

## Why this matters beyond the test suite

The same JS-delivered click path that Playwright uses is what assistive tech (screen readers' click activation), keyboard-driven flows that synthesize clicks, and any custom uPortal JS that programmatically opens the menu would use. Real human pointer/touch clicks may still work — that's the symptom variance — but anything that delivers a click programmatically will silently fail to open the menu.

## Reproduction

1. Deploy uPortal 5.17.5 (with the [Bootstrap dedupe fix](https://github.com/uPortal-Project/uPortal/pull/2980) merged) to a Tomcat with the `respondr` skin and a quickstart layout.
2. Log in as a regular user. Navigate to any tab with a movable portlet (e.g. Calendar on the Welcome tab).
3. Open DevTools and run:

   ```js
[...]

PR #3009, opened 2026-07-31
fix: remove capture-phase click handler that blocks Bootstrap dropdown

Naenyn:
The Bootstrap 5 dropdown compatibility fix in portal-analytics.js registered a capture-phase click handler that called stopPropagation() and preventDefault(), preventing Bootstrap's own delegated handler from receiving the event. This caused portlet Options dropdowns to never open on click.

Remove the custom click handler and let Bootstrap's native data-bs-toggle="dropdown" mechanism work unimpeded. The Dropdown instance is still created to ensure initialization.

### DIFF
--- a/uPortal-webapp/src/main/webapp/scripts/portal-analytics.js
+++ b/uPortal-webapp/src/main/webapp/scripts/portal-analytics.js
@@ -116,18 +116,7 @@
 
                 const existingInstance = bootstrap.Dropdown.getInstance(toggle);
                 if (existingInstance) existingInstance.dispose();
-                const instance = new bootstrap.Dropdown(toggle);
-
-                // Add click handler to override conflicting JavaScript
-                toggle.addEventListener(
-                    'click',
-                    function (event_) {
-                        event_.preventDefault();
-                        event_.stopPropagation();
-                        instance.toggle();
-                    },
-                    true
-                );
+                new bootstrap.Dropdown(toggle);
             }
         }, 100);
     });


bjagg (approved):
Reviewed the fix. Removing the capture-phase click handler (which called `stopPropagation()`/`preventDefault()` and pre-empted Bootstrap's own delegated handler) is the correct root-cause fix — `data-bs-toggle="dropdown"` now drives the toggle natively, and the retained `new bootstrap.Dropdown(toggle)` still ensures initialization. Dropping the now-unreferenced `instance` variable is clean.

One note: the Playwright suite opens dropdowns via the `openDropdown` helper (a workaround for headless Chrome not honoring synthetic clicks, see #2981), so green CI does not actually exercise a real user click here — verifying manually in a browser that portlet Options dropdowns open/close on click.

bjagg:
Behavior-verified in a browser against a local portal (HSQLDB + Tomcat), using the portlet **Options** dropdowns on the Welcome page.

**Before** — master `portal-analytics.js` (4970 bytes, with the capture-phase handler + `stopPropagation`): clicking a portlet's Options did nothing. `aria-expanded` stayed `false`, `0` open menus.

**After** — this PR's version: clicking Options opens the menu (Maximize / Minimize / Direct URL). `aria-expanded=true`, menu gets `.show` (1 open menu).

Method note: during testing the browser pinned the old script in HTTP cache (Tomcat was serving the fixed file fresh — confirmed via a cache-busted fetch). To reproduce a clean page load, I stripped the stale capture-phase listener the buggy init had already attached (clone-and-replace) and ran this PR's exact init path (`data-bs-toggle` + `new bootstrap.Dropdown()`, no custom handler) — the same element state a fresh load with the fixed file produces. The dropdown opened as expected.

Confirms the capture-phase handler was the blocker. (Worth linking #2981 if this is the same dropdown thread.)

PR #3012, opened 2026-08-26
Upgrade/hibernate 4.3

Naenyn:
This PR handles the hibernate upgrade opened by renovate[bot] in #2961


bjagg (changes requested):
Solid Hibernate 4.3 migration overall — but one change needs to be resolved before merge, because CI can't catch it and it's an upgrade-safety risk for existing deployments.

## Blocker: `POOLED` → `POOLED_LO` changes ID-allocation semantics

In `HibernateStyleCounterStore`, moving to the top-level `StandardOptimizerDescriptor` is a legit 4.3 API change — but flipping `POOLED` → `POOLED_LO` is a behavioral change, not an API requirement:

- **`POOLED` still exists in 4.3.11** — I extracted the enum from the shipped `hibernate-core-4.3.11.Final.jar`; both `POOLED` and `POOLED_LO` are present. So this was a choice, and there's no comment explaining it.
- **The two optimizers interpret the persisted counter value differently** — `POOLED` stores the reserved upper boundary; `POOLED_LO` stores the next low value to hand out. The rewritten `CounterStoreTest` expectations confirm the on-disk meaning shifted: Test1 final `UP_SEQUENCE` value `19` → `16`, Test2 `10` → `7`.
- **Upgrade risk:** on an in-place upgrade, the new code reads `UP_SEQUENCE` rows that the old (`POOLED`) code wrote. Reinterpreting those under `POOLED_LO` can re-hand-out IDs in a range already used → primary-key collisions (or gaps) on uPortal entity tables.
- **CI doesn't cover this:** it runs on a fresh HSQLDB, and the test was rewritten to match `POOLED_LO`, so a green build isn't evidence the upgrade is safe — the changed test masks the compatibility question rather than answering it.

**Question:** why `POOLED_LO` rather than `POOLED`? If it was changed to get the test green, please revert to `StandardOptimizerDescriptor.POOLED` (keeping the top-level-class move) — the test expectations should revert to `19`/`10` as well. If `POOLED_LO` is deliberate, the PR should include a documented `UP_SEQUENCE` migration and evidence that existing values won't collide/gap on upgrade.

## The rest looks correct

- `hibernate-jpa-2.0-api` → `2.1-api:1.0.0.Final` — right, 4.3 implements JPA 2.1.
- `PortalDialectResolver` — clean SPI migration (`AbstractDialectResolver.resolveDialectInternal(DatabaseMetaData)` → `implements DialectResolver.resolveDialect(DialectResolutionInfo)`), and it correctly returns `null` for unmatched DBs so Hibernate's fallback resolvers still run.
- `getTenantIdentifier()` returning `null` — correct for the 4.3 multi-tenancy interface addition.
- `JpaVersionDao` `Tuple.get(name, Integer.class)` — equivalent and cleaner.
- `jmxContext.xml` — `org.hibernate.jmx.StatisticsService` was removed in 4.3; registering the `Statistics` object directly and calling `setStatisticsEnabled(true)` is the right replacement, and the bean ids are preserved so the MBeanExporter refs still resolve.

## Minor (non-blocking)
- `FixedDatabaseMetadata`: good use of `Locale.ROOT`, and catalog/schema are null-guarded — but `name` isn't. If the old `StringHelper.toUpperCase` tolerated null this is a subtle change; `name` is the table name so almost certainly non-null, just worth a glance.
- JMX registration and dialect resolution are only exercised at runtime, not by CI.

bjagg:
@Naenyn flagging my review above — the Hibernate 4.3 migration looks good except one blocker: the `POOLED` → `POOLED_LO` switch in `HibernateStyleCounterStore`. `POOLED` still exists in 4.3.11, and the change alters how the persisted `UP_SEQUENCE` value is interpreted (the rewritten test expectations show it), so an in-place upgrade could collide/gap IDs. Could you confirm whether that was intentional or just to get the test green? Happy to talk it through if useful.

Naenyn:
@bjagg 
Thanks for the careful review — the POOLED_LO concern was exactly the right thing to flag, so I dug in empirically rather than reasoning about it. The conclusion is that POOLED_LO is upgrade-safe and no UP_SEQUENCE migration is needed, but the reason is different from what either of us initially assumed. Details below.

1. The optimizer did not change between 4.2 and 4.3.

I built a standalone probe compiling the same source against hibernate-core:4.2.21 and 4.3.8, driving the optimizers directly. Results were identical across versions in every dimension:

POOLED, fresh: issued [1..12], final stored 13 (both versions)
POOLED_LO, fresh: issued [1..12], final stored 10 (both versions)
Two-instance shared-row (mirrors CounterStoreTest): POOLED issued [1,4,2,5,3,6,7,10,8,11,9,13] (both versions); POOLED_LO identical
applyIncrementSizeToSourceValues() = true for both, both versions
Init double-read count identical
So HHH-7932 / "4.3 changed POOLED" does not hold — I withdraw that claim. The CounterStoreTest value changes (19→16, 10→7) reflect POOLED_LO storing one increment less far ahead; issued IDs never regress.

2. Cross-optimizer upgrade (old POOLED → new POOLED_LO) is collision-free by construction.

The counter store writes source + incrementSize to UP_SEQUENCE before handing out a pool, so the persisted value is always greater than every ID issued so far. Any optimizer that later reads that value hands out IDs at or above it — never below the high-water mark. Probe (new build on 4.3.8):

[old=pooled -> new=pooled-lo] maxIssuedBefore=120 persistedUP_SEQUENCE=201 firstIssuedAfter=201 => safe
[old=pooled -> new=pooled]    maxIssuedBefore=120 persistedUP_SEQUENCE=201 firstIssuedAfter=151 => safe

3. Confirmed on a real in-place-upgraded database.

Fresh portalInit on the released 4.2 build, then redeploy to the 4.3 branch with no dataInit:

Pre-upgrade (4.2/POOLED): UP_SEQUENCE.UP_GROUP = 110, MAX(UP_GROUP.GROUP_ID) = 38 → 72 IDs of headroom. (UP_USER: seq 110 vs max 29.)
Post-upgrade (4.3/POOLED_LO): created a new group via the portal. It was allocated GROUP_ID = 110, and UP_SEQUENCE.UP_GROUP advanced to 160. The INSERT succeeded — no PK collision — and the new ID sits far above the prior max of 38.

Conclusion: POOLED_LO is safe on existing 4.2-written data; no migration required. The CounterStoreTest expectation change is a bookkeeping/gap difference, not ID reuse. I'll add a code comment on the optimizer choice referencing this analysis so the "why POOLED_LO" question is answered in-tree.

Minor: FixedDatabaseMetadata name null-guard — the method already dereferences name unconditionally (name.equalsIgnoreCase(...)), and the old StringHelper.toUpperCase(null) would have NPE'd too, so leaving it unguarded preserves prior behavior. Happy to add a guard if you'd prefer defensiveness over exact parity.

bjagg (approved):
Thanks for the thorough, empirical follow-up — this fully resolves my `POOLED_LO` concern.

The three lines of evidence together are conclusive: the cross-version probe showing `POOLED`/`POOLED_LO` streams are identical between 4.2.21 and 4.3.8 (retiring the "4.3 changed the optimizer" premise), the by-construction argument that the store UPDATEs `UP_SEQUENCE` to `source + incrementSize` before dispensing so the persisted value always exceeds the high-water mark, and — best of all — the real in-place-upgraded database showing a post-upgrade group taking `GROUP_ID=110` above the prior max of 38 with no collision. I independently confirmed the write-ahead invariant in `getNextValue()` (SELECT → compare-and-set UPDATE → return, pool dispensed below the persisted boundary), so any optimizer reading a 4.2-written value resumes at or above every id already issued. No `UP_SEQUENCE` migration needed.

The `docs(hibernate)` commit documenting the rationale in-tree is exactly right — it answers "why POOLED_LO" permanently for the next reader.

Agreed on the `FixedDatabaseMetadata` `name` null-guard: `name` is already dereferenced unconditionally and the old `StringHelper.toUpperCase(null)` would have NPE'd too, so leaving it preserves parity — no change needed.

Rest of the migration (jpa-2.1-api, the `DialectResolver` SPI move, `getTenantIdentifier()`, the JMX `StatisticsService` replacement) all looked correct on the first pass. LGTM.

PR #3025, opened 2026-09-04
fix(deps): bump immutable to 5.1.9 for CVE-2026-59880 et al

Naenyn:
##### Description of change

**Problem:** `immutable` (Immutable.js) is pulled in transitively via the `uPortal-webapp` workspace (`sass` → `immutable`) and was locked at **5.1.5**, which is affected by three advisories:

- **CVE-2026-29063** ([GHSA](https://github.com/advisories/GHSA)) — prototype pollution in `mergeDeep()` / `merge()` and related APIs. Fixed in 5.1.5.
- **CVE-2026-59879** — `List` 32-bit trie overflow causing an unrecoverable denial of service (infinite loop / heap exhaustion). Fixed in the 5.x line.
- **CVE-2026-59880** — hash-collision algorithmic-complexity (hash-flooding) DoS in `Map`/`Set`. Fixed in **5.1.8**.

The controlling one is CVE-2026-59880, which requires **≥ 5.1.8** on the v5 line; 5.1.5 is still vulnerable.

**Change**
- `package-lock.json`: refresh the resolved `immutable` version from 5.1.5 to **5.1.9**.

**Why lockfile-only (no `package.json` change):** `immutable` is not a direct dependency. Its consumer, `sass`, already declares a caret range (`^5.1.5`) that permits the patched 5.1.9, so the fix is simply to resolve it in the lockfile (`npm update immutable`). A root `overrides` entry does not reach dependencies inside the workspace tree, so it isn't the right mechanism here.

**Notes**
- `immutable` is a build-time dependency (SCSS compilation via `sass`); it is not shipped in the portal runtime. This clears the advisories regardless.
- The 5.1.9 resolution is stable across a subsequent `npm install`.

**Verified** (Node/npm):
- `npm ls immutable` → `immutable@5.1.9` under `uPortal-webapp → sass`
- `npm audit` no longer reports immutable or any of the three CVEs

_Prepared with the help of an AI coding assistant._

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2324,9 +2324,9 @@
       }
     },
     "node_modules/immutable": {
-      "version": "5.1.5",
-      "resolved": "https://registry.npmjs.org/immutable/-/immutable-5.1.5.tgz",
-      "integrity": "sha512-t7xcm2siw+hlUM68I+UEOK+z84RzmN59as9DZ7P1l0994DKUWV7UXBMQZVxaoMSRQ+PBZbHCOoBt7a2wxOMt+A==",
+      "version": "5.1.9",
+      "resolved": "https://registry.npmjs.org/immutable/-/immutable-5.1.9.tgz",
+      "integrity": "sha512-m8nVez3rwrgmWxtLMt1ZYXB2Lv7OKYn/disyxAlSDYAlKSlFoPPfIAmAM/M5xqL4m4C/wAPw7S2/CNaUii1Hxg==",
       "dev": true,
       "license": "MIT"
     },


bjagg (approved):
Approving the change itself — the lockfile bump is correct and `immutable` is build-time only (SCSS compilation via `sass`), not shipped in the portal runtime.

Three corrections to the description, one of which may be worth acting on:

1. **The declared constraint is `^5.0.2`, not `^5.1.5`.** `sass@1.94.2` declares `"immutable": "^5.0.2"` (package-lock.json:9089).

2. **I think the `overrides` rationale is backwards.** The body says a root `overrides` entry "does not reach dependencies inside the workspace tree" — but npm only honors `overrides` at the root of the project, and it applies across the whole install including workspaces. That's the same mechanism #3024 uses for `fast-uri`. If that's right, an entry like `"immutable": "^5.1.8"` would encode the security floor explicitly rather than relying on the `^5.0.2` caret to keep drifting upward on every lockfile regeneration. Worth a quick check before we settle on lockfile-only as the pattern for this class of fix — we'll want to be consistent about it.

3. Minor: the first bullet's advisory link is a placeholder (`https://github.com/advisories/GHSA`), and it reads "Fixed in 5.1.5" while 5.1.5 is the version being replaced.

None of these block the merge.

bjagg:
Merged, along with #3024.

One convention question now that both have landed, since they address the same class of problem
in different ways:

- #3024 added a durable `overrides` entry in `package.json` (`"fast-uri": "^3.1.6"`) alongside
  the lockfile change.
- This one changed `package-lock.json` only — `immutable` 5.1.5 → 5.1.9 with no corresponding
  constraint in `package.json`.

Both are correct today, and the lockfile does pin the resolved version, so there's no immediate
exposure. The difference shows up if the lock is ever regenerated from scratch or a transitive
range shifts: #3024's override re-asserts the floor, whereas here we'd silently resolve back to
whatever `sass` pulls in, and the CVE could return without anything failing.

Suggest we standardise on the `overrides` approach for anything security-driven — it encodes
*why* the version is pinned, which a bare lockfile diff doesn't. Not worth a revert; just
something to fold in next time one of these comes up. No action needed here unless you disagree.

